PR #6, opened 2022-02-18
fix: Handle all cases when a controller should be initialized and disposed

elias8:
This PR improves when a controller should be initialized and disposed. This will give a user the option to dispose a controller when the controller is created by the user. Currently, disposing a user-created controller is handled by the package itself, but by letting a user dispose, we can prevent calling the `dispose` method being called twice.


juliansteenbakker:

PR #12, opened 2022-02-22
rotate scanner camera preview on orientation change 

koltai75:
Currently when rotating the phone to landscape, the QR scanner is working but the camera preview is not rotated.
This fix rotates the preview with -90 degrees in landscape.
Since Flutter does not offer a solution to detect difference between landscape up and down, it will work only if the user rotates to landscape to the left.
If rotated to right, the preview will be upside down.

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/material.dart';
 import 'package:mobile_scanner/mobile_scanner.dart';
+import 'dart:math' as math;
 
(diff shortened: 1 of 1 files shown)


juliansteenbakker:
Thank you for your contribution! I do think we need to solve this, but we should support both orientations. I've started a discussion #17 to see how we can solve this issue.

juliansteenbakker:
I am going to implement an updated version of flutter_native_device_orientation to support all cases for now, so i am closing this PR.

koltai75:
Thanks for that, if I can help for example in testing, please let me know.

PR #13, opened 2022-02-22
Enhancement: Add null-safety variable

luqmanhakem:
Added null-safety to prevent the packages from throwing errors log if there is not enough sufficient data from the QR code.

**Issue:** QR code (UrlBookmark Type) is required to return title and url. Null exception thrown upon testing at generated QR code that returns empty title.

**Solution:** Add nullable String if the QR code returns empty title. Also, added nullable variable to other variables.

### DIFF
--- a/lib/src/objects/barcode.dart
+++ b/lib/src/objects/barcode.dart
@@ -202,7 +202,7 @@ class Address {
   final List<String> addressLines;
 
   /// Gets type of the address.
-  /// 
+  ///
   /// Returns null if not available.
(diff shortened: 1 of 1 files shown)


juliansteenbakker:
Thank you for your contribution! Can you please run `flutter format .` and push the changes?

luqmanhakem:
@juliansteenbakker Welcome! Done. 

Looking forward to contribute more for this package 🚀🚀🌝

PR #27, opened 2022-02-24
Call cameraProvider.unbindAll() before create preview

myoganugraha:
After user confirm the camera request, the apps will be force close because
```
Fatal Exception: java.lang.IllegalArgumentException
No supported surface combination is found for camera device - Id : 0. May be attempting to bind too many use cases. Existing surfaces: [SurfaceConfig{configType=PRIV, configSize=PREVIEW}, SurfaceConfig{configType=YUV, configSize=ANALYSIS}] New configs: [androidx.camera.core.impl.y0@b351737, androidx.camera.core.impl.q1@d784c36]
```

so unbindAll() camera provider before binding should be fix the problem

Tested on Google Pixel 3, Android 12
Flutter 2.23, Dart 2.12


juliansteenbakker:
Thanks for your contribution!

PR #40, opened 2022-03-04
Feature: Image picker

### LINKED ISSUE #10: Scan picture from gallery
Hi 👋🏼  Do you plan to add support for scanning a picture from gallery? Do you welcome PRs?

juliansteenbakker:
fixes #10


MkLHX:
Hello!
First, Thank you for this great package!

I'm trying to use QRCode scan from image using image picker.

I tried with this example:
https://github.com/juliansteenbakker/mobile_scanner/blob/master/example/lib/barcode_scanner_controller.dart

The analyzeImage method always return False.

How can I debug this?

The QR Code I'm trying to read works with the camera code side.

PR #55, opened 2022-03-12
Added support for analysis from file.

m-ma-dev:
Hi :)
Here is my proposal to support the analysis from file (Android/iOS/MacOS).
I saw that you already did it in the meanwhile, so I am making this PR to allow you to pick some parts (or even the whole).


juliansteenbakker:
Thank you for your contribution! After some review i think i will pull out the parts for macOS since that integration doesn't exists yet.

m-ma-dev:
Nice 😁

(Just in case, the new formats are supported by Apple Vision, not the Google MLKit).

Issue #28, opened 2022-02-25
Build failed with an Exception 


MuhammadFaizan980:
Following error is being thrown during compilation:

`C:\Users\Dell\Documents\flutter\.pub-cache\hosted\pub.dartlang.org\mobile_scanner-0.1.2\android\src\main\kotlin\dev\steenbakker\mobile_scanner\MobileScanner.kt: (27, 1): Class 'MobileScanner' is not abstract and does not implement abstract member public abstract fun onRequestPermissionsResult(p0: Int, p1: Array<(out) String!>, p2: IntArray): Boolean defined in io.flutter.plugin.common.PluginRegistry.RequestPermissionsResultListener
e: C:\Users\Dell\Documents\flutter\.pub-cache\hosted\pub.dartlang.org\mobile_scanner-0.1.2\android\src\main\kotlin\dev\steenbakker\mobile_scanner\MobileScanner.kt: (65, 5): 'onRequestPermissionsResult' overrides nothing
e: C:\Users\Dell\Documents\flutter\.pub-cache\hosted\pub.dartlang.org\mobile_scanner-0.1.2\android\src\main\kotlin\dev\steenbakker\mobile_scanner\MobileScanner.kt: (66, 66): Type mismatch: inferred type is Array<CapturedType(out String)>? but Array<(out) String!> was expected
e: C:\Users\Dell\Documents\flutter\.pub-cache\hosted\pub.dartlang.org\mobile_scanner-0.1.2\android\src\main\kotlin\dev\steenbakker\mobile_scanner\MobileScanner.kt: (66, 79): Type mismatch: inferred type is IntArray? but IntArray was expected`

MuhammadFaizan980:
Anyone facing above, I fixed the issue by switching to flutter stable and upgrading to latest, previously I was using master channel

espresso3389:
The actual cause of the issue is Flutter engine's recent change: [shell/platform/android/io/flutter/plugin/common/PluginRegistry.java](https://github.com/flutter/engine/blob/b5f48f56bd4e2f5a440b99337636835edef19832/shell/platform/android/io/flutter/plugin/common/PluginRegistry.java#L352).
They added `@NonNull` annotations on the function declaration.

#63 is just a PR for the issue but it's only for future referencing purpose.

For the moment, if someone working on master faces the issue, my fork has a workaround for it:

```yaml
  mobile_scanner:
    git: https://github.com/espresso3389/mobile_scanner
```

[...]

PR #63, opened 2022-03-15
For #28, we need a little change on method to override.

espresso3389:
Flutter master recently added many @NonNull annotations on interface/class definitions and it breaks method overrides.

I think the PR should not be merged until the change lands on stable though I want to share the PR for future referencing purpose.


juliansteenbakker:

Issue #64, opened 2022-03-15
[web] Bad state: stream has already been listened too


sarbogast:
On my web app, I have 2 different screens where I want to use MobileScanner. Both are stateful widgets where I initialize the controller in the state, and dispose it in the state's `dispose()` function.

When I use the scanner in the first screen, everything works fine. Then I switch to the other screen and I get this error:

```
======== Exception caught by flutter web plugins ===================================================
The following StateError was thrown during a framework-to-plugin message:
Bad state: Stream has already been listened to.

When the exception was thrown, this was the stack: 
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 251:49  throw_
dart-sdk/lib/async/stream_controller.dart 676:7                               [_subscribe]
dart-sdk/lib/async/stream_controller.dart 827:19                              [_createSubscription]
dart-sdk/lib/async/stream_impl.dart 473:9                                     listen
packages/flutter_web_plugins/src/plugin_event_channel.dart 136:38             _listen
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 45:50            <fn>
dart-sdk/lib/async/zone.dart 1685:54                                          runUnary
dart-sdk/lib/async/future_impl.dart 159:18                                    handleValue
dart-sdk/lib/async/future_impl.dart 766:44                                    handleValueCallback
dart-sdk/lib/async/future_impl.dart 795:13                                    _propagateToListeners
dart-sdk/lib/async/future_impl.dart 465:9                                     callback
dart-sdk/lib/async/schedule_microtask.dart 40:11                              _microtaskLoop
dart-sdk/lib/async/schedule_microtask.dart 49:5                               _startMicrotaskLoop
dart-sdk/lib/_internal/js_dev_runtime/patch/async_patch.dart 166:15           <fn>
====================================================================================================

======== Exception caught by services library ======================================================
The following MissingPluginException was thrown while activating platform stream on channel dev.steenbakker.mobile_scanner/scanner/event:
MissingPluginException(No implementation found for method listen on channel dev.steenbakker.mobile_scanner/scanner/event)

When the exception was thrown, this was the stack: 
dart-sdk/lib/_internal/js_dev_runtime/private/ddc_runtime/errors.dart 251:49  throw_
packages/flutter/src/services/platform_channel.dart 175:7                     _invokeMethod
[...]

PR #94, opened 2022-04-08
fix: bad state stream has already been listened to

EricRode:
This PR fixes Issue #64 by setting stream controller to broadcast. The issue occurred because one can only listen to a normal stream at once. If stream is a broadcast stream one can listen multiple times.


juliansteenbakker:

PR #104, opened 2022-04-17
Enhance documentation: barcode.rawValue can be null 

brandsimon:
Thank you for this library. I just tried to use it and flutter complained about `barcode.rawValue` can be null.
If there is a cleaner way (in dart) to make this work, please let me know and I am happy to adapt the PR.


juliansteenbakker:

Issue #11, opened 2022-02-22
support for barcode reading in specific regions


casvanluijtelaar:
I'm migrating a project from qr_code_scanner to this but I'm missing the support for searching in a specific camera area. I'd like use my custom overlay like show below and actually limit the scan area to any desired rect.

![54BD44A1-39FB-4D43-BAF1-8ED7C316ECE7](https://user-images.githubusercontent.com/67225314/155099542-5dbf55e3-e4af-42d1-9ec8-5431ea48e924.png)

thamarakshan-pilla:
hi, i would like to implement this scan region too.
Is this feature currently being worked upon?
If not, i would really love to help out

casvanluijtelaar:
> hi, i would like to implement this scan region too. Is this feature currently being worked upon? If not, i would really love to help out

I guess it's not so go for it ;)

thamarakshan-pilla:
I have created PR for this feature.
Please check it out and approve it.
Also this is my first ever open source contribution soo thank you for this opportunity :)

quiquegarcia:
> I have created PR for this feature. Please check it out and approve it. Also this is my first ever open source contribution soo thank you for this opportunity :)

Someone knows when this PR will be merged?

@thamarakshan-pilla where can we see your PR? maybe we can use a fork to test it

wxyash:
+1, would love this.

casvanluijtelaar:
It's being worked on [here](https://github.com/juliansteenbakker/mobile_scanner/pull/176), input is always helpful. the previous PR by thamarakshan-pilla is useless as it does not actually limit the scan area it just overlays some visuals.

juliansteenbakker:
With version 3.0.0-beta.4 you can use an custom scanning area like shown in the example app: https://github.com/juliansteenbakker/mobile_scanner/blob/master/example/lib/barcode_scanner_window.dart 

Please note that this is still a beta so there can be breaking changes in the near future.

Ccmi:
Hi
ScanWindow is wrong when fit:BoxFit.cover. Areas other than the left and right sides of the designated area can still be identified;

PR #123, opened 2022-05-07
Scan region

thamarakshan-pilla:
Hi, I have added the scan region which is customizable through user input arguments. It is enabled by default and can be disabled by the user. Please check it out and approve this PR because I really like this mobile scanner and I was planning on implementing it in one of my projects but the only thing it is missing is this scan region :)

Issue reference : #11 

Here is a screenshot of the scan region with the default values
Screenshot_1651964641
 
This is my first ever open source contribution soo I'm sorry if I have made any mistakes.
Thank you for this plugin :)

### DIFF
--- a/lib/src/custom_border.dart
+++ b/lib/src/custom_border.dart
@@ -1,0 +1,45 @@
+import 'package:flutter/material.dart';
+
+class BorderPainter extends CustomPainter {
+  BorderPainter({
+    required this.strokeWidth,
+    required this.boxWidth,
+    required this.boxHeight,
+    required this.strokeColor,
+    required this.cornerSide,
+  });
+
+  final double strokeWidth;
+  final double boxWidth;
+  final double boxHeight;
+  final Color strokeColor;
+  final double cornerSide;
+
+  @override
+  void paint(Canvas canvas, Size size) {
+    final Paint paint = Paint()
+      ..color = strokeColor
+      ..strokeWidth = strokeWidth
+      ..style = PaintingStyle.stroke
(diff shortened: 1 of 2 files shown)


thamarakshan-pilla:
I have only added this feature for non web option

WingCH:
1. It seems that only the UI part, If the qrcode is outside the scanarea, `onDetect` will still trigger

https://user-images.githubusercontent.com/19588517/167277283-9e071ada-6487-4ad3-b834-9cb1ee6749e3.mov

2.  scanarea ui size should not be affected by MobileScanner's BoxFit 

image

Screenshot_20220508-090237

casvanluijtelaar:
this is not what is requested in the issue. The issue describes a way to clip the input image before sending it to the detector. In my opinion the package should never include a pre made overlay widget.

juliansteenbakker:
Thank you for your contribution. I'm on vacation right now till 23 of may. I will check and merge when i'm back.

juliansteenbakker:
It seems that this PR only adds an overlay to the view, but does not restrict this. I think if we would implement this, we should also restrict the scan area, otherwise this would not make any sense to me.

casvanluijtelaar:
> It seems that this PR only adds an overlay to the view, but does not restrict this. I think if we would implement this, we should also restrict the scan area, otherwise this would not make any sense to me.

absolutely I would even argue against an integrated overlay.

juliansteenbakker:
I'm closing this for now, but feel free to reopen this once improvements has been made.

PR #150, opened 2022-05-22
Fix return value of analyzeImage

apoleo88:


juliansteenbakker:
Thanks for your contribution. I do however think its better to change `as bool` to `as Future<bool>` since that is the type.

Issue #158, opened 2022-05-24
Error when i build aar 


tibztibz:
flutter doctor
[✓] Flutter (Channel stable, 2.5.3, on Ubuntu 20.04.4 LTS 5.13.0-41-generic, locale fr_FR.UTF-8)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.1.0-rc1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2021.2)
[✓] Android Studio (version 2021.1)
[✓] Android Studio
[✓] VS Code (version 1.67.1)
[✓] Connected device (2 available)

Hello,

When i want build aar , i have this error :

`cd "/home/tibo/projets/itga/diaginboxmp/clients/diaginboxmp_module" && flutter build aar --no-profile --build-number=0.0.0 <

Building without sound null safety
For more information see https://dart.dev/null-safety/unsound-null-safety

Running Gradle task 'assembleAarDebug'...                          18,9s

../../../../../flutter/.pub-cache/hosted/pub.dartlang.org/mobile_scanner-1.1.1/lib/src/mobile_scanner.dart:50:5: Warning: The class 'WidgetsBinding' cannot be null.
Try replacing '?.' with '.'
    WidgetsBinding?.instance.addObserver(this);
    ^^^^^^^^^^^^^^
../../../../../flutter/.pub-cache/hosted/pub.dartlang.org/mobile_scanner-1.1.1/lib/src/mobile_scanner.dart:50:30: 
Error: Method 'addObserver' cannot be called on 'WidgetsBinding?' because it is potentially null.
 - 'WidgetsBinding' is from 'package:flutter/src/widgets/binding.dart' ('../../../../../flutter/packages/flutter/lib/src/widgets/binding.dart').
Try calling using ?. instead.
    WidgetsBinding?.instance.addObserver(this);
                             ^^^^^^^^^^^
../../../../../flutter/.pub-cache/hosted/pub.dartlang.org/mobile_scanner-1.1.1/lib/src/mobile_scanner.dart:134:5: Warning: The class 'WidgetsBinding' cannot be null.
Try replacing '?.' with '.'
    WidgetsBinding?.instance.removeObserver(this);
    ^^^^^^^^^^^^^^
[...]

Issue #156, opened 2022-05-24
Swift Compiler Error (Xcode): 'ISO8601DateFormatter' is only available in iOS 10.0 or newer


munkius:
Getting this when compiling (using Flutter 2.10.3). My `Podfile` is set to: `platform :ios, '12.0'`.

Everything has been working fine, my app compiles and runs perfectly until I add `mobile_scanner` as a dependency. Then I get this error:
```
Could not build the precompiled application for the device.
Swift Compiler Error (Xcode): 'ISO8601DateFormatter' is only available in iOS 10.0 or newer
/Users/munkius/.asdf/installs/flutter/2.10.3/.pub-cache/hosted/pub.dartlang.org/mobile_scanner-1.1.0/ios/Classes/Util.swift:92:15
2

Uncategorized (Xcode): Command CompileSwiftSources failed with a nonzero exit code
```

Flutter doctor output:
[✓] Flutter (Channel stable, 2.10.3, on macOS 12.3.1 21E258 darwin-x64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 32.0.0)
[✓] Xcode - develop for iOS and macOS (Xcode 13.4)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2020.3)
[✓] VS Code (version 1.67.2)
[✓] Connected device (3 available)
[✓] HTTP Host Availability

Originally mentioned in #133. Cc'ing @kyleorin, @Zelfapp as they're having the same issue.

juliansteenbakker:
In #133 i mentioned adding `platform :ios, '10.0'` at the top of the Podfile.

Could you also try to add the following to the bottom of the Podfile? This will make sure all deployment targets are set to iOS 10:

```Podfile
post_install do |installer|
  installer.pods_project.targets.each do |target|
    target.build_configurations.each do |config|
      config.build_settings['ENABLE_BITCODE'] = 'NO'
      config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '10.0'
    end
  end
end
```
If you already have a section with `installer.pods_project.targets.each do |target|` make sure to add the `config._build_settings` lines to that block since its not allowed to add this block twice.

munkius:
Doh, sorry for wasting your time here. I had this block, yet it was set to `9.0` instead of `10.0` or beyond.

That change did solve the main compilation issue, but did introduce some other failures. Let me try to test it out fully before concluding it works fine.

[...]

PR #162, opened 2022-05-26
Preventing compilation error when adding or removing a WidgetsBinding observer

munkius:
When trying to implement the fixes suggested in #156 I stumbled upon a compiler error with the `WidgetsBinding` observers. This is a simple fix to prevent that.

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -47,7 +47,7 @@ class _MobileScannerState extends State<MobileScanner>
   @override
   void initState() {
     super.initState();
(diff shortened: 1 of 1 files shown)


t-bolt:
I think this should also fix #158

juliansteenbakker:
Ok so there is a bit of a dilemma with this. If i add this PR then users with flutter 3.0.0 will get a warning. If i don't add it this plugin will not work for users that dont run Flutter 3.0.0.

What i'm going to do is add this back, release it as v1.1.2, and create a new version v2.0.0 which doesn't have this and only works for flutter 3.0.0 without warnings.

PR #180, opened 2022-06-09
add autoResume Option

yuchan2215:
Resuming the app also starts the camera, but this is inconvenient if the developer wants to explicitly control it.
This pull request fixes this.
I am using Deepl.


juliansteenbakker:

PR #185, opened 2022-06-10
refactor: applying lint's suggestions

h1376h:
Hey! Thanks a lot for your amazing package, this PR contains breaking changes and no changes in behavior and only applies most of suggested lints and [analysis options](https://github.com/flutter/flutter/blob/master/analysis_options.yaml) to make the code cleaner and more readable.

These two lints are ignored in most files:
`always_specify_types`, `strict_raw_type`

I can fix most cases of those two in another PR.


juliansteenbakker:
Thank you for your contribution, however i use the lint package for all my plugins. I do see that it can be stricter, however this is something that i would like to apply to all my plugins to keep it consistent.

PR #253, opened 2022-08-16
expose the binaryData from the jsQR library

JakubPatrik:
Hi!, 

currently the jsQR exposes `data` and `binaryData` properties of the object as a result of its `scan: QRCode` method.
However, the mobile_scanner only ever uses the data.

In case the QRCode payload does not return a stringified data property, it still does return `binaryData` which corresponds 1:1 to the `rawBytes` of the mobile_scanner Barcode object. Otherwise, if you scan a QR code which payload is only binary data, the mobile_scanner returns null.

PS: I'm not a jsQR expert, but it seems to me that logically binaryData will never be null, otherwise the detection itself failed?

Hence the PR, if you have any comments I'm happy to review them.

### DIFF
--- a/lib/mobile_scanner_web_plugin.dart
+++ b/lib/mobile_scanner_web_plugin.dart
@@ -200,7 +200,11 @@ class MobileScannerWebPlugin {
 
     final code = jsQR(imgData.data, canvas.width, canvas.height);
     if (code != null) {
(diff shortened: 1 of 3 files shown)


juliansteenbakker:
Thank you for your work! There are some merge conflicts. Can you fix those? After that i can merge it and release it.

JakubPatrik:
Sure thing, just resolved those :)

PR #256, opened 2022-08-23
ci: Add Release Please to auto-generate changelogs

teolemon:
### What
Hello,
Pierre from Open Food Facts (https://github.com/openfoodfacts/smooth-app)
We're currently investigating the whole scanning situation with Flutter, as well as plugins with CameraX and MLKit, and here's a proposal to generate changelogs for mobile_scanner

ci: Add Release Please to auto-generate changelogs
ci: Enforce semantic Pull Requests to make changelogs possible

### Part of
- #260

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -2,7 +2,7 @@
  on:
    push:
      branches:
-       - main
+       - master
  jobs:
    release-please:
      runs-on: ubuntu-latest
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,13 @@
+ name: Run release-please
+ on:
+   push:
+     branches:
+       - main
+ jobs:
+   release-please:
+     runs-on: ubuntu-latest
+     steps:
+       - uses: GoogleCloudPlatform/release-please-action@v3.3.0
+         with:
+           token: ${{ secrets.GITHUB_TOKEN }}
+           release-type: simple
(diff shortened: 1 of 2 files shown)


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,13 @@
...
+   push:
+     branches:
+       - main
+ jobs:
+   release-please:
+     runs-on: ubuntu-latest
+     steps:
+       - uses: GoogleCloudPlatform/release-please-action@v3.3.0
+         with:
+           token: ${{ secrets.GITHUB_TOKEN }}
```
juliansteenbakker, line 12: How is this gonna work since i don't have any GCP key configured?

.github/workflows/release-please.yml
```diff
@@ -0,0 +1,13 @@
+ name: Run release-please
+ on:
+   push:
+     branches:
+       - main
```
juliansteenbakker, line 5: This branch uses master as main branch, not main

juliansteenbakker (changes requested):
Thank you for your contribution. I do however have some questions about this.

.github/workflows/release-please.yml
teolemon: You don't need one, it's just a project to manage releases, which happens to be made by Google cloud folks, and here it's just the standard GitHub toke

.github/workflows/release-please.yml
teolemon, line 5:
```suggestion
       - master
```

.github/workflows/release-please.yml
juliansteenbakker: Thank you for the clarification! This seems like a good addition 😄

juliansteenbakker (approved):

PR #279, opened 2022-09-07
fix: Fix the error of requestCode value range.

ipcjs:
### What
- Fix the error of requestCode value range

### DIFF
--- a/example/android/app/src/main/AndroidManifest.xml
+++ b/example/android/app/src/main/AndroidManifest.xml
@@ -5,7 +5,7 @@
         android:label="mobile_scanner_example"
         android:icon="@mipmap/ic_launcher">
         <activity
-            android:name="io.flutter.embedding.android.FlutterActivity"
+            android:name=".MainActivity"
             android:exported="true"
             android:launchMode="singleTop"
             android:theme="@style/LaunchTheme"
--- a/example/android/app/src/main/kotlin/dev/steenbakker/mobile_scanner_example/MainActivity.kt
+++ b/example/android/app/src/main/kotlin/dev/steenbakker/mobile_scanner_example/MainActivity.kt
@@ -1,6 +1,6 @@
 package dev.steenbakker.mobile_scanner_example
 
-import io.flutter.embedding.android.FlutterActivity
+import io.flutter.embedding.android.FlutterFragmentActivity
 
-class MainActivity: FlutterActivity() {
+class MainActivity : FlutterFragmentActivity() {
 }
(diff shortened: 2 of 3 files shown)


ipcjs:
Can this PR be merged?

example/android/app/src/main/AndroidManifest.xml
```diff
@@ -5,7 +5,7 @@
         android:label="mobile_scanner_example"
         android:icon="@mipmap/ic_launcher">
         <activity
-            android:name="io.flutter.embedding.android.FlutterActivity"
+            android:name=".MainActivity"
```
juliansteenbakker, line 8: Why would you add main activity, if we can use io.flutter.embedding.android.FlutterFragmentActivity directly?

example/android/app/src/main/kotlin/dev/steenbakker/mobile_scanner_example/MainActivity.kt
```diff
@@ -1,6 +1,6 @@
 package dev.steenbakker.mobile_scanner_example
 
-import io.flutter.embedding.android.FlutterActivity
+import io.flutter.embedding.android.FlutterFragmentActivity
 
-class MainActivity: FlutterActivity() {
+class MainActivity : FlutterFragmentActivity() {
```
juliansteenbakker, line 5: What is the advantage of FlutterFragmentActivity over FlutterActivity? I can't really seem to find an answer for this online.

juliansteenbakker (commented):
Sorry for the late response. I have a few questions about these changes.

example/android/app/src/main/AndroidManifest.xml
ipcjs: Theoretically it is possible. But since we have created a `MainActivity`, why not use it?

juliansteenbakker: Fair point! I think its best to just remove the mainactivity file since it doesn't add anything else

example/android/app/src/main/kotlin/dev/steenbakker/mobile_scanner_example/MainActivity.kt
ipcjs: this is doc: https://api.flutter.dev/javadoc/io/flutter/embedding/android/FlutterFragmentActivity.html

I use `FlutterFragmentActivity` in my project mainly to circumvent the problem: https://github.com/flutter/flutter/issues/98133

example/android/app/src/main/AndroidManifest.xml
ipcjs: In the actual project `MainActivity` will generally have some code, but of course in this demo, it can be removed.

Do I need to delete it?

juliansteenbakker: Well its alright like this. Maybe its clearer for some people since its only he example app.

juliansteenbakker (approved):

PR #292, opened 2022-09-19
fix: removed WidgetsFlutterBinding.ensureInitialized() for WEB plugin

saiful-apx:
- This blocks error reporting via runZonedGuarded
- Also this is should be called before calling runApp method via main method


juliansteenbakker:

PR #304, opened 2022-09-26
fix: should check for the raw bytes instead of string, in case of a malformed utf8 string

wolfenrain:
If the QR code contains binary data that is not UTF8 (aka malformed code points) then the `rawValue` will be null. And then the if statement to check if `lastScanned` equals the `rawValue` will never return true as the value never gets set.

So instead of comparing the string representations I changed the code to compare the `Uint8List`, that way the first scan of a malformed QR code will succeed.

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -66,7 +66,7 @@ class _MobileScannerState extends State<MobileScanner>
     }
   }
 
-  String? lastScanned;
+  Uint8List? lastScanned;
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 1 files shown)


brandsimon:
I tested it, it fixed an issue I had with scanning binary qr-code (The scanner did not go away. The code worked fine with UTF-8 qr-codes). My current workaround is to use `allowDuplicates` and on the first successful scan, remove the Widget with `Navigator.of(context).pop();`.

juliansteenbakker (approved):

juliansteenbakker:

hellohejinyu:
image

`allowDuplicates: false` not working for me. I see the source code is comparing the `Uint8List`. But comparing two `Uin8List` always returns false, or am I missing something.🤔

image

@wolfenrain @juliansteenbakker

brandsimon:
@hellohejinyu 
you are right, I provided a PR to fix it. I don't know, why it fixed my problem.
Did you encounter the problem on Web or Mac? E.g. on Web, a new Uint8List is created, probably on Android MLKit returns the same List, if the value did not change?

PR #316, opened 2022-10-03
fix: Invalid argument: Instance of 'Ratio'

Apologiz:
fix: Unhandled Exception: Invalid argument: Instance of 'Ratio'

### DIFF
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -173,7 +173,7 @@ class MobileScannerController {
     // Set the starting arguments for the camera
     final Map arguments = {};
     arguments['facing'] = facing.index;
-    if (ratio != null) arguments['ratio'] = ratio;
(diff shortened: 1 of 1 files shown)


teolemon:
@Apologiz : merge conflict + small linting issue
Run flutter analyze
Analyzing mobile_scanner...                                     

  error • The property 'index' can't be unconditionally accessed because the receiver can be 'null' • lib/src/mobile_scanner_controller.dart:176:51 • unchecked_use_of_nullable_value

1 issue found. (ran in 17.2s)
Error: Process completed with exit code 1.

juliansteenbakker:
Thank you for your contribtion, but ratio has been removed in the latest version so this is no longer an issue.

Hauglid:
Which version is that? It's still "available" in 2.1.0 but crashes the app.

juliansteenbakker:
Version 3.0.0-beta.4. Will release a stable version next week after some testing.

PR #329, opened 2022-10-14
fix: fixed an issue where the lastScanned and new scanned value are n…

### LINKED ISSUE #328: The beta version is firing on duplicates
I've noticed that when using the beta version 3.0.0-beta.1, the duplicate check doesn't work properly.

This behaviour results in duplicate values being sent even though the initial setup was not to send anything.

vlazdra:
- Fixes #328


vlazdra:
The complete refactor overrides this implementation.

No longer relevant. Closing.

PR #345, opened 2022-10-31
fix: use listEquals to compare lists

brandsimon:


juliansteenbakker:
Thank you for your contribution! I am aware of this problem, however i am refactoring some code in the branch feature/return-image and i am working on implementing duplicate detection on the platform side instead of the flutter side in order to improve performance. I'll keep this PR open for now just in case.

juliansteenbakker:
I'm closing this since we use a different approach for detecting duplicates now.

PR #364, opened 2022-11-12
feat: Add scale API

WenheLI:
This PR adds the ability to adjust scale at flutter land.
Have tested on the Android side;
I got no iOS devices at this moment, if would be great if someone could help me test it out as well.


juliansteenbakker:
Thank you for your contribution. I will check and see if i can implement it in the latest version.

juliansteenbakker:
Thank you very much for your contribution! I've added several iOS improvements and are merging these in #399.

Issue #371, opened 2022-11-19
MobileScannerController.start() could use better error handling


navaronbracke:
The API for `MobileScannerController.start()` has some issues with usability.

1) Having to specify a separate function for `onPermissionSet` feels weird (not to mention its return type is a `dynamic Function(bool)` instead of `void Function(bool)`)
2) If start() throws an error, there is no robust way to determine what kind of error it is. We just get a `MobileScannerException` with a string message. If the message changes, our code might break.

I propose the following:
- deprecate the `onPermissionSet` argument on the controller
- add a new enum called `MobileScannerErrorCode` with enum values for the permission denied error & generic error
- replace the error message in `MobileScannerException` with an error code of the above enum

I'd be happy to implement this change if you want, its pretty small imo.

navaronbracke:
FYI: I have a PR in the works, I'll hopefully get it out tomorrow.

PR #373, opened 2022-11-20
fix: Fix start error handling

navaronbracke:
This PR updates the error handling for the `MobileScanner.start()` function.

It also updates some pieces of the public API and fixes several bugs.

See the updated changelog for an overview.

Fixes: #371


example/lib/main.dart
```diff
@@ -51,17 +50,6 @@ class MyHome extends StatelessWidget {
               child:
                   const Text('MobileScanner with Controller (returning image)'),
             ),
-            ElevatedButton(
```
navaronbracke, line 54: This one is no longer relevant, since the controller is now required.

lib/mobile_scanner.dart
```diff
@@ -2,11 +2,13 @@ library mobile_scanner;
 
 export 'src/enums/camera_facing.dart';
 export 'src/enums/detection_speed.dart';
+export 'src/enums/mobile_scanner_error_code.dart';
```
navaronbracke, line 5: I exported these so that users can handle errors properly. Previously the `MobileScannerException` was not exported.

lib/src/mobile_scanner.dart
```diff
@@ -1,52 +1,43 @@
+import 'dart:async';
+
 import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:mobile_scanner/src/mobile_scanner_controller.dart';
 import 'package:mobile_scanner/src/objects/barcode_capture.dart';
 import 'package:mobile_scanner/src/objects/mobile_scanner_arguments.dart';
 
-typedef MobileScannerCallback = void Function(BarcodeCapture barcodes);
```
navaronbracke, line 7: I removed these typedefs, since they add no value and the function can be inlined without formatting issues.

lib/src/mobile_scanner.dart
```diff
@@ -55,104 +46,94 @@ class MobileScanner extends StatefulWidget {
...
+            builder: (_, constraints) {
+              return SizedBox.fromSize(
+                size: constraints.biggest,
+                child: FittedBox(
+                  fit: widget.fit,
+                  child: SizedBox(
+                    width: value.size.width,
+                    height: value.size.height,
+                    child: kIsWeb
+                        ? HtmlElementView(viewType: value.webId!)
```
navaronbracke, line 121: Isn't using `HtmlElementView` / `Texture` directly unsafe? On the other hand, this plugin is not a federated plugin,
so there is no `MobileScannerPlatformInterface` / `mobile_scanner_android` `mobile_scanner_web` distinction.

Maybe you could do that in the future though. Then the platform interface has a method that creates a `Widget` which is one of the two depending on the platform.

package `video_player` does that too.

lib/src/mobile_scanner.dart
```diff
@@ -55,104 +46,94 @@ class MobileScanner extends StatefulWidget {
...
+              );
+            },
+          ),
+        );
       },
     );
   }
 
-  @override
-  void didUpdateWidget(covariant MobileScanner oldWidget) {
```
navaronbracke, line 135: This is no longer needed, since the controller cannot be null, and we use `widget.controller` directly.

lib/src/mobile_scanner.dart
juliansteenbakker: That would be a better approach, but for now i think it's working without problems.

lib/src/mobile_scanner.dart
```diff
@@ -1,52 +1,43 @@
...
-  /// Create a [MobileScanner] with a [controller], the [controller] must has been initialized.
+  /// Create a new [MobileScanner] using the provided [controller]
+  /// and [onBarcodeDetected] callback.
   const MobileScanner({
-    super.key,
-    required this.onDetect,
-    this.onStart,
-    this.controller,
-    this.autoResume = true,
+    required this.controller,
```
juliansteenbakker, line 35: I know that requiring a controller fixes some problems in the codebase, however i don't know if this is the right approach. For example, all Flutter widgets that have controller structure such as [PageView](https://api.flutter.dev/flutter/widgets/PageView-class.html) don't require a controller. So i would still rather want the controller to be optional. I understand that we would need a different error approach, but maybe we could rather implement a callback function which sends the error code back to the user instead of throwing an exception. What do you think about this?

lib/src/mobile_scanner.dart
```diff
@@ -1,52 +1,43 @@
...
+  /// Defaults to [BoxFit.cover].
+  final BoxFit fit;
 
-  /// Function that gets called when the scanner is started.
-  ///
-  /// [arguments] The start arguments of the scanner. This contains the size of
-  /// the scanner which can be used to draw a box over the scanner.
-  final MobileScannerArgumentsCallback? onStart;
+  /// The function that signals when new barcodes were detected by the [controller].
+  final void Function(BarcodeCapture barcodes) onBarcodeDetected;
```
juliansteenbakker, line 20: I think it's best to let it be `onDetect` since it are not necessarily barcodes but also QR codes and such which can be detected. I think it's better to change the class name of `BarcodeCapture` to something more generic than to rename this callback function.

juliansteenbakker (changes requested):
Thank you for your contribution! I think a lot of changes are good, but i still have a few concerns about the controller being required.

lib/src/mobile_scanner.dart
navaronbracke: I understand that you'd want the controller to be optional. I'll revert that change.

In terms of handling errors, I think that having to only use a `try/catch / catchError()` is a better public API
and it fits with the natural way of working with futures. Also callbacks _can_ get messy, so I'd rather avoid them.

lib/src/mobile_scanner.dart
navaronbracke: Good point, I'll revert this change.

navaronbracke: Done

lib/src/mobile_scanner.dart
navaronbracke: @juliansteenbakker I reverted the required controller, but I kept the error handling as is.

lib/src/mobile_scanner.dart
```diff
@@ -55,104 +47,136 @@ class MobileScanner extends StatefulWidget {
...
-        controller.dispose();
-        controller = widget.controller!;
+        _controller?.dispose();
+        _controller = widget.controller;
       }
     } else {
       if (widget.controller == null) {
-        controller =
-            MobileScannerController(onPermissionSet: widget.onPermissionSet);
+        _controller = MobileScannerController();
```
navaronbracke, line 135: @juliansteenbakker this controller is never started, is that intended?

navaronbracke:
@juliansteenbakker apparently, there is a bug in the example app where a controller is provided. I think it needs to rebuild when the permission is granted. (it also does not show anything when it is rejected) I'll get to fixing that.

Also, I see that when calling stop and the controller is already stopped, the platform code throws. Should we just ignore that case instead of throwing?

Lastly, I see that the example app uses a logging service, thats only in the example app, right?

navaronbracke:
@juliansteenbakker I'm going to split off the fixes unrelated to the error handling / permission bug into a separate PR, so that we can get those merged first.

After that I'll rebase this branch and hopefully we can fix the issue with permissions then.

juliansteenbakker:
>  I'm going to split off the fixes unrelated to the error handling / permission bug into a separate PR, so that we can get those merged first.

Sounds good to me!

navaronbracke:
I put those fixes in here https://github.com/juliansteenbakker/mobile_scanner/pull/375

juliansteenbakker:
Thanks for creating all these PR's. I don't have time this week anymore to check these, but i will check these first thing next week.

lib/src/mobile_scanner.dart
```diff
@@ -55,104 +52,109 @@ class MobileScanner extends StatefulWidget {
...
-                ),
-              ),
-            ),
-          );
+          return widget.placeholderBuilder?.call(context, child) ??
+              const ColoredBox(color: Colors.black);
         }
+
+        return ClipRect(
+          child: LayoutBuilder(
```
saibotma, line 131: Is the `LayoutBuilder` actually required? Why is using `MediaQuery` not sufficient anymore? I am asking because `LayoutBuilder` can't be used inside an `IntrinsicHeight` widget.

navaronbracke: You could use 

```
final Size size = MediaQuery.sizeOf(context);

final double screenWidth = size.width;
```
instead if you want.

The sole purpose of `LayoutBuilder` was to avoid using MediaQuery to get just the screen size.

Doing a regular `MediaQuery.of(context)` makes the widget rebuild if anything in `MediaQueryData` changes, i.e. `textScaleFactor`. Using the more specified methods on MediaQuery allows to use the correct aspect of the InheritedModel, therefor filtering rebuilds more efficiently.

See https://api.flutter.dev/flutter/widgets/InheritedModel-class.html

saibotma: Using `MediaQuery.sizeOf` sounds promising. Nevertheless, I actually don't understand why having it like this is not sufficient. Why is it actually necessary to wrap it in two `SizedBox`es and a `FittedBox`?

return ClipRect(
    child: kIsWeb
        ? HtmlElementView(viewType: value.webId!)
        : Texture(textureId: value.textureId!),
);

PR #375, opened 2022-11-22
fix: Misc fixes

navaronbracke:
This PR adds a few fixes for the mobile scanner.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -10,6 +10,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:7.3.1'
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
     }
 }
 
--- a/android/gradle/wrapper/gradle-wrapper.properties
+++ b/android/gradle/wrapper/gradle-wrapper.properties
@@ -1,0 +1,6 @@
+#Fri Jun 23 08:50:38 CEST 2017
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-all.zip
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -94,6 +94,20 @@ class MobileScannerController {
 
   bool? _hasTorch;
 
+  /// Returns whether the device has a torch.
+  ///
+  /// Throws an error if the controller is not initialized.
+  bool get hasTorch {
+    if (_hasTorch == null) {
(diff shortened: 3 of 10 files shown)


android/build.gradle
```diff
@@ -10,6 +10,7 @@ buildscript {
 
     dependencies {
         classpath 'com.android.tools.build:gradle:7.3.1'
+        classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
```
navaronbracke, line 13: @juliansteenbakker when I opened the native Android project, Gradle sync failed for me.
After adding this line (and the missing gradle-wrapper.properties) it worked again. Maybe you forgot to include this?

android/gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,6 @@
+#Fri Jun 23 08:50:38 CEST 2017
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+zipStoreBase=GRADLE_USER_HOME
+zipStorePath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-all.zip
```
navaronbracke, line 6: If gradle ever gets updated, the distribution url should be updated.

lib/src/mobile_scanner_controller.dart
```diff
@@ -94,6 +94,20 @@ class MobileScannerController {
 
   bool? _hasTorch;
 
+  /// Returns whether the device has a torch.
+  ///
+  /// Throws an error if the controller is not initialized.
+  bool get hasTorch {
```
navaronbracke, line 100: Since toggleTorch() now does nothing if you don't have a torch, I figured you'd need this getter.

lib/src/mobile_scanner_controller.dart
```diff
@@ -212,12 +226,16 @@ class MobileScannerController {
   ///
   /// Only works if torch is available.
   Future<void> toggleTorch() async {
-    if (_hasTorch == null) {
+    final hasTorch = _hasTorch;
```
navaronbracke, line 229: I don't know if you knew but `type promotion` for nullable class fields does not work yet in Dart. Putting it in a local variable fixes that.

navaronbracke:
@juliansteenbakker one thing I'm not sure about is the following:

If the user passes in a custom controller, should the `MobileScanner` widget not start & dispose the controller? In that case we could assume that a widget higher up the widget tree (which provided the controller) does that for us.

Also, in `didUpdateWidget()` if we create a new controller from scratch, shouldn't that controller be started?

android/build.gradle
juliansteenbakker: Never encountered an issue with it, but i've added it.

juliansteenbakker:
I've merged all changes in #400 because of the merge conflicts. Thanks again for all your work!

Issue #54, opened 2022-03-11
Migrate from jsQR to qr-scanner for full barcode support.


juliansteenbakker:
On web, jsQR is limited to QR codes only. We should migrate to qr-scanner or something similar to support other barcodes aswell.

tiamatthei:
Hi there!
Quick question, is there a specific reason that the StreamController in the mobile_scanner_web_plugin.dart file is not a broadcast StreamController. I'm having problems using the QR reader more than once in the same web app. Thanks in advance for your response.

EricRode:
The controller is since PR #94 a broadcast Streamcontroller

p-mazhnik:
`qr-scanner` (https://github.com/nimiq/qr-scanner) is also limited to QR codes

thblckjkr:
It looks like [this implementation](https://github.com/mebjas/html5-qrcode) has a better support for barcodes and cross platform browwsers.

TheOneWithTheBraid:
Maybe I can do this one day; I wrote the initial implementation in https://github.com/juliuscanute/qr_code_scanner/pull/306 too and actually promised to still improve it.

tashi146:
> It looks like [this implementation](https://github.com/mebjas/html5-qrcode) has a better support for barcodes and cross platform browwsers.

I have look into this library and build [simple_barcode_scanner](https://pub.dev/packages/simple_barcode_scanner) but it is not promising as you can see the issue [here](https://github.com/CodingWithTashi/simple_barcode_scanner/issues/1)

thblckjkr:
> but it is not promising as you can see the issue [here](https://github.com/CodingWithTashi/simple_barcode_scanner/issues/1)

If the underlying problem is specifically that *It can't read EAN-13 barcodes with leading Zero* https://github.com/mebjas/html5-qrcode/issues/361#issue-1075181567 I think it's very promising, I would argue that it's maybe a better implementation than the current one.

PR #376, opened 2022-11-22
feat(web): add zxing-js barcode reader

p-mazhnik:
Closes https://github.com/juliansteenbakker/mobile_scanner/issues/54

Replace `jsqr` library with [zxing-js](https://github.com/zxing-js/library) for full barcode support.  

As I mentioned in the [issue comment](https://github.com/juliansteenbakker/mobile_scanner/issues/54#issuecomment-1107658953), `qr-scanner` (https://github.com/nimiq/qr-scanner) is also limited to QR codes. 

I saw suggestions to use `html5-qrcode` (https://github.com/mebjas/html5-qrcode) in the issue but their API does not allow barcode decoding by providing the stream / video element directly to the API. They only work with an HTML element's `id`, which we can't use because the HTML element isn't connected to the HTML DOM yet, when we start decoding. And actually `html5-qrcode` library is a wrapper over `zxing-js`.

And this PR provides good abstractions so it will be easier to change smth later if needed.

PR features:
- abstract readers-related code to a separate `WebBarcodeReaderBase` class and mixins
- provide developers an ability to use their own reader by extending `WebBarcodeReaderBase` class and overriding `barCodeReader` property in `MobileScannerWebPlugin`
- we can still use `jsQR` library. 
Overriding example:
```dart
void main() {
  if (kIsWeb) {
    MobileScannerWebPlugin.barCodeReader =
        JsQrCodeReader(videoContainer: MobileScannerWebPlugin.vidDiv);
  }
  runApp(const MaterialApp(home: MyHome()));
}
```
- This PR adds `zxing-js` as a default library, but we can change it back to `jsQR`, keeping PR changes.

Note that I plan to take a look at torch support as a next step (different PR)


juliansteenbakker:
Thank you for all the PR's! I don't have time anymore this week to check and test it all, but i will make sure to check these next week.

MB175:
looking forward to that :D

p-mazhnik:
@MB175 
Did you add 
```html
<script type="text/javascript" src="https://unpkg.com/@zxing/library@0.19.1"></script>
```
to your index.html?

MB175:
> @MB175 Did you add
> 
> ```
> <script type="text/javascript" src="https://unpkg.com/@zxing/library@0.19.1"></script>
> ```
> 
> to your index.html?

Thanks, works :D

Stupid mistake on my side

juliansteenbakker:
Wow, great work! LGTM!

PR #391, opened 2022-11-29
fix: Update dispose method

EArminjon:
If developer give a controller, developer will dispose it. The package should not do it. Else we will have a double call to dispose.


EArminjon:
dispose call the stop method. Which create an error when you called stop before.

PR #393, opened 2022-12-02
fix: error black screen when request permission android

renanzdm:


renanzdm:
@juliansteenbakker

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -86,16 +87,10 @@ class MobileScanner(
     /**
      * Request camera permissions.
      */
-    fun requestPermission(permissionCallback: PermissionCallback) {
+    fun requestPermission() {
         listener
             ?: PluginRegistry.RequestPermissionsResultListener { requestCode, _, grantResults ->
```
juliansteenbakker, line 92: The ?: should be an =. I've changed this on the master branch and now everything is working.

juliansteenbakker:
I've fixed this on the master branch so this PR is no longer needed. Thank you however for looking into this and raising this issue.

renanzdm:
I thank you for the package, could you adjust this in  3.0.0-beta.2?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
renanzdm: I thank you for the package, could you adjust this in 3.0.0-beta.2?

PR #341, opened 2022-10-23
feat: scan-window-branch-conflict-fix

the-recondite:
Fix conflicts so that [juliansteenbakker:feature/scan-window](https://github.com/juliansteenbakker/mobile_scanner/tree/feature/scan-window) can be merged into master

PR #411, opened 2022-12-12
feat: add scan window overlay and area

juliansteenbakker:
Supersedes #341


flodaniel:
@juliansteenbakker I was just looking for this feature and if you brought it over from the old plugin! Awesome to see 🥳 
Any plans on when this will be released?

PR #429, opened 2022-12-19
fix: Ignore invalid textureId in captureOutput

zhengcan:
This PR checks textureId in beginning of captureOutput to avoid Unexpected exception in macOS.
In my demo app, it seems to work as expected.


patniemeyer:
I think this fix may also need to be applied in the corresponding place in `ios/Classes/MobileScanner.swift`.   Thanks.

PR #471, opened 2023-02-03
fix(web): fix loading js scripts in debug mode in flutter 3.7

p-mazhnik:
This should work with the lower Flutter version as well
We need these changes because of regression of https://github.com/flutter/flutter/issues/58428


juliansteenbakker:
LGTM!

PR #483, opened 2023-02-13
feat(iOS): Support the most recent camera device types

### LINKED ISSUE #468: Camera Discovery on iOS
Because the beta contains a re-write of the iOS logic, I think augmenting the camera discovery logic could be a great addition for that release.

`if #available(iOS 10.0, *) {
            device = AVCaptureDevice.DiscoverySession(deviceTypes: [.builtInWideAngleCamera], mediaType: .video, position: cameraPosition).devices.first
        } else {
            device = AVCaptureDevice.devices(for: .video).filter({$0.position == cameraPosition}).first
        }`
        
I believe that the current logic only allows for the discovery of the wide angle camera type, or wide angle camera type fallback. 

But, iPhone 12, 13, and 14 all have multiple physical back cameras represented by a special camera types that allow for this feature: "Automatic switching from one camera to the other when the zoom factor, light level, and focus position allow"

https://developer.apple.com/documentation/avfoundation/avcapturedevicetype?language=objc

So if this logic could be adapted to allow for the discovery of those camera types, scanning performance would likely be improved by the free camera negotiation logic provided by Apple, this would be especially true for close distance scanning where the wide angle lens can't focus as well (or at all past a certain distance).

CoryADavis:
closes #468

I didn't notice a scanning speed enhancement, but users can now focus and scan from closer distances on iPhone 12, 13, and 14. Specific to iPhone14, I was actually able to find barcodes of a particular size and distortion that required this feature in order to be scanned at all.

### DIFF
--- a/ios/Classes/MobileScanner.swift
+++ b/ios/Classes/MobileScanner.swift
@@ -122,7 +122,11 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
         textureId = registry?.register(self)
 
         // Open the camera device
-        if #available(iOS 10.0, *) {
+        if #available(iOS 13.0, *) {
+            device = AVCaptureDevice.DiscoverySession(deviceTypes: [.builtInTripleCamera, .builtInDualWideCamera, .builtInDualCamera, .builtInWideAngleCamera], mediaType: .video, position: cameraPosition).devices.first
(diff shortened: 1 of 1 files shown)


juliansteenbakker:
Looks good! A future wish from me is to make all these settings configurable so the user can decide which camera type he can use (if available), but this is a good start.

Issue #490, opened 2023-02-15
Recognation local picture error


xiaowei4554:
hello.I used mobile_scanner in my project and I just want to recognition a barcode form my local picture, and there is an error occure.It can't work, there is an error code:_CastError (type '_Map<Object?, Object?>' is not a subtype of type 'List<dynamic>' in type cast). It in file of mobile_scanner_controller.dart at line 338.It worked with camera scan.And then I download your demo and run it. The same error occurred,  I just want to know whether the return types of scan results obtained by these two methods are consistent，If not, please tell me the solution

loic-hamdi:
Following - Facing same issue

navaronbracke:
@xiaowei4554 Do you have a local sample to test with? Maybe there is indeed a case where the dat that we get back is not converted correctly.

juliansteenbakker:
I'm closing this as there is no activity, and newer versions of this plug are already released. If this still happens with the latest version let me know by commenting.

PR #492, opened 2023-02-16
fix: Load image from gallery

loic-hamdi:
Fix: [https://github.com/juliansteenbakker/mobile_scanner/issues/490](https://github.com/juliansteenbakker/mobile_scanner/issues/490)

### DIFF
--- a/ios/Classes/SwiftMobileScannerPlugin.swift
+++ b/ios/Classes/SwiftMobileScannerPlugin.swift
@@ -225,22 +225,22 @@ public class SwiftMobileScannerPlugin: NSObject, FlutterPlugin {
             if barcodes!.count <= 0 {
                 let event: [String: Any?] = ["name": "error", "message": "no cade"]
                 barcodeHandler.publishEvent(event)
-                result(false)   //add
+                result(false)  
             }else{
                 for barcode in barcodes! {
                     let event: [String: Any?] = ["name": "barcode", "data": barcode.data]
                     barcodeHandler.publishEvent(event)
-                    result(true)  //add
+                    result(true)  
                 }
             }
             
         } else if error != nil {
             barcodeHandler.publishEvent(["name": "error", "message": error?.localizedDescription])
-            result(false) //add
+            result(false) 
         }else{
             let event: [String: Any?] = ["name": "error", "message": "no cade"]
             barcodeHandler.publishEvent(event)
-            result(false) //add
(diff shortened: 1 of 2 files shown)


loic-hamdi:
@juliansteenbakker  I think there is still an issue (at least on iOS, didn't try on other platform) for 
```dart
 await controller.analyzeImage(imageFile.path);
 ```
It returns false even when a barcode is found

But my PR fixes the data type issue that was breaking

xiaowei4554:
OK，thanks. This error has been solved.
However, there is another problem.In the file of mobile_scanner_controller.dart.The method of analyzeImage always return a null value in iOS platform. I suppose that the method of channel in iOS always return a null value. So i check the file of SwiftMobileScannerPlugin.swift.I think the current way should be like this:

`private func analyzeImage(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
        let uiImage = UIImage(contentsOfFile: call.arguments as? String ?? "")
        
        if (uiImage == nil) {
            result(FlutterError(code: "MobileScanner",
                                message: "No image found in analyzeImage!",
                                details: nil))
            return
        }
        mobileScanner.analyzeImage(image: uiImage!, position: AVCaptureDevice.Position.back, callback: { [self] barcodes, error in
            if error == nil && barcodes != nil {
                if barcodes!.count <= 0 {
                    let event: [String: Any?] = ["name": "error", "message": "no cade"]
                    barcodeHandler.publishEvent(event)
                    result(false)   //add
                }else{
                    for barcode in barcodes! {
                        let event: [String: Any?] = ["name": "barcode", "data": barcode.data]
                        barcodeHandler.publishEvent(event)
                        result(true)  //add
                    }
                }
                
            } else if error != nil {
                barcodeHandler.publishEvent(["name": "error", "message": error?.localizedDescription])
                result(false) //add
            }else{
                let event: [String: Any?] = ["name": "error", "message": "no cade"]
                barcodeHandler.publishEvent(event)
                result(false) //add
            }
        })
//        result(nil) //no need
    }`

I fix this and run this fix-demo, it has a real value of true or false, not null.This way look like worked.I ask you to give a proper and correct solution to fix this problem. Thanks again.

loic-hamdi:
@xiaowei4554 thanks,   I tried your code and it works also on my side.
I will update the code and push it to a new PR for @juliansteenbakker 👍

PR #497, opened 2023-02-17
fix: Load image from gallery & iOS Swift callback

loic-hamdi:
Fix: [https://github.com/juliansteenbakker/mobile_scanner/issues/490](https://github.com/juliansteenbakker/mobile_scanner/issues/490)

Following discussion [https://github.com/juliansteenbakker/mobile_scanner/pull/492](https://github.com/juliansteenbakker/mobile_scanner/pull/492)


juliansteenbakker:
Thanks for bringing this to my attention, however i did not found this solution the right way to do this. I've fixed this in #512

PR #500, opened 2023-02-20
feat(iOS): Tweak iOS camera focus settings

jorgenpt:
- Disable face-driven autofocus on iOS 15.4+
- Enable continuous autofocus if available

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## NEXT
+
+Improvements:
+* [iOS] No longer automatically focus on faces
+
 ## 3.0.0
 This big release contains all improvements from the beta releases.
 In addition to that, this release contains:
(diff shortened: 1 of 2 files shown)


juliansteenbakker:
Thank you for your contribtion. I see however some changes in this code that where applied by a previous PR. So what i'm going to do is see if we can make the camera selection dynamic for the user.

jorgenpt:
@juliansteenbakker I did not mean to modify the list of devices considered, I simply meant to collapse the iOS 10.2 and 10.0 checks since they're no longer needed. I will update the PR to still use the same device list. (I don't think it's necessary to expose the specific cameras used -- I think the virtual "multi-camera" devices are good enough, as they're driven by zoom level.)

CoryADavis:
I tend to agree, I can't think of scenarios where you'd want to choose. But, the current discovery list should give you the best virtual/physical camera for the job no matter what iPhone you have.

I wasn't sure what iOS versions the library was intending to support when I kept the iOS 10/10.2 checks in the previous PR. We only support back to iOS 15 in my app actually.

jorgenpt:
> I wasn't sure what iOS versions the library was intending to support when I kept the iOS 10/10.2 checks in the previous PR.

The [mobile_scanner.podspace](https://github.com/juliansteenbakker/mobile_scanner/blob/master/ios/mobile_scanner.podspec) lists iOS 11 as the supported version, so that's what I went by. :)

EmreSURK:
> The [mobile_scanner.podspec](https://github.com/juliansteenbakker/mobile_scanner/blob/master/ios/mobile_scanner.podspec) lists iOS 11 as the supported version, so that's what I went by. :)

I believe dropping support for iOS 10 and 11 is fine.

PR #504, opened 2023-02-20
fix: Multiple-listener-registiration

### LINKED ISSUE #496: Multiple listener registration to barcodes when app resumes
The method **_startScanner** register a new listener subscription on **_controller.barcodes**; each time the app resume the **_startScanner** is invoked and a new subscription is created, by the old one is still presents.

Effect: the onDetect method will be called multiple time for same barcode

enesbayar:
This PR fixes #496


juliansteenbakker:

Issue #499, opened 2023-02-19
Some problems with 3.0.0


VeryazovSergey:
Hello! Thanks for great plugin! But I faced with some problems...

In January I finished developing and published first Android version of my app using version 3.0.0-beta4
It was not using autoStart feature of controller (autoStart: false) but functions in exact places to start and stop controller.
A month later I started to prepare iOS version, moved to 3.0.0 and found some problems:
- Scanner was working/active only with  autoStart: true - both in iOS and Android
- XCode gave me special warning: -[AVCaptureSession startRunning] should be called from background thread. Calling it on the main thread can lead to UI unresponsiveness
- App was crashing with the same message: "Thread Performance Checker: -[AVCaptureSession startRunning] should be called from background thread. Calling it on the main thread can lead to UI unresponsiveness"
- I had to disable some option of XCode to avoid this exception.
- And at final - the function of analyzing photo from gallery didn't work on iOS at all - checked same pictures on Android - all fine - but iOS version do not recognize them.

So that's it.
 I would provide any other additional information if needed.

princ3od:
Hey @VeryazovSergey, I have struggled for hours with your final point. And I discovered that the issue may be unrelated to this package. 
Actually, the MLKit can not read **low-quality** QR code images. View the MLKit docs [here](https://developers.google.com/ml-kit/vision/barcode-scanning/ios#input-image-guidelines).
<img width="906" alt="image" src="https://user-images.githubusercontent.com/17960926/232202818-8983c306-d533-47ba-b848-331770958e70.png">

In my case, my app has a feature to save QR codes for other people to scan. Initially, I saved the QR code to a lower-quality image to reduce file size. And this image can not be `analyzing`. After saving QR code to highest quality image, the app can scan the QR Code smoothly.

PR #508, opened 2023-02-23
fix: Initialize subscription for any autoStart value

flodaniel:
As also reported in https://github.com/juliansteenbakker/mobile_scanner/issues/499, and from our own usage of `mobile_scanner: 3.0.0`, I noticed that the `_barcodesSubscription` is not initialized when `autoStart: false` was used. 

This PR fixes this problem bei always initializing the subscription. This should not have any adverse effects as the actual `_controller.start()` call is still coming after the autoStart check.

VS-Code also applied default formatting for Markdown files, which I hope is no problem :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,153 +1,191 @@
+## 3.0.1
+
+Fixes:
+
+- Fixes a bug where scanning did not work at all when using `autoStart:false`
+
 ## 3.0.0
+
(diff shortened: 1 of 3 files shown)


juliansteenbakker:
Thank you for your contribution. However, there is another bug with this function #504, so what i'm going to do is merge that one first, and then move the function to above the autostart function to fix this bug.

PR #519, opened 2023-02-28
fix(web): waiting for js libs to load

### LINKED ISSUE #501: Wait for loading of js libraries
On web, if I use `MobileScanner` widget on first screen, I get an error seemingly caused by reader's js libraries not yet fully loaded:

e.g `Cannot read properties of undefined (reading 'BrowserMultiFormatReader')` with default zxing reader

Is there any way to be notified when libraries are loaded and widget is usable ?

p-mazhnik:
Fixes #501


juliansteenbakker:
Thanks !

Issue #426, opened 2022-12-15
Crashing on Android emulator


mentoc3000:
Steps to reproduce:

1. Run Android emulator
2. In the camera settings, load an image of a barcode
3. Run the example app
4. Tap MobileScanner with Controller
5. If necessary, move around the rooms to find the bar code
6. MobileScanner scans the barcode and then crashes

<details>
<summary>Console Log</summary>

Launching lib\main.dart on Android SDK built for x86 in debug mode...
√  Built build\app\outputs\flutter-apk\app-debug.apk.
Connecting to VM Service at ws://127.0.0.1:62541/dE_ZBgYp2Po=/ws
D/TransportRuntime.CctTransportBackend( 6256): Making request to: https://firebaselogging.googleapis.com/v0cc/log/batch?format=json_proto3
I/TransportRuntime.CctTransportBackend( 6256): Status Code: 200
I/TransportRuntime.CctTransportBackend( 6256): Content-Type: application/json; charset=UTF-8
I/TransportRuntime.CctTransportBackend( 6256): Content-Encoding: gzip
D/TransportRuntime.SQLiteEventStore( 6256): Storing event with priority=DEFAULT, name=FIREBASE_ML_SDK for destination cct
D/TransportRuntime.JobInfoScheduler( 6256): Scheduling upload for context TransportContext(cct, DEFAULT, MSRodHRwczovL2ZpcmViYXNlbG9nZ2luZy5nb29nbGVhcGlzLmNvbS92MGNjL2xvZy9iYXRjaD9mb3JtYXQ9anNvbl9wcm90bzNc) with jobId=-2079511396 in 854999ms(Backend next call timestamp 1671135499132). Attempt 1
I/CameraManagerGlobal( 6256): Connecting to camera service
D/CameraRepository( 6256): Added camera: 0
I/Camera2CameraInfo( 6256): Device Level: INFO_SUPPORTED_HARDWARE_LEVEL_LIMITED
D/CameraRepository( 6256): Added camera: 1
I/Camera2CameraInfo( 6256): Device Level: INFO_SUPPORTED_HARDWARE_LEVEL_FULL
D/CameraValidator( 6256): Verifying camera lens facing on generic_x86, lensFacingInteger: null
D/DeferrableSurface( 6256): Surface created[total_surfaces=1, used_surfaces=0](androidx.camera.core.SurfaceRequest$2@8a8444d}
[...]

PR #529, opened 2023-03-04
fix: Do not return image if returnImage is false / not specified.

michael-markl:
This PR avoids the error in #426 in the case `returnImage` is specified as null. Also, it actually fixes the intended behavior of the `returnImage` argument.

### DIFF
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -130,7 +130,7 @@ class MobileScannerController {
         arguments['formats'] = formats!.map((e) => e.index).toList();
       }
     }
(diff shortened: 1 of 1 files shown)


Alejandroem:
I'm having this same error on a package, when can we get this into the releases?

i-rme:
LGFM, this change solves the crash when scanning on the Android Emulator for me

temirlanzhangazy (approved):

tkomyakov:
please merge! and publish to pub.dev

Bobis1888:
please merge we need a fix!

juliansteenbakker:
Sorry for the late response, LGTM !

Issue #542, opened 2023-03-18
Crash In iOS when Switch Camera


ARASHz4:
In iOS when i want to switch camera to front app will crash
my mobile_scanner version is 3.1.0

PR #545, opened 2023-03-20
Fix: iOS front camera switching crash issue

### LINKED ISSUE #543: [iOS]: Front camera is not supported. App crashes when trying to use front camera.
Platform: iOS, iPhone 13 pro

`*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '*** -[AVCaptureDevice setAutomaticallyAdjustsFaceDrivenAutoFocusEnabled:] Not supported - use -isFocusModeSupported: to discover if autofocus is supported'
*** First throw call stack:
(0x190430e38 0x1895bf8d8 0x1a968b860 0x105e96f1c 0x105ea69d4 0x105ea54f0 0x105ea5a64 0x10c00fd18 0x10bac154c 0x197a47460 0x197a48f88 0x197a577f4 0x197a57444 0x1904c16c8 0x1904a302c 0x1904a7eb0 0x1ca69d368 0x19299d668 0x19299d2cc 0x104f52308 0x1aeda0960)
libc++abi: terminating with uncaught exception of type NSException
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
    frame #0: 0x00000001cdf4b674 libsystem_kernel.dylib`__pthread_kill + 8
libsystem_kernel.dylib`:
->  0x1cdf4b674 <+8>:  b.lo   0x1cdf4b694               ; <+40>
    0x1cdf4b678 <+12>: pacibsp
    0x1cdf4b67c <+16>: stp    x29, x30, [sp, #-0x10]!
    0x1cdf4b680 <+20>: mov    x29, sp
Target 0: (Runner) stopped.
Lost connection to device.
Exited`

CyberWake:
This PR Fixes #543 & #542

This issue was caused since `device.automaticallyAdjustsFaceDrivenAutoFocusEnabled` as being executed without checking if autofocus mode is present in the selected camera device. Which caused it to fail on fewer iOS modes.

Adding a check like `device.isFocusModeSupported(.autoFocus)` with the required iOS version was solution to the issue.

### DIFF
--- a/ios/Classes/MobileScanner.swift
+++ b/ios/Classes/MobileScanner.swift
@@ -138,7 +138,7 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
             if device.isFocusModeSupported(.continuousAutoFocus) {
                 device.focusMode = .continuousAutoFocus
             }
-            if #available(iOS 15.4, *) {
+            if #available(iOS 15.4, *) , device.isFocusModeSupported(.autoFocus){
                 device.automaticallyAdjustsFaceDrivenAutoFocusEnabled = false
             }
             device.unlockForConfiguration()


juliansteenbakker:
Thanks! LGTM!

PR #559, opened 2023-03-27
fix: Use MediaQuery.sizeOf(context) instead of topmost LayoutBuilder

navaronbracke:
This PR fixes an issue where `IntrinsicHeight` could not be used for the placeholder / error builder.

It also cleans up minor usages of the `Size` class.

- Addresses https://github.com/juliansteenbakker/mobile_scanner/pull/373#discussion_r1148913892

cc @saibotma Is this what you were referring to?

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -259,9 +259,8 @@ class _MobileScannerState extends State<MobileScanner>
                     size: constraints.biggest,
                     child: FittedBox(
                       fit: widget.fit,
-                      child: SizedBox(
-                        width: value.size.width,
-                        height: value.size.height,
+                      child: SizedBox.fromSize(
+                        size: value.size,
                         child: kIsWeb
                             ? HtmlElementView(viewType: value.webId!)
                             : Texture(textureId: value.textureId!),
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -232,45 +232,43 @@ class _MobileScannerState extends State<MobileScanner>
 
   @override
   Widget build(BuildContext context) {
-    return LayoutBuilder(
-      builder: (context, constraints) {
-        return ValueListenableBuilder<MobileScannerArguments?>(
-          valueListenable: _controller.startArguments,
-          builder: (context, value, child) {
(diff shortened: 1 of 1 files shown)


lib/src/mobile_scanner.dart
```diff
@@ -232,46 +232,43 @@ class _MobileScannerState extends State<MobileScanner>
...
-            }
-
-            if (widget.scanWindow != null && scanWindow == null) {
-              scanWindow = calculateScanWindowRelativeToTextureInPercentage(
-                widget.fit,
-                widget.scanWindow!,
-                value.size,
-                Size(constraints.maxWidth, constraints.maxHeight),
-              );
+    final Size size = MediaQuery.sizeOf(context);
```
navaronbracke, line 235: Replaced the topmost layout builder with `MediaQuery.sizeOf(context)`. This allows using the aspect of the InheritedModel, which filters rebuilds when the size changes.

navaronbracke:
@juliansteenbakker Apparently the `MediaQuery.sizeOf(context)` (and several of its friends) are not on stable yet.

I was running on the latest beta, hence why I didn't see it.

test/mobile_scanner_test.dart
```diff
@@ -7,12 +7,14 @@ void main() {
   TestWidgetsFlutterBinding.ensureInitialized();
 
   setUp(() {
-    channel.setMockMethodCallHandler((MethodCall methodCall) async {
```
navaronbracke, line 10: The old API is deprecated on the current stable.

navaronbracke:
@juliansteenbakker Not sure why that one action is failing? Running `dart format --set-exit-if-changed .` does nothing on my end. Your action still uses `flutter format` though, which is deprecated.

teolemon:
https://github.com/juliansteenbakker/mobile_scanner/pull/651

navaronbracke:
Do I push an empty commit to rerun the CI tasks? The last check still fails due to the deprecated format command.

navaronbracke:
@teolemon In your PR to use `dart format`, you missed removing the `-n` option, which is no longer accepted.

juliansteenbakker:
Sorry for the late response, LGTM!

PR #586, opened 2023-04-10
feat: Add overlay to scanner

svenopdehipt:
This PR adds an option to add an overlay to the scanner which is only visible when the scanner has started.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,8 @@ Improvements:
 * [Android] Updated com.google.mlkit:barcode-scanning from 17.0.3 to 17.1.0
 
 Bugs fixed:
-* Fixed onDetect not working with analyzeImage
-* [iOS] explicit returned type for compactMap
+* Fixed onDetect not working with analyzeImage when autoStart is false in MobileScannerController
+* [iOS] Explicit returned type for compactMap
 
 ## 3.1.1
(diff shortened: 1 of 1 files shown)


svenopdehipt:
Is something wrong with this pull request?

richard457:
@juliansteenbakker, please help in merging this PR.

navaronbracke:
@richard457 The only thing this PR does is wrap the mobile scanner widget in a stack, so I don't see the benefit of this PR.

svenopdehipt:
@navaronbracke this pr adds the option to add an overlay to the scanner which is only shown when there is no error and the scanner isn't loading

richard457:
Yes, I guess that is what was needed in qr_code_scanner we had which we were able to use like
```dart
QRView(
        key: qrKey,
        onQRViewCreated: (controller) {
          _onQRViewCreated(controller, model);
        },
        overlay: QrScannerOverlayShape(
            borderColor: const Color(0xFF375778),
            borderRadius: 10,
            borderLength: 30,
            borderWidth: 10,
            cutOutSize: scanArea),
        onPermissionSet: (ctrl, p) => _onPermissionSet(context, ctrl, p),
      )
```

richard457:
is there another way we can achieve the same?[navaronbracke](https://github.com/navaronbracke)

Issue #554, opened 2023-03-24
[iOS] Impossible to focus and scan the QR code due to picking the wide back camera


daniloapr:
The `mobile_scanner` is picking the wide back camera (0.5x zoom) instead of the standard one (1.0x) in some iOS devices, and that makes it impossible to read small QR codes as the camera looses focus. 

Is it possible to use the 1.0x camera as default or somehow allowing us to pick other available back cameras?  

We have tested in the following devices:
- iPhone 11: wide camera, could not scan
- iPhone 11 Pro: wide camera, could not scan
- iPhone 12: regular camera, could scan 
- Samsung S10: regular camera, could scan.

@juliansteenbakker Would you be willing to accept new PRs? I would appreciate if you could guide on how make this fix.

rvonkriegenbergh:
Is there any progress being made on this? Is it possible to replace the enum CameraFacing (front, back) with a method that detects all installed cameras on a device and lets us set the one we want to use? I've seen this in other camera/barcode packages, on a different platform though (Net MAUI).

AndreasLyngholm:
I am facing the same issues!

ewertonls:
I'm also having the same issue, but it also occurs on web in any Galaxy S

juliansteenbakker:
I'm closing this as there is no activity, and newer versions of this plug are already released. If this still happens with the latest version let me know by commenting.

PR #592, opened 2023-04-12
feat: Improvements about zoom scale value and ultra-wide camera

ryuta46:
This PR is for improvements zoom scale feature, especially for avoiding to select ultra wide camera unintentionally. Please review and merge if looks good.

## New features

* added `zoomScale` value notifier in MobileScannerController for the application to know the zoom scale value set actually.  
  The value is notified from the native SDK(CameraX/AVFoundation).

* added `resetZoomScale()` in MobileScannerController to reset zoom ratio with 1x.  
  Both Android and iOS, if the device have ultra-wide camera, calling `setZoomScale` with small value causes to use ultra-wide camera and may be diffcult to detect barcodes.
  `resetZoomScale()` is useful to use standard camera with zoom 1x.
  `setZoomScale()` with the specific value can realize same effect, but added `resetZoomScale` for avoiding floating point errors.
  The application can know what zoom scale value is selected actually by subscribing `zoomScale` above after calling `resetZoomScale`.

## Changed behaviors

* (iOS only) call `resetZoomScale` while starting scan.
  Android camera is initialized with a zoom of 1x, whereas iOS is initialized with the minimum zoom value, which causes to select the ultra-wide camera unintentionally (#554).
  Fixed this issue by calling `resetZoomScale`

* (iOS only) remove zoom animation with `ramp` function to match Android behavior.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -138,6 +138,7 @@ class MobileScanner(
         torch: Boolean,
         detectionSpeed: DetectionSpeed,
         torchStateCallback: TorchStateCallback,
+        zoomScaleStateCallback: ZoomScaleStateCallback,
         mobileScannerStartedCallback: MobileScannerStartedCallback,
         detectionTimeout: Long
     ) {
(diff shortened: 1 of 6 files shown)


juliansteenbakker:
Thanks! LGTM!

kgaulin:
hi, the title check is red. When this PR will be in a new version ? thank you

PR #603, opened 2023-04-20
fix: Add namespace to mobile_scanner for Android

### LINKED ISSUE #602: Gradle 8 support, Flutter 3.7.12+
https://github.com/flutter/flutter/issues/125181

Flutter app android module can't be upgraded to gradle 8, until all dependencies which use native code are also upgraded or at least have namespace in app/build.gradle.

android {
    namespace = "com.example.myapp"
    ...
}

navaronbracke:
This PR adds the `namespace` attribute to `mobile_scanner`'s build.gradle. The `package` attribute is left in the Manifest for now, as it is required for older projects that do not recognise the new namespace attribute.

The example app already had its namespace attribute set, so that is left as-is.

Fixes #602

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.2.1
+Bugs fixed:
+* [Android] Fix Gradle 8 compatibility by adding the `namespace` attribute to the build.gradle.
+
 ## 3.2.0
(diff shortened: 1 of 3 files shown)


juliansteenbakker:
Thanks!

Issue #627, opened 2023-05-15
Expected in: /usr/lib/libc++.1.dylib


GRSource:
When debug iOS11.0 and launch crash log:

**dyld: Symbol not found: __ZTISt18bad_variant_access
  Referenced from: /var/containers/Bundle/Application/3908D51F-C774-421C-93B9-A3502128F938/Runner.app/Runner
  Expected in: /usr/lib/libc++.1.dylib
 in /var/containers/Bundle/Application/3908D51F-C774-421C-93B9-A3502128F938/Runner.app/Runner**

and flutter doctor -v：
<img decoding="async" src="https://i.imgur.com/q3lPSZW.png" width="100%">

letrungdo:
I can reproduce too on iOS 11.2.6

letrungdo:
It work fine in version 2.1.0

juliansteenbakker:
I am closing all older issues. If this issue still exists in the latest version, please let me know.

PR #654, opened 2023-06-01
fix: crash in iOS 11

letrungdo:
Fix issue: https://github.com/juliansteenbakker/mobile_scanner/issues/627


juliansteenbakker:
Downgrading the MLKit version is not a solution

PR #655, opened 2023-06-01
fix: Remove -n option from format command

navaronbracke:
This fixes the format command in CI, as the `-n` option is not supported by `dart format`.
It also fixes the analysis failures in CI.

See also https://github.com/juliansteenbakker/mobile_scanner/pull/651#discussion_r1210305593

cc @teolemon

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,1 +1,7 @@
 include: package:lint/analysis_options_package.yaml
+
+linter:
+  rules:
+    - combinators_ordering
+    - require_trailing_commas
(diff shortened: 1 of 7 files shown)


analysis_options.yaml
```diff
@@ -1 +1,7 @@
-include: package:lint/analysis_options_package.yaml
\ No newline at end of file
+include: package:lint/analysis_options_package.yaml
+
+linter:
```
navaronbracke, line 3: @teolemon These rules were applied in CI, but not locally

PR #667, opened 2023-06-15
fix: Start running the capture session on a background thread

ened:
This removes a threading warning (and potentially jank).


juliansteenbakker:
Thanks! LGTM!

PR #669, opened 2023-06-19
refactor: Removed JS library injection which breaks CSP.

moria-euc:
Due to CSP requirements scripts cannot be loaded from remote urls. CSP compliant version of Flutter web app can be built using `--csp` flag.
This PR removes dynamic script injection and let user add library by his own choice - local or remote.
According changes about necessary setup steps are added to README.md.


juliansteenbakker:

PR #670, opened 2023-06-19
fix(ScanImage): fix android image result is not correct format and orientation

phanbaohuy96:
### What
- fix(ScanImage): fix android image result is not correct format and orientation


juliansteenbakker:
Thanks for your contribution!

PR #671, opened 2023-06-19
feat: Add overlay to scanner

svenopdehipt:
### What
- This PR adds an option to add an overlay to the scanner which is only visible when the scanner has started.
- Reopening of https://github.com/juliansteenbakker/mobile_scanner/pull/586 due to restructuring of fork.


juliansteenbakker:
LGTM ! Thanks for your contribution! I'm gonna change it a littlebit so that it doesn't use the `stack` when no overlay is given.

Issue #401, opened 2022-12-08
[feature] Convert the plugin to use `plugin_platform_interface`


navaronbracke:
Currently the plugin provides support for Android, iOS, MacOS and web.
However, the current implementation mixes `Texture` & `HTMLElementView` [directly in the implementation](https://github.com/juliansteenbakker/mobile_scanner/blob/master/lib/src/mobile_scanner.dart#L140-L142)
This can be improved by migrating the plugin to use the platform interface way of doing things.

This would also make implementing https://github.com/juliansteenbakker/mobile_scanner/issues/14 a bit easier, since only the windows implementation needs to be set up using the PlatformInterface method

TODO:
- [ ] introduce a platform interface
    - [ ] Widget buildScannerView()
    - [ ] The `MobileScannerArguments` class should be private, as it exposed implementation details
- [ ] reimplement web using `package:web`, which replaces `dart:html`
    - [ ] provide a web specific API to switch between barcode readers
    - [ ] migrate to Dart 3.3 extension types (currently not possible)
    - [ ] bring the camera handling improvements to the web    - [ ] start should throw if there is no camera
       - [ ] start should handle invalid camera direction in case the direction is not available
       - [ ] start should handle the torch state if the torch is not available
       - [ ] stop should do nothing if there is no camera
       - [ ] toggleTorch should do nothing if the current camera (if any) does not have a torch
- [ ] deprecate `hasTorch` and introduce a new `unavailable` torch mode
- [ ] remove lifecycle handling from the plugin, as it leads to lifecycle bugs #691 
   - [ ] remove the autoStart parameter
   - [ ] remove the deprecated onPermissionSet parameter
   - [ ] remove the `onStart` & `onScannerStarted` parameters
- [ ] refactor the `toggleTorch()`, `resetZoomScale()`, `setZoomScale()`, `stop()` methods to return instead of throw if the camera is null

navaronbracke:
@teolemon I would like to start working on this in the coming days.

My plan is to set up the plugin in such a way that we abstract implementation details behind the platform interface.
This would hide implementation details (such as webId vs textureId) and make fixing things a bit easier (I think)

navaronbracke:
@juliansteenbakker Just an FYI: I'll be tackling this one next.

[...]

PR #672, opened 2023-06-20
feat: Add option to disable automatic loading of JavaScript

svenopdehipt:
Add the option to disable automatic loading of JavaScript, which is necessary to conform to new EU-laws which don't permit to load from third party domains per the user agrees to (which he can't when it's happening on page load).

### DIFF
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -68,7 +68,7 @@ class MobileScannerController {
   static const EventChannel _eventChannel =
       EventChannel('dev.steenbakker.mobile_scanner/scanner/event');
 
-  /// Call this method to automatically load the javascript libraries, which are necessary for the plugin.
+  /// Call this method to initialize the plugin.
   static Future<void> initialize() async {
     if (kIsWeb) {
       await _methodChannel.invokeMethod('initialize');
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -68,6 +68,13 @@ class MobileScannerController {
   static const EventChannel _eventChannel =
       EventChannel('dev.steenbakker.mobile_scanner/scanner/event');
 
+  /// Call this method to automatically load the javascript libraries, which are necessary for the plugin.
+  static Future<void> initialize() async {
+    if (kIsWeb) {
+      await _methodChannel.invokeMethod('initialize');
+    }
+  }
+
   @Deprecated(
     'Instead, use the result of calling `start()` to determine if permissions were granted.',
   )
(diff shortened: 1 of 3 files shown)


navaronbracke:
`which is necessary to conform to new EU-laws` what specific law are you referring to here?

svenopdehipt:
https://blog.runcloud.io/google-fonts-gdpr/ This is an example with google fonts, but it also applies to other content which is loaded from 3rd-party servers. The law itself isn't new but the verdict is.

navaronbracke:
In that case I'd switch to using a locally bundled version of the library? (referring to the locally hosting of the asset)

svenopdehipt:
Yes, you have to host the library yourself

navaronbracke:
What I actually meant is that instead of disabling the automatic loading of the JS library, we change the loading of the library as follows:

- put the library JS sources under the `/web` folder in the plugin (like how other plugins do it when targeting the web)
- load the library from the asset in `/web`
- keep the behavior of automatically loading the library (now that it loads from the asset)

This way we are compliant with the law, and the consumers of the plugin don't have to do anything extra.

svenopdehipt:
I didn't know of this possibility. This would also solve the problem.

navaronbracke:
@svenopdehipt I had a quick look at it at it seems that my assumption was wrong.

If I add `./web/zxing.js` to the plugin and then try to load that, it does not work.
I get 

```
Refused to execute script from 'http://localhost:63940/zxing.js' because its MIME type ('text/html') is not executable, and strict MIME type checking is enabled.
```
due to the fact that the file is not found. It does not show up in the file system in Chrome devtools either. Probably because DDC/Dart2js does not include it in the build of the web app.

Also when looking at first party plugins (i.e. `google_maps_flutter`) they do include the script manually in their index.html

What if we move the loading of the library into an `initialize()` method? This method then would load the script in a script tag like how it happens now. Just on demand, instead of automatically.

I did the same for a different plugin for which I am adding web support.
In that scenario, a developer can implement asking permission before loading the library.

WDYT?

svenopdehipt:
I would have time to change the code today. The only question if the function shouldn't be placed in src/mobile_scanner_controller.dart instead of mobile_scanner_web.dart. When importing mobile_scanner_web.dart, which imports dart:html, I have to do a conditional import with a stub as a fallback to support dart vm.

navaronbracke:
Well, that is up to you to decide I think.

This package does not use the `plugin_platform_interface` package, the way it is supposed to.
That would eliminate the stub / conditional import. 

See also: #401

lib/src/mobile_scanner_controller.dart
```diff
@@ -68,6 +68,13 @@ class MobileScannerController {
   static const EventChannel _eventChannel =
       EventChannel('dev.steenbakker.mobile_scanner/scanner/event');
 
+  /// Call this method to automatically load the javascript libraries, which are necessary for the plugin.
```
navaronbracke, line 71: Same here, `JavaScript` does not apply when running on mobile / desktop.

PR #675, opened 2023-06-25
fix: Respect detectionTimeout on iOS

jorgenpt:
- Respect `detectionTimeout` on iOS devices, instead of arbitrarily waiting 10 frames
- Don't start a second scan until the first one is done, to keep memory usage more fixed if the device is slow

### DIFF
--- a/macos/Classes/MobileScannerPlugin.swift
+++ b/macos/Classes/MobileScannerPlugin.swift
@@ -26,6 +26,8 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     var scanWindow: CGRect?
     
     var detectionSpeed: DetectionSpeed = DetectionSpeed.noDuplicates
+
+    var timeoutSeconds: Double = 0
     
     
 //    var analyzeMode: Int = 0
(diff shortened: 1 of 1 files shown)


macos/Classes/MobileScannerPlugin.swift
```diff
@@ -26,6 +26,8 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     var scanWindow: CGRect?
     
     var detectionSpeed: DetectionSpeed = DetectionSpeed.noDuplicates
+
+    var timeoutSeconds: Double = 0
```
speszke, line 30: How would you find a little enhancement, introducing `TimeInterval` instead of raw Double? As a _typealias_ it doesn't change anything in the below logics but is more obvious that we are operating on seconds. 
```suggestion
    var timeout: TimeInterval = 0
```

macos/Classes/MobileScannerPlugin.swift
```diff
@@ -90,8 +92,8 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         return Unmanaged<CVPixelBuffer>.passRetained(latestBuffer)
     }
     
-    var i = 0
-    
+    var nextScanTime = 0.0
```
speszke, line 95:
```suggestion
    var nextScan: TimeInterval = 0.0
```

juliansteenbakker:
Thanks for the contribution!

macos/Classes/MobileScannerPlugin.swift
```diff
@@ -220,7 +224,9 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         let torch: Bool = argReader.bool(key: "torch") ?? false
         let facing: Int = argReader.int(key: "facing") ?? 1
         let speed: Int = (call.arguments as! Dictionary<String, Any?>)["speed"] as? Int ?? 0
+        let timeoutMs: Int = (call.arguments as! Dictionary<String, Any?>)["timeout"] as? Int ?? 0
         
+        timeoutSeconds = timeoutMs * 1000.0
```
juliansteenbakker, line 229: it should be divided instead of multiplied. Ive fixed it already.

Issue #666, opened 2023-06-15
java.lang.NullPointerException


lehoangkhang4869:
I see ARN error on my google play console (android 11 sdk 30): 
Exception java.lang.NullPointerException:
  at dev.steenbakker.mobile_scanner.MobileScanner.start$lambda-11$lambda-7 (MobileScanner.java)
  at androidx.camera.core.Preview.lambda$sendSurfaceRequest$1 (Preview.java)
  at android.os.Handler.handleCallback (Handler.java:938)
  at android.os.Handler.dispatchMessage (Handler.java:99)
  at android.os.Looper.loop (Looper.java:223)
  at android.app.ActivityThread.main (ActivityThread.java:7977)
  at java.lang.reflect.Method.invoke
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:603)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:948)
I looked in the project source, but I don't see the file MobileScanner.java, I only see MobileScanner.kt @@

rohansohonee1:
Stacktrace from firebase:
```
Fatal Exception: java.lang.NullPointerException:
       at dev.steenbakker.mobile_scanner.MobileScanner.start$lambda$10$lambda$6(MobileScanner.kt:173)
       at androidx.camera.core.Preview.lambda$sendSurfaceRequestIfReady$1(Preview.java)
       at android.os.Handler.handleCallback(Handler.java:942)
       at android.os.Handler.dispatchMessage(Handler.java:99)
       at android.os.Looper.loopOnce(Looper.java:226)
       at android.os.Looper.loop(Looper.java:313)
       at android.app.ActivityThread.main(ActivityThread.java:8757)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:571)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1067)
```

Device details:
```
Device
Brand:Samsung
Model:Galaxy Tab A8
Orientation:Portrait
RAM free: 765.36 MB
Disk free: 34.41 GB
Operating System
Version:Android 13
Orientation:Portrait
Rooted:No
Crash
Date:Jun 20, 2023, 4:32:14 PM
```

memoriasIT:
Can reproduce as well. This causes a crash on my app.
[...]

PR #710, opened 2023-08-06
feat: In the `SurfacePreview`, ensure the camera is not stopped

g123k:
Hello,

In some rare cases, the callback of the `SurfaceProvider` may be call after the camera is stopped.
In that case, `textureEntry` is null, generating a `NullPointerException` (#666)
 
This PR ensure that the camera is not stopped in the callback.


teolemon:
@juliansteenbakker We're about to release Open Food Facts with this fix. It would be a plus for us to be able to ship directly mobile_scanner vs our fork.

juliansteenbakker:
LGTM!

juliansteenbakker:
@teolemon i've just released v3.4.0.

PR #711, opened 2023-08-06
fix: macos build error

svenopdehipt:
Fix some macOS build errors:

- Date.now (https://developer.apple.com/documentation/foundation/date/3766590-now) is only available in macOS 12.0 and above while Date.init() (https://developer.apple.com/documentation/foundation/date/1780470-init) is available in macOS 10.10 and above
- Int must be explicitly converted to Double
- self must be specified explicitly to access imagesCurrentlyBeingProcessed in closure


juliansteenbakker:

PR #713, opened 2023-08-07
fix: Adapt app lifecycle states. In Android would the permission pop …

androi7:
The App Lifecycle States do not work correctly on Android. 

When the permission popup appears and the user denies the camera permission it would trigger the resumed lifecycle that would start the controller and invokes the permission pop up again.

The variable names are confusing because the inactive lifecycle does not relate with the _resumeFromBackground. But it needs to be true, because the controller is stopped anyway.

The code can be taken as reference and rewritten in a cleaner way.


juliansteenbakker:

Issue #715, opened 2023-08-07
 Build failed with an exception on 3.4.0


arnabonetraker:
Flutter Build Error with `package 3.4.0`
```
/mobile_scanner-3.4.0/lib/src/mobile_scanner.dart:251:34: Error: Member not found: 'MediaQuery.sizeOf'.
    final Size size = MediaQuery.sizeOf(context);
                                 ^^^^^^
Target kernel_snapshot failed: Exception


FAILURE: Build failed with an exception.
```

mustafayilmazdev:
Same here!

juliansteenbakker:
Fixed in v3.4.1

Issue #633, opened 2023-05-17
scanWindow not applied correctly


glemartret:
I'm trying to restrict the scan window to the size of the camera preview which is constrained by a `SizedBox` but out of screen barcodes (above and below my SizedBox) are still scanned.

Maybe I misuse `scanWindow` parameter ? Or is there an issue with the cover BoxFit ?

```
SizedBox(
	height: 250,
	width: double.infinity,
	child: MobileScanner(
		fit: BoxFit.cover,
		scanWindow: const Rect.fromLTWH(0, 0, double.infinity, 250),
		onDetect: (BarcodeCapture barcode) async {
			// ...
		}
	),
)
```

navaronbracke:
IIRC MLKit does not allow for specifying the scan area. You'll have to work with the `corners` property of the `Barcode` in the `BarcodeCapture` to figure out if the scanned barcode should be 'accepted'.

sdkysfzai:
I actually modified the Mobile Scanner widget to make it work.

sdkysfzai:
I had the same issue, I fixed it by doing modification to the package itself. Lmk if you need it and I'll make a fork and update it.

sdkysfzai:
> I'm trying to restrict the scan window to the size of the camera preview which is constrained by a `SizedBox` but out of screen barcodes (above and below my SizedBox) are still scanned.
> 
> Maybe I misuse `scanWindow` parameter ? Or is there an issue with the cover BoxFit ?
> 
> ```
> SizedBox(
> 	height: 250,
> 	width: double.infinity,
> 	child: MobileScanner(
> 		fit: BoxFit.cover,
> 		scanWindow: const Rect.fromLTWH(0, 0, double.infinity, 250),
> 		onDetect: (BarcodeCapture barcode) async {
> 			// ...
> 		}
> 	),
> )
> ```

https://github.com/sdkysfzai/mobile_scanner_upgrade
You can use this, it fixes this issue.

holailusoria:
Looks like scanWindow still doesn't work.
If I don't assign anything to this parameter, I can scan any area of the screen, which is to be expected.
As soon as I assign any value, I can't scan at all.

k435467:
+1 😢
[...]

PR #721, opened 2023-08-08
fix: Reworked scan window calculation

MBulli:
I'm using this pub in my app and found that the scanWindow parameter is not working correctly even in the latest version.
Imho the calculation of the relative scan window for the native code is not correct.

Note that the v3.3 implementation works fine as long as BoxFit.contain is used (like in the example). But in my use case, I'm displaying only a small part of the whole scan image using BoxFit.cover, where the calculation of the relative scan window fails.

I think in v3.4 the changes from @sdkysfzai improved the situation but is not correct for every BoxFit value. 

I started writing my changes based on the v3.3 version. For this reason, I wrote a completely new implementation instead of tweaking @sdkysfzai's code.
Also, the changes in 96af70a248ee0c494f32dbe1205de4b24417cf2c ended up in v3.4, which also broke my implementation. My code needs the widget size but MediaQuery returns the screen size. That's why I reverted that change. I don't know the reason why the LayoutBuilder was removed from the widget tree, maybe @navaronbracke can explain it. 

In this PR I added unit tests. At the moment the test cases don't cover every possible situation, so this still needs some more work, but, I started with the values from my use case. The expected values are based on svg images which that I created with Inkscape. If you want I can share those files.

Finally, I think that some part of the calculation should be refactored out into a separate public method so that the barcode corners in ``BarcodeCapture`` can be transformed from scan image coordinates to widget coordinates, which I would consider useful.

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -6,6 +6,7 @@ import 'package:mobile_scanner/src/mobile_scanner_controller.dart';
 import 'package:mobile_scanner/src/mobile_scanner_exception.dart';
 import 'package:mobile_scanner/src/objects/barcode_capture.dart';
 import 'package:mobile_scanner/src/objects/mobile_scanner_arguments.dart';
+import 'package:mobile_scanner/src/scan_window_calculation.dart';
 
 /// The function signature for the error builder.
 typedef MobileScannerErrorBuilder = Widget Function(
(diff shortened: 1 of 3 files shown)


MBulli:
see #633

juliansteenbakker:
Thank you for your contribution! Always nice to see some tests being added 😄

teolemon:
Run dart format --set-exit-if-changed .
Formatted test/scan_window_test.dart
Formatted 34 files (1 changed) in 0.7[9](https://github.com/juliansteenbakker/mobile_scanner/actions/runs/5796878108/job/15721330832?pr=721#step:5:10) seconds.
Error: Process completed with exit code 1.

navaronbracke:
@MBulli I removed the use of the LayoutBuilder, merely because somebody complained about not being able to use it in a specific scenario. I do not recall what the scenario was exactly.

I think the original reporter probably got around it by wrapping in an `Align` (which is what I would do in a Sliver for example, since LayoutBuilders don't directly work there)

I agree on reintroducing it (as it is more performant, since the subtree is usually smaller).

As a plus, this also makes the reasoning behind https://github.com/juliansteenbakker/mobile_scanner/issues/715 redundant again.

jlin5:
@MBulli @juliansteenbakker Any updates?

navaronbracke:
@MBulli still has to to a `dart format`, as one check is failing due to formatting changes.

juliansteenbakker:
I'm on it, sorry for the delay. I'm testing it and will merge it asap @EAminjon

MBulli:
@navaronbracke @juliansteenbakker  Sorry, just caught up on it.

EArminjon:
[heart]         enguerrand arminjon reacted to your message:
________________________________

navaronbracke (changes requested):

navaronbracke:
I'm fine with the nits to land this ASAP. I can address this later in a further cleanup.

navaronbracke:
I will cherry-pick @MBulli his commits into a new branch and apply the nits myself, so that we can land this patch today onto master.

MBulli:
@navaronbracke Thanks!

PR #725, opened 2023-08-08
Mac os fixes

ryanduffyne:
These fixes will handle limiting the bar code types to the types passed in, and then making sure to return the type that was found as part of the object response.

It also fixes a bug where the time to next scan was getting multiplied by 1000 when it should be divided by 1000.  Otherwise you are adding in the case of a default value of 250 milliseconds 250,000 seconds to the next scan time instead of .25 seconds like it should be.

### DIFF
--- a/macos/Classes/MobileScannerPlugin.swift
+++ b/macos/Classes/MobileScannerPlugin.swift
@@ -28,6 +28,8 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     var detectionSpeed: DetectionSpeed = DetectionSpeed.noDuplicates
 
     var timeoutSeconds: Double = 0
+  
+  var symbologies:[VNBarcodeSymbology] = []
     
     
 //    var analyzeMode: Int = 0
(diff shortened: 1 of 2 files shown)


macos/Classes/MobileScannerPlugin.swift
```diff
@@ -109,45 +112,53 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+                        }
+                    })
+                    if(self?.symbologies.isEmpty == false){
+                        // add the symbologies the user wishes to support
+                        barcodeRequest.symbologies = self!.symbologies
+                    }
+                    try imageRequestHandler.perform([barcodeRequest])
                 } catch {
-                  print(error)
+                    print(error)
```
navaronbracke, line 159: Same here

ryanduffyne:
I have made the requested fixes.

navaronbracke (approved):
LGTM. We could dedupe the two implementations of `isBarcodeInScanWindow()`, but that can be done for a cleanup PR.

Issue #720, opened 2023-08-08
Crash on iOS with the latest version (3.4.x)


g123k:
Hi everyone,

On iOS, I have the following crash when returning to a Flutter application:

```
mobile_scanner/MobileScanner.swift:224: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value
mobile_scanner/MobileScanner.swift:224: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value
* thread #1, queue = 'com.apple.main-thread', stop reason = Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value
    frame #0: 0x00000001a6554a2c libswiftCore.dylib`_swift_runtime_on_report
libswiftCore.dylib`:
->  0x1a6554a2c <+0>: ret    
libswiftCore.dylib`:
    0x1a6554a30 <+0>: b      0x1a6554a2c               ; _swift_runtime_on_report
libswiftCore.dylib`:
    0x1a6554a34 <+0>: adrp   x8, 366461
    0x1a6554a38 <+4>: ldrb   w0, [x8, #0x994]
Target 0: (Runner) stopped.
```

g123k:
@ened Hi!

Your PR https://github.com/juliansteenbakker/mobile_scanner/commit/cd506ebde87ad3219071cb2308b2ba44a64fd205 introduces this crash. Any idea how to fix it?

ened:
Looking at it now..

ened:
I added a PR #726 but unfortunately did not test it - if you could take a look (`dependency_overrides`) and retest it would be great. @g123k

g123k:
> I added a PR #726 but unfortunately did not test it - if you could take a look (`dependency_overrides`) and retest it would be great. @g123k

Thanks for your quick changes, but even by setting IPHONEOS_DEPLOYMENT_TARGET to 15, I have the following error:

```
Could not build the precompiled application for the device.
Swift Compiler Error (Xcode): Concurrency is only available in iOS 13.0.0 or newer
/Users/g123k/.pub-cache/git/mobile_scanner-3baf60b2ab6ba09c91029ae62677dc20a63c0562/ios/Classes/MobileScanner.swift:122:189

Swift Compiler Error (Xcode): Concurrency is only available in iOS 13.0.0 or newer
[...]

PR #726, opened 2023-08-09
WIP Switch to iOS 13, use Tasks to handle threads and add error handling

ened:


ios/Classes/MobileScanner.swift
```diff
@@ -197,36 +197,41 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
...
             }
         }
+        
+        let dimensions = CMVideoFormatDescriptionGetDimensions(self.device.activeFormat.formatDescription)
+        
+        if self.device == nil {
+            throw MobileScannerError.noCamera
+        }
+
+        return MobileScannerStartParameters(
```
navaronbracke, line 229: Does this return on the main dispatch queue still?

ened:
Happy to take another look. Closed it due to bug fix already in repo (for the iOS crashes).

navaronbracke:
@ened Is this still on your radar? This might help with fixing another case of #720 occurring (not sure though).

Unfortunately, there have been some changes to the iOS side of the plugin that caused some merge conflicts.
However, adding the `Task` back in for iOS should be trivial.

I can take a look at this as well if needed.

juliansteenbakker:
Hi all, since this PR is stale, i am going to merge it in to one of our branches so we don't alter the origin of @ened. This plugin still needs to support iOS 12, however async does support iOS 12. So i am going to change and test this, and merge if everything works out fine. Source: https://developer.apple.com/documentation/network/nwprotocolframer/instance/3200472-async

Thanks for your work @ened !

PR #748, opened 2023-09-09
build: Update Google Play Services and CameraX

weitsai:
Update Google Play Services dependency to 18.3.0 and CameraX to 1.3.0-rc01


juliansteenbakker:
Thanks for the PR, but im waiting till the stable release of camera2 instead of the RC variant.

Issue #745, opened 2023-09-08
The widgetSize changes affecting ...RelativeToTextureInPercentage?


indeedshaw:
Before version 3.4.0, widgetSize could be retrieved exactly. With the latest update, widgetSize becomes the screen size?
In my use case, the scanning page is not the screen size, which seems caused the scanning window to not contain QR codes well, so I thought, At this time the results calculated by calculateScanWindowRelativeToTextureInPercentage or do we expect?
[https://github.com/juliansteenbakker/mobile_scanner/blob/70f71a7956d7bca8d03e77c5ad3338b38e39cb7f/lib/src/mobile_scanner.dart#L251C51-L251C51](url)
`final Size size = MediaQuery.of(context).size`

example case:
mobile_scanner: ^3.4.1
`SizedBox(
                height: 500,
                child: MobileScanner(
                  onDetect: (BarcodeCapture capture) =>
                      _handleOnDetect(capture),
                  controller: _controller,
                  scanWindow: Rect.fromCenter(
                      center: Offset((screenWidth - 300) / 2,
                          (screenHeight - 300) / 2),
                      width: 300,
                      height: 300),
                ),
              ),`

indeedshaw:
Although the way things are computed today is a bit different than before, the idea is really brilliant
[https://github.com/juliansteenbakker/mobile_scanner/pull/176](url)

indeedshaw:
That's a brilliant idea
[https://github.com/juliansteenbakker/mobile_scanner/pull/176#issue-1259922838](url)

jlin5:
I'm experiencing the same issue as well with the same use case. It looks like **LayoutBuilder** was removed to solve an issue where **IntrinsicHeight** could not be used for the placeholder / error builder. https://github.com/juliansteenbakker/mobile_scanner/pull/559

Can the **LayoutBuilder** be added back to get the correct widget size? Otherwise the **calculateScanWindowRelativeToTextureInPercentage** method will have to change it's calculation based on the screen size instead of the widget size.

@juliansteenbakker @navaronbracke

navaronbracke:
@navaronbracke I can revert my original change so that a LayoutBuilder is used.

I'll have to test if I can reproduce the issue with the IntrinsicHeight, maybe just wrapping said widget in an Align fixes that issue as well?

jlin5:
[...]

PR #753, opened 2023-09-18
fix: Add back LayoutBuilder to calculate the correct widgetSize

jlin5:
- Added back the LayoutBuilder widget to get the correct values for the widgetSize instead of the screen size.
- Pull request for https://github.com/juliansteenbakker/mobile_scanner/issues/745
- This pull request might not be needed since there is another pull request that would resolve the issue https://github.com/juliansteenbakker/mobile_scanner/pull/721


EArminjon:
@juliansteenbakker :), little PR

juliansteenbakker:
Thanks for the work!

Issue #752, opened 2023-09-14
Feature Request: Add Functionality to Change Camera Quality


freidev:
Hello,

First of all, thank you for creating this useful package. It has been very helpful for my project.

I have a feature request that I believe could make the package even better. Currently, it seems like there is no option to change the camera quality or resolution when using mobile_scanner.

This feature would be extremely beneficial for applications that require high-quality scanning or image capture. Providing an API to adjust the camera resolution could improve the user experience and the versatility of your package.

Here's how I envision the API might look:

MobileScannerController( cameraResolution: ResolutionPreset.high);
Or maybe as a method to change it dynamically:
mobileScannerController.setCameraResolution(ResolutionPreset.high);

I understand that maintaining an open-source project is a lot of work, and I appreciate the effort you've put into this package. I would be thrilled to see this feature implemented and am looking forward to your thoughts on this request.

Thank you for considering this feature!

Best regards,
Freilyn Bernabe

EArminjon:
Currently this package have trouble to scan the following barcode. Sadly this barcode is wide and 'easy'.
![Capture d’écran 2023-09-26 à 09 35 07](https://github.com/juliansteenbakker/mobile_scanner/assets/37028599/02ca0ac4-e361-4802-ae43-df63b864c607)

I could understand that the following barcodes can sadly fail but we make fix that. To make it possible we need to have a way to improve the scan image quality @juliansteenbakker  :/ !
![Capture d’écran 2023-09-26 à 09 36 15](https://github.com/juliansteenbakker/mobile_scanner/assets/37028599/81548b78-2345-4a32-8faf-03db0dfc692b)

Issue #749, opened 2023-09-11
Scanner returns different values from the same barcode


AliEasy:
When scanning barcodes, each time it returns a different value.

Plugin mobile_scanner 3.4.1 on flutter_windows_3.10.2-stable

Code:
```
             MobileScanner(
                fit: BoxFit.fitWidth,
                controller: MobileScannerController(
                  detectionSpeed: DetectionSpeed.noDuplicates,
                  torchEnabled: false,
                ),
                onDetect: (capture) async {
                  if (!_inProcess) {
                    _inProcess = true;
                    final List<Barcode> barcodes = capture.barcodes;
                    if (barcodes.isNotEmpty) {
                      var barcode = barcodes[0];
                      String serial = barcode.rawValue ?? '';
                      print(serial);
                    }
                    _inProcess = false;
                  }
                },
              )
```

Barcode:
![code](https://github.com/juliansteenbakker/mobile_scanner/assets/45623453/2545074a-fe20-48e0-bad6-22cc418b0cd0)

results:
987654321098
981654327098

gigafunk:
reading off paper or a screen? I notice sometimes it will return some wrong numbers when reading off a digital screen of some sort, usually while its still kind of farther away. I was thinking it was just noise from the movement of the code into a "good spot".  I found it to be far more accurate off of real printed barcode.

AliEasy:
It's happening on both paper and screen

I get couple of different values by scanning  below barcode on a real product:

![1000006610.jpg](https://user-images.githubusercontent.com/45623453/266997768-ecefb6e5-74ad-41f4-8152-230fd58daef2.jpg)

My device is Galaxy A51 but also tested on another android device and still happening

EArminjon:
Maybe this is related to the image quality given to ML Kit. This package seems to give to ml Kit an image of 640/480 as default resolution...

plkgq:
Try setting sessionPreset to AVCaptureSession.Preset.medium and it will be recognized

Issue #674, opened 2023-06-23
Feature request: Customize analysis/preview camera resolution


nblum37:
Hi,

for especially bigger QR-Codes the package struggles to detect them. Currently, the internal analysis and preview resolution are not bound to any constraints or suggestions. At most of the testest smartphones, this can lead to an internal camera resolution of 640 x 480, being too small to detect the high-density QR-Codes. It would be nice to have a way to suggest a resolution like the ones recommended in the MLKit documentation. I would really appreciate it if someone would consider this feature request. Own attempts to create a PR failed due to strange aspect ratio behavior of the preview.

EArminjon:
This feature could fix https://github.com/juliansteenbakker/mobile_scanner/issues/749

gigafunk:
i can see it being the a possible fix for (https://github.com/juliansteenbakker/mobile_scanner/issues/645)  

Theory is (you should skip this if you value your time) the difference in aspect ratio from ios camera (16:9 for my test device) to the aspect ratio of 640 x 480 (4:3) interface in a way to  cause the  perceived barcode image ratios  to be way out of the limits (far to long for the height, making the low height codes way too long) for the barcodes the ml kit was trained on.  Turning the phone 90 degrees will shift the ratio warp the other direction ,  making the low height barcode taller and more in line with the mlkit training data.



I am making this all up, but I have myself convinced. I don't even need a preview, it would be nice just to have it default to 1280x720 or even a lower one, 640x360,  or 960x540, anything 16:9 would probably improve all these little weird issues.

PR #612, opened 2023-05-05
fix: adding back in setTargetResolution

bswhite1:
- Fixes the scan window on Samsung A23. Also verified it works on Moto G Play, Moto G Power (2021), and Pixel 6.
- Pull request for https://github.com/juliansteenbakker/mobile_scanner/issues/604

PR #760, opened 2023-09-26
fix: increase-camera-quality

EArminjon:
ML Kit need to have image in good quality else it will not be able to process images with precision. Bad precision can lead to bad code recognition and produce wrong detection.

This PR can maybe replace this PR and fix the aspect ratio issue mentioned : https://github.com/juliansteenbakker/mobile_scanner/pull/612
This PR can fix theses issues : https://github.com/juliansteenbakker/mobile_scanner/issues/674, https://github.com/juliansteenbakker/mobile_scanner/issues/752, https://github.com/juliansteenbakker/mobile_scanner/pull/749

I'm not a native expert. I tested and it works well, if you see something weird, please comment.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -22,6 +22,10 @@ import dev.steenbakker.mobile_scanner.utils.YuvToRgbConverter
 import io.flutter.view.TextureRegistry
 import java.io.ByteArrayOutputStream
 import kotlin.math.roundToInt
+import android.util.Size
+import android.hardware.display.DisplayManager
+import android.view.WindowManager
+import android.content.Context
 
 
 class MobileScanner(
@@ -166,6 +170,24 @@ class MobileScanner(
         return scaledScanWindow.contains(barcodeBoundingBox)
     }
 
+    // Return the best resolution for the actual device orientation.
+    // By default camera set its resolution to width 480 and height 640 which is too low for ML KIT.
+    // If we return an higher resolution than device can handle, camera package take the most relavant one available.
+    // Resolution set must take care of device orientation to preserve aspect ratio.
+    private fun getResolution(windowManager: WindowManager): Size {
+        val rotation = windowManager.defaultDisplay.rotation
+        val widthMaxRes = 480 * 4;
+        val heightMaxRes = 640 * 4;
+
+        val targetResolution = if (rotation == Surface.ROTATION_0 || rotation == Surface.ROTATION_180) {
+            Size(widthMaxRes, heightMaxRes) // Portrait mode
+        } else {
+            Size(heightMaxRes, widthMaxRes) // Landscape mode
+        }
+        return targetResolution
+    }
+
+
     /**
      * Start barcode scanning by initializing the camera and barcode scanner.
      */
@@ -229,7 +251,19 @@ class MobileScanner(
             // Build the analyzer to be passed on to MLKit
             val analysisBuilder = ImageAnalysis.Builder()
                 .setBackpressureStrategy(ImageAnalysis.STRATEGY_KEEP_ONLY_LATEST)
-//                analysisBuilder.setTargetResolution(Size(1440, 1920))
+            val displayManager = activity.applicationContext.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
+            val windowManager = activity.applicationContext.getSystemService(Context.WINDOW_SERVICE) as WindowManager
+            // Set initial resolution
+            analysisBuilder.setTargetResolution(getResolution(windowManager))
+            // Listen future orientation
+            displayManager.registerDisplayListener(object : DisplayManager.DisplayListener {
+                override fun onDisplayAdded(displayId: Int) {}
+                override fun onDisplayRemoved(displayId: Int) {}
+                override fun onDisplayChanged(displayId: Int) {
+                    analysisBuilder.setTargetResolution(getResolution(windowManager))
+                }
+            }, null)
+
             val analysis = analysisBuilder.build().apply { setAnalyzer(executor, captureOutput) }
 
             camera = cameraProvider!!.bindToLifecycle(
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -43,6 +43,8 @@ class MobileScanner(
     private var scanner = BarcodeScanning.getClient()
     private var lastScanned: List<String?>? = null
     private var scannerTimeout = false
+    private var displayManager: DisplayManager? = null
+    private var displayListener: DisplayManager.DisplayListener? = null
 
     /// Configurable variables
     var scanWindow: List<Float>? = null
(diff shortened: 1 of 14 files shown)


mocmeow:
I've tested your fix and it do really improve the scan distance and speed, difference is noticeably specially on not a really good cam like 5 or 8mpx cam and 16 digits barcode. No problem has been found so far, I'm waiting for a merge soon. 
Great job, thank you!

bswhite1 (approved):

EArminjon:
Hello @juliansteenbakker

What do you think of this PR ?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -166,6 +170,24 @@ class MobileScanner(
...
     }
 
+    // Return the best resolution for the actual device orientation.
+    // By default camera set its resolution to width 480 and height 640 which is too low for ML KIT.
+    // If we return an higher resolution than device can handle, camera package take the most relavant one available.
+    // Resolution set must take care of device orientation to preserve aspect ratio.
+    private fun getResolution(windowManager: WindowManager): Size {
+        val rotation = windowManager.defaultDisplay.rotation
+        val widthMaxRes = 480 * 4;
+        val heightMaxRes = 640 * 4;
```
juliansteenbakker, line 180: Thanks for the update and sorry for the delay. I have just one question: The resolution is now set to 2560 * 1920. Wouldn't it be better to make this user configurable, so in case there is an issue on some device, the resolution can be scaled down. I can imagine some older devices having trouble with this high resolution.

navaronbracke: I agree with this point. The `camera` package does something similar.

EArminjon: Okay, i can let user provide two parameters : maxResolutionHeight and maxResolutionWidth.
Where must i add that ? Inside MobileScannerController properties or inside the MobileScanner properties ?
As it is only used by Android, must i put that inside an AndroidSettings models ?

EArminjon: Also note that, Android said if the resolution is not available, Android will take the closest one.

navaronbracke: @EArminjon The configuration should be passed to the Dart-side and the native side. So you'll have to do some plumbing here and there.

If this is only for Android, then documenting that limitation is good enough.

I would go with a https://api.flutter.dev/flutter/dart-ui/Size-class.html as the type of this parameter (so that we do not have 2 values to pass around)

You can use the newer ResolutionSelector API to make it work on newer version of Android.

EArminjon: Yep i will provide a size, more efficient.
I tried to use the new api but I didn't achieve to make something working. Android native is not really easy for me...

EArminjon:
Usage : 

```dart
MobileScannerController(
    androidResolution: const Size(1920, 2560),
)
```

lib/src/mobile_scanner_controller.dart
```diff
@@ -58,9 +59,25 @@ class MobileScannerController {
   /// Automatically start the mobileScanner on initialization.
   final bool autoStart;
 
+  /// Can be used to override default Android camera resolution.
```
navaronbracke, line 62: Could you rename this property to just `cameraResolution` ? Maybe one day, we can use this to configure resolutions on other platforms as well.

```suggestion
/// The desired resolution for the camera.
///
/// When this value is provided, the camera will try to match this resolution,
/// or fallback to the closest resolution available.
/// When this is null, Android defaults to a resolution of 640x480.
///
/// Bear in mind that changing the resolution has an effect on the aspect ratio.
///
/// When the camera orientation changes,
/// the resolution will be flipped to match the new dimensions of the display.
///
/// Currently only supported on Android.
final Size? cameraResolution;
```

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -166,6 +170,24 @@ class MobileScanner(
         return scaledScanWindow.contains(barcodeBoundingBox)
     }
 
+    // Return the best resolution for the actual device orientation.
+    // By default camera set its resolution to width 480 and height 640 which is too low for ML KIT.
+    // If we return an higher resolution than device can handle, camera package take the most relevant one available.
```
navaronbracke, line 175: What do you mean with 

```
If we return an higher resolution than device can handle, camera package take the most relevant one available.
```

Is the CameraX library handling it for us already? (Not a CameraX expert myself)

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -229,7 +252,22 @@ class MobileScanner(
...
                 .setBackpressureStrategy(ImageAnalysis.STRATEGY_KEEP_ONLY_LATEST)
-//                analysisBuilder.setTargetResolution(Size(1440, 1920))
+            val displayManager = activity.applicationContext.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
+            val windowManager = activity.applicationContext.getSystemService(Context.WINDOW_SERVICE) as WindowManager
+
+            if (androidResolution != null) {
+                // Override initial resolution
+                analysisBuilder.setTargetResolution(getResolution(windowManager, androidResolution))
+                // Listen future orientation change to apply the custom resolution
+                displayManager.registerDisplayListener(object : DisplayManager.DisplayListener {
```
navaronbracke, line 262: I have a feeling that this display listener does several things wrong for now.

1. We recreate it every time
2. We never unset the listener when we no longer need it

Could you look into this? If you can't find a solution right away, could you add a TODO in the code and create a Github issue? Then we can follow up with this later.

My first guess would be to set up a display listener as a property of the MobileScanner,
keep track of the analysisBuilder (if possible) and do some delegation with that.

navaronbracke (changes requested):
Some minor remarks, but otherwise this looks good.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
EArminjon: Android camera library will adapt the requested resolution if not possible : 
image

EArminjon:
> @EArminjon did you manage to use the newer API, which replaces the deprecated setTargetResolution on newer Android versions? That is going to be a requirement for this as well.
That also mean current camera2 package to camera2 version 1.3. 

I'm not native expert, could take responsability of this :/

navaronbracke:
I can take a look for that for you.

navaronbracke:
@EArminjon I found a solution for the display listener (and the deprecated window display getter)

EArminjon:
@navaronbracke  PR updated : 
- variable renamed from androidResolution to cameraResolution.
- listener is now disposed (ty for pointing this)

navaronbracke:
@EArminjon Apparently we did things side-by-side :)

I fixed the display listener issue in a slightly different way in https://github.com/juliansteenbakker/mobile_scanner/pull/785
so that we don't need to keep track of the window manager as well.

I also fixed a small warning about a private method in the Android code and fixed a deprecated API usage.

Your original commits are included in that PR as well (except for the things we did in parallel)

EArminjon:
> @EArminjon Apparently we did things side-by-side :)
> 
> I fixed the display listener issue in a slightly different way in #785 so that we don't need to keep track of the window manager as well.
> 
> I also fixed a small warning about a private method in the Android code and fixed a deprecated API usage.
> 
> Your original commits are included in that PR as well (except for the things we did in parallel)

I'm happy to see that my fix was closest to yours, i'm pretty junior in Android Native. 
Ty for making this better ressource cleanup <3 !

I close this PR right ?

navaronbracke:
Yes that is correct. While looking at your latest commit I also found an error I made, we can indeed use `DisplayManager.DisplayListener` with an anonymous object (instead of using a separate class)

I added that to the other PR as well. Thank you!

Issue #729, opened 2023-08-11
Build failed when updating to latest version


mhd-ghaith-abtah:
when I updated to latest version "mobile_scanner 3.4.1" I got this error when trying to run my project:
 `ERROR:D8: com.android.tools.r8.kotlin.H

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:mergeExtDexDebug'.
> Could not resolve all files for configuration ':app:debugRuntimeClasspath'.
   > Failed to transform kotlin-stdlib-1.9.0.jar (org.jetbrains.kotlin:kotlin-stdlib:1.9.0) to match attributes {artifactType=android-dex, asm-transformed-variant=NONE, dexing-enable-desugaring=true, dexing-enable-jacoco-instrumentation=false, dexing-is-debuggable=true, dexing-min-sdk=21, org.gradle.category=library, org.gradle.libraryelements=jar, org.gradle.status=release, org.gradle.usage=java-runtime}.
      > Execution failed for DexingWithClasspathTransform: /Users/ghaithabtah/.gradle/caches/transforms-3/a0583c7606e960d1c6435c51b205f3f3/transformed/jetified-kotlin-stdlib-1.9.0.jar.
         > Error while dexing.

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 27s

┌─ Flutter Fix ──────────────────────────────────────────────────────────────┐
│ [!] The shrinker may have failed to optimize the Java bytecode.            │
│ To disable the shrinker, pass the `--no-shrink` flag to this command.      │
│ To learn more, see: https://developer.android.com/studio/build/shrink-code │
└────────────────────────────────────────────────────────────────────────────┘
Exception: Gradle task assembleDebug failed with exit code 1`
I have seen your build.gradle file and you are using ext.kotlin_version = '1.9.0' which is not widely supported yet.
could please downgrade to 1.8.0 at least
this is my build.gradle file:
 `buildscript {
    ext.kotlin_version = '1.8.0'
    repositories {
        google()
        mavenCentral()
    }

    dependencies {
        classpath 'com.android.tools.build:gradle:7.4.2'
        // START: FlutterFire Configuration
        classpath 'com.google.gms:google-services:4.3.15'
        // END: FlutterFire Configuration
[...]

PR #776, opened 2023-10-11
fix: Change Kotlin version from 1.9.10 to 1.7.22 to fix Android builds

vbuberen:
This change fixes the issue #729 

As I shared in the mentioned issue to use Kotlin 1.9.x project needs to be already on Android Gradle Plugin 8.x, which is not something that every project can already do now. Thus, I suggest to rollback Kotlin version for better compatibility like I had to do in Plus Plugins as well (see https://github.com/fluttercommunity/plus_plugins/issues/2251)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,7 +2,7 @@ group 'dev.steenbakker.mobile_scanner'
 version '1.0-SNAPSHOT'
 
 buildscript {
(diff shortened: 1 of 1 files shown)


navaronbracke (approved):

juliansteenbakker:
Thanks for your work @vbuberen !

PR #784, opened 2023-10-13
fix: Cleanups for plugin_platform_interface work

navaronbracke:
This PR adds some small cleanups that didn't really fall into a category.

- use `const` in the index.html, per the updated template
- fixes the "invalid use of private API" lint warnings (not linted in our setup, but still)
- fixes `use_super_parameters` violations
- adds a trailing newline to the pubspec.yaml and analysis_options.yaml
- adds a small update to a gitignore file, that is added by the `flutter create` template 
- removes a broken widget test (will be replaced in the future)
- removes `example/.metadata`, as the .metadata file lives in the root of the project
- updates some links in the example app readme
- bumps dependencies for the example app, and upgrades the example app to Dart 2.17
- updates some comments (additions from `flutter create`)


ios/Classes/MobileScannerUtilities.swift
```diff
@@ -1,10 +1,3 @@
-//
-//  Util.swift
```
navaronbracke, line 2: The file is not named `Util.swift` so I removed this comment

ios/Classes/MobileScannerPlugin.swift
```diff
@@ -63,14 +62,13 @@ public class SwiftMobileScannerPlugin: NSObject, FlutterPlugin {
...
         super.init()
     }
-
+    
     public static func register(with registrar: FlutterPluginRegistrar) {
-        let instance = SwiftMobileScannerPlugin(barcodeHandler: BarcodeHandler(registrar: registrar), registry: registrar.textures())
-        let methodChannel = FlutterMethodChannel(name:
-                                                    "dev.steenbakker.mobile_scanner/scanner/method", binaryMessenger: registrar.messenger())
-        registrar.addMethodCallDelegate(instance, channel: methodChannel)
+        let channel = FlutterMethodChannel(name: "dev.steenbakker.mobile_scanner/scanner/method", binaryMessenger: registrar.messenger())
```
navaronbracke, line 67: The template uses `channel` instead of `methodChannel` for the let here

example/test/widget_test.dart
navaronbracke, line 1: This is an old, failing test.

example/pubspec.yaml
```diff
@@ -1,22 +1,38 @@
 name: mobile_scanner_example
 description: Demonstrates how to use the mobile_scanner plugin.
+# The following line prevents the package from being accidentally published to
+# pub.dev using `flutter pub publish`. This is preferred for private packages.
 publish_to: 'none' # Remove this line if you wish to publish to pub.dev
+version: 0.0.1
```
navaronbracke, line 6: Not having the version results in a build failure for iOS apparently.

example/android/settings.gradle
```diff
@@ -1,11 +1,20 @@
-include ':app'
+pluginManagement {
```
navaronbracke, line 1: This was updated by the template. The `Gradle plugin management` block is the new part.

example/android/app/src/main/kotlin/dev/steenbakker/mobile_scanner_example/MainActivity.kt
```diff
@@ -1,6 +1,6 @@
 package dev.steenbakker.mobile_scanner_example
 
-import io.flutter.embedding.android.FlutterFragmentActivity
+import io.flutter.embedding.android.FlutterActivity
```
navaronbracke, line 3: I never had to specifically use a FragmentActivity to use mobile_scanner, so I reverted the change here to what is set up in the template.

example/ios/Runner.xcodeproj/project.pbxproj
navaronbracke, line 1: This was updated during a sample run on iOS to test things. Mostly the new RunnerTests block (actual tests will be added later!)

navaronbracke:
@juliansteenbakker for a quick review. Only cleanup that can be split off from my other work.

example/android/app/src/main/kotlin/dev/steenbakker/mobile_scanner_example/MainActivity.kt
juliansteenbakker: Strange, can't think of any reason i used this.

example/android/settings.gradle
juliansteenbakker: Looks good to me

pubspec.yaml
```diff
@@ -17,7 +17,7 @@ dependencies:
 dev_dependencies:
   flutter_test:
     sdk: flutter
-  lint: ">=1.10.0 <3.0.0"
+  flutter_lints: ^2.0.3
```
juliansteenbakker, line 20: Do you know which one is stricter? Last time i checked i thought that `lint` was stricter, but things could've been changed.

navaronbracke: I'll compare `flutter_lints` to https://github.com/passsy/dart-lint/blob/master/lib/strict.yaml and then we know which is stricter :)

I'll revert this change for now, we can switch back later.

navaronbracke:
@juliansteenbakker You are right, `lints` is stricter. It defines a whole lot of rules more then `flutter_lints` does.

navaronbracke:
Marking as a draft until the next release is published, as these changes should not be included in that one just yet.

PR #786, opened 2023-10-13
feat: Prepare changelog for version 3.5.0

navaronbracke:
@juliansteenbakker I updated the changelog with the fixes that we approved over the last few days.


navaronbracke:
Marking as a draft, since there are a few small fixes that can still be included.

PR #798, opened 2023-10-18
feat: Scanner with overlay screen added

Spyy004:
Here are the things that I have done
1) Added an overlay text which would show Please scan QR code when no QR code is scanned
2) Added a scanwindow and added white border radius to it. 
3) Added two extra buttons namely flashlight and toggleCamera view.
4) Modified the functionality of FAB. The FAB is now invisible once the camera state opens.

Please review the PR and tell me if you need any changes to it or any new additions to it.

### DIFF
--- a/example/lib/mobile_scanner_overlay.dart
+++ b/example/lib/mobile_scanner_overlay.dart
@@ -1,29 +1,28 @@
-//TODO: Create example with scanner overlay
 
 // import 'dart:ui';
-//
+
 // import 'package:flutter/material.dart';
 // import 'package:mobile_scanner/mobile_scanner.dart';
-//
+
 // void main() {
 //   runApp(const AnalyzeView());
 // }
-//
+
 // class AnalyzeView extends StatefulWidget {
 //   const AnalyzeView({Key? key}) : super(key: key);
-//
+
 //   @override
 //   _AnalyzeViewState createState() => _AnalyzeViewState();
 // }
-//
+
 // class _AnalyzeViewState extends State<AnalyzeView>
 //     with SingleTickerProviderStateMixin {
 //   List<Offset> points = [];
-//
+
 //   // CameraController cameraController = CameraController(context, width: 320, height: 150);
-//
+
 //   String? barcode;
-//
+
 //   @override
 //   Widget build(BuildContext context) {
 //     return MaterialApp(
@@ -34,7 +33,7 @@
 //               MobileScanner(
 //                   // fitScreen: false,
 //                   // controller: cameraController,
-//                   onDetect: (barcode, args) {
+//                   onDetect: (barcode) {
 //                 if (this.barcode != barcode.rawValue) {
 //                   this.barcode = barcode.rawValue;
 //                   if (barcode.corners != null) {
@@ -92,21 +91,21 @@
 //       ),
 //     );
 //   }
-//
+
 //   @override
 //   void dispose() {
 //     // cameraController.dispose();
 //     super.dispose();
 //   }
-//
+
 //   void display(Barcode barcode) {
 //     Navigator.of(context).popAndPushNamed('display', arguments: barcode);
 //   }
 // }
-//
+
 // class OpenPainter extends CustomPainter {
 //   final List<Offset> points;
-//
+
 //   OpenPainter(this.points);
 //   @override
 //   void paint(Canvas canvas, Size size) {
@@ -116,11 +115,11 @@
 //     //draw points on canvas
 //     canvas.drawPoints(PointMode.points, points, paint1);
 //   }
-//
+
 //   @override
 //   bool shouldRepaint(CustomPainter oldDelegate) => true;
 // }
-//
+
 // class OpacityCurve extends Curve {
 //   @override
 //   double transform(double t) {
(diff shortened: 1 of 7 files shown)


navaronbracke:
I will try to review it in the coming days.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,29 +1,28 @@
-//TODO: Create example with scanner overlay
```
navaronbracke, line 1: Could you remove the commented out code here? Since we now have a sample, this old code is not needed anymore.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
-// //   }
-// // }
+
+import 'package:flutter/material.dart';
+import 'package:flutter/rendering.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
+import 'package:mobile_scanner_example/barcode_scanner_window.dart';
+import 'package:mobile_scanner_example/scanner_error_widget.dart';
+
+void main() {
```
navaronbracke, line 142: Why is there a separate `void main() {}` entrypoint here? The example app already includes an entrypoint in `example/lib/main.dart`.

The widget `BarcodeScannerWithOverlay` should be the only one in this file, as the example app already includes a route to display that widget. So you can remove the `void main() {}` and `HomePage` widget in this file.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+class HomeState extends State<HomePage> {
+  @override
+  Widget build(BuildContext context) {
+    return MaterialApp(home: BarcodeScannerWithOverlay());
+  }
+}
+
+class BarcodeScannerWithOverlay extends StatefulWidget {
+  @override
+  _MyAppState createState() => _MyAppState();
```
navaronbracke, line 161: Could you rename the state class to include the name of the widget? This is a common pattern when creating StatefulWidgets.

```suggestion
class BarcodeScannerWithOverlay extends StatefulWidget {
  @override
   _BarcodeScannerWithOverlayState createState() => _BarcodeScannerWithOverlayState();
}

class _BarcodeScannerWithOverlayState extends State<BarcodeScannerWithOverlay> {
  // ...
}
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+}
+
+class _MyAppState extends State<BarcodeScannerWithOverlay> {
+  String? qr;
+  String overlayText = "Please scan QR Code";
+  bool camState = false;
+  late BarcodeCapture currentBarcodeCapture ;
```
navaronbracke, line 168: Prefer using a nullable value over a late one. If a late variable is not initialized, you get a `LateInitializationError`, which is harder to detect.

There is also an extra space here.

```suggestion
 BarcodeCapture? currentBarcodeCapture;
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+  final MobileScannerController controller = MobileScannerController(
+    facing: CameraFacing.back,
+    autoStart: false
+  );
+  @override
```
navaronbracke, line 173: This `initState()` method does nothing? It can be removed.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+  }
+}
+
+class BarcodeScannerWithOverlay extends StatefulWidget {
+  @override
+  _MyAppState createState() => _MyAppState();
+}
+
+class _MyAppState extends State<BarcodeScannerWithOverlay> {
+  String? qr;
```
navaronbracke, line 165: You can get the QR code from the BarcodeCapture value, so the `qr` property is not needed?

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+  String overlayText = "Please scan QR Code";
+  bool camState = false;
```
navaronbracke, line 167:
```suggestion
  bool cameraStarted = false;
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+  late BarcodeCapture currentBarcodeCapture ;
+  final MobileScannerController controller = MobileScannerController(
+    facing: CameraFacing.back,
```
navaronbracke, line 170: Since you mention having to scan a QR code, specifying the formats is a good thing to do.

```suggestion
    facing: CameraFacing.back,
    formats: const [BarcodeFormat.qrCode],
    autoStart: false,
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+    autoStart: false
+  );
+  @override
+  initState() {
+    super.initState();
+  }
+
+  @override
+  dispose() {
+    super.dispose();
```
navaronbracke, line 180: For lifecycle methods, if it is a positive method (create, resume etc), the `super` goes first.
If it is a negative method (dispose, stop etc), the `super` goes last.

```suggestion
  @override
  dispose() {
    controller.dispose();
    super.dispose();
  }
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+    controller.dispose();
+  }
+
+  configureCameraSettings() {
```
navaronbracke, line 184:
```suggestion
  startCamera() {
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+    setState(() {
+      camState = !camState;
+      controller.start();
+    });
+  }
+
+  onBarcodeDetect(BarcodeCapture barcodeCapture) {
```
navaronbracke, line 191: Could you fix the indentation for this method, by using `dart format` ?

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+  setState(() {
+    currentBarcodeCapture = barcodeCapture;
+    overlayText = barcodeCapture.barcodes.last.displayValue!;
+  });
+  }
+
```
navaronbracke, line 197: There are 2 empty lines here that can be removed.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final scanWindow = Rect.fromCenter(
+      center: MediaQuery.of(context).size.center(Offset.zero),
+      width: 200,
+      height: 200,
+    );
+    final height = MediaQuery.of(context).size.height;
```
navaronbracke, line 207: Could you do this instead?

`final Size size = MediaQuery.sizeOf(context);`

If `MediaQuery.sizeOf` is not found, you should probably increase the minimum Flutter version of the example in `example/pubspec.yaml`

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+          children: <Widget>[
+            Expanded(
+                child: camState
+                    ? Stack(
+                      fit: StackFit.expand,
+                      children: [
+                        Center(
+                        child: MobileScanner(
+                          fit: BoxFit.contain,
+                          onDetect: (BarcodeCapture barcodeCapture) => {
```
navaronbracke, line 226: Could you pass the function directly?

```suggestion
onDetect: onBarcodeDetect,
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+                         errorBuilder: (context, error, child) {
+                    return ScannerErrorWidget(error: error);
+                },
+                        ),
+                
+                ),
+                        CustomPaint(
+                          painter:  ScannerOverlay(scanWindow),
+                        ),
+                        Positioned(
```
navaronbracke, line 251: Could you use an `Align` with some padding for the child instead of a Positioned? Then we do not have "magic numbers" for the left and bottom. You can use `Alignment.bottomCenter` for positioning the child.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+                          bottom: 0.07 * height,
+                          left: 0.35 * width,
+                          child: Row(
+                            mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+                            children: [
+                              IconButton(onPressed: ()=>{
```
navaronbracke, line 257:
```suggestion
onPressed: () => controller.toggleTorch(),
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+                                controller.toggleTorch()
+                              }, icon:  Icon(Icons.flashlight_on, color: controller.torchEnabled ? Colors.yellow : Colors.black,),),
+                              IconButton(onPressed: ()=>{
```
navaronbracke, line 260:
```suggestion
onPressed: () => controller.switchCamera(),
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+                      children: [
+                        Center(
+                        child: MobileScanner(
+                          fit: BoxFit.contain,
+                          onDetect: (BarcodeCapture barcodeCapture) => {
+                            onBarcodeDetect(barcodeCapture)
+
+                        },
+                        overlay: 
+                      Positioned(
```
navaronbracke, line 231: Could you use an `Align` with some padding for the child instead of a Positioned? Then we do not have "magic numbers" for the bottom. You can use Alignment.bottomCenter for positioning the child.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -133,96 +132,199 @@
...
+                         bottom: height * 0.2,
+                        child:Opacity(opacity: 0.7, child:
```
navaronbracke, line 233: There are some formatting issues here. Could you apply `dart format` to this file. You might need to add comma's where they are missing.

navaronbracke (changes requested):
I have some small remarks, but you did good already!

example/lib/mobile_scanner_overlay.dart
Spyy004: Can you please explain what would be the issue if we use height directly ?

example/lib/mobile_scanner_overlay.dart
Spyy004: Any specific number you have in mind for the padding ?

Spyy004:
Thanks for the detailed review and feedback. I have updated the PR. Please have a look

example/lib/mobile_scanner_overlay.dart
navaronbracke: Doing `MediaQuery.of(context).size` makes the widget rebuild when any property of the MediaQuery changes.

So it rebuilds when the size changes, but _also_ when for example the `boldTextOverride` changes.

If we do `MediaQuery.sizeOf(context);` then we rebuild only when the size changes.
This method uses the size as an "aspect".

See https://api.flutter.dev/flutter/widgets/InheritedModel-class.html

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,228 +1,192 @@
...
+  }
+
+  void startCamera() {
+    setState(() {
+      camStarted = !camStarted;
+      controller.start();
+    });
+  }
+
+  void onBarcodeDetect(BarcodeCapture barcodeCapture) {
```
navaronbracke, line 34: The display value can be null. We can use the raw value as fallback.
In the case where that is also null, just use a message like this?

```suggestion
  void onBarcodeDetect(BarcodeCapture barcodeCapture) {
    final barcode = barcodeCapture.barcodes.last;
  
    setState(() {
      overlayText = barcode.displayValue ?? barcode.rawValue ?? 'Barcode has no displayable value';
    });
  }
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,228 +1,192 @@
...
+class BarcodeScannerWithOverlay extends StatefulWidget {
+  @override
+  _BarcodeScannerWithOverlayState createState() =>
+      _BarcodeScannerWithOverlayState();
+}
+
+class _BarcodeScannerWithOverlayState extends State<BarcodeScannerWithOverlay> {
+  String overlayText = "Please scan QR Code";
+  bool camStarted = false;
+  BarcodeCapture? currentBarcodeCapture;
```
navaronbracke, line 14: Apologies, it seems that `currentBarcodeCapture` is unused now, since we use the detected barcode
to set the overlay text directly.

Can you remove this unused value?

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,228 +1,192 @@
...
+                                  ),
+                                ),
+                              ],
+                            ),
+                          ),
+                        ),
+                      ],
+                    )
+                  : const Center(
+                      child: Text("Tap on Camera to activate QR Scanner")),
```
navaronbracke, line 127: There is a missing comma here, after the Text widget.

```suggestion
                      child: Text("Tap on Camera to activate QR Scanner"),),
```

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,228 +1,192 @@
...
+    final borderRect = RRect.fromRectAndCorners(
+      scanWindow,
+      topLeft: const Radius.circular(12.0),
+      topRight: const Radius.circular(12.0),
+      bottomLeft: const Radius.circular(12.0),
+      bottomRight: const Radius.circular(12.0),
+    );
+
+    // Draw the white border
+    canvas.drawPath(backgroundWithCutout, backgroundPaint);
```
navaronbracke, line 184: It seems that only the border rectangle has rounded corners. Can you have the cutout also provide rounded corners? Otherwise there is a small issue with the corners not having the same border radius.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,228 +1,192 @@
...
+                          painter: ScannerOverlay(scanWindow),
+                        ),
+                        Padding(
+                          padding: const EdgeInsets.all(16.0),
+                          child: Align(
+                            alignment: Alignment.bottomCenter,
+                            child: Row(
+                              mainAxisAlignment: MainAxisAlignment.spaceEvenly,
+                              children: [
+                                IconButton(
```
navaronbracke, line 104: The `torchEnabled` flag is only to turn on the torch when the scanner starts.

Can you instead use a ValueListenableBuilder and pass the torchState of the controller?
Then it updates when the torch is on or off.

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,228 +1,192 @@
...
-// //             });
-// //           }),
-// //     );
-// //   }
-// // }
+import 'package:flutter/material.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
+import 'package:mobile_scanner_example/scanner_error_widget.dart';
+
+class BarcodeScannerWithOverlay extends StatefulWidget {
```
navaronbracke, line 5: If I start the example and deny the camera permission, I get an unhandled exception and the screen stays white.

Either add a TODO comment to fix this later, or handle the error codes that come from the `start` method.

navaronbracke (changes requested):
I ran the sample on my phone and it looks amazing already.

Just some small remarks for the updated sample.

example/lib/mobile_scanner_overlay.dart
Spyy004: So I was thinking of adding a try catch block in startCamera() and handling the exception by showing a snackbar. I guess that would work

navaronbracke: Yes, that could work as well. If I recall correctly the official `camera` example does that too.

Spyy004:
I have updated the PR and made the changes as you requested. Please have a look on the handling camera exception part. I think what I wrote should work. Once again, thanks for the feedback

example/lib/mobile_scanner_overlay.dart
```diff
@@ -1,228 +1,223 @@
...
+  @override
+  _BarcodeScannerWithOverlayState createState() =>
+      _BarcodeScannerWithOverlayState();
+}
+
+class _BarcodeScannerWithOverlayState extends State<BarcodeScannerWithOverlay> {
+  String overlayText = "Please scan QR Code";
+  bool camStarted = false;
+
+  final ValueNotifier<bool> torchStateNotifier = ValueNotifier<bool>(false);
```
navaronbracke, line 15: There is no need for a separate ValueNotifier, as the controller already has one for the torch state (and updates it when the value changes).

Could you pass `controller.torchState` to the widget below?
Its value is an enum rather than a boolean, but it contains values of off and on states.

navaronbracke (changes requested):
One last remark about the value notifier.

navaronbracke (approved):
Since the ValueNotifier issue is not something that breaks the sample, I'll adjust that later.

Then this sample can be included in the next release.

Spyy004:
Oh you merged it. I was going to fix this thing now. Sorry and thanks!
Great experience learning new things!

navaronbracke:
No problem! I am currently looking into getting the next release published. That's why I had a look at it myself :)

Issue #799, opened 2023-10-18
CRASH ANDROID API >=30 


EArminjon:
Crash present on master branch 

MobileScanner.kt

```kotlin
val rotation = if (Build.VERSION.SDK_INT >= 30) {
            activity.applicationContext.display!!.rotation
        } else {
```
```bash
java.lang.UnsupportedOperationException: Tried to obtain display from a Context not associated with one. Only visual Contexts (such as Activity or one created with Context#createWindowContext) or ones created with Context#createDisplayContext are associated with displays. Other types of Contexts are typically related to background entities and may return an arbitrary display.
E/AndroidRuntime(22817): 	at android.app.ContextImpl.getDisplay(ContextImpl.java:3025)
E/AndroidRuntime(22817): 	at android.content.ContextWrapper.getDisplay(ContextWrapper.java:1192)
E/AndroidRuntime(22817): 	at dev.steenbakker.mobile_scanner.MobileScanner.getResolution(MobileScanner.kt:185)
E/AndroidRuntime(22817): 	at dev.steenbakker.mobile_scanner.MobileScanner.start$lambda$11(MobileScanner.kt:272)
E/AndroidRuntime(22817): 	at dev.steenbakker.mobile_scanner.MobileScanner.$r8$lambda$QBaqIq_loyYbiAjfq_M2yLd3JMk(Unknown Source:0)
E/AndroidRuntime(22817): 	at dev.steenbakker.mobile_scanner.MobileScanner$$ExternalSyntheticLambda2.run(Unknown Source:18)
E/AndroidRuntime(22817): 	at android.os.Handler.handleCallback(Handler.java:958)
E/AndroidRuntime(22817): 	at android.os.Handler.dispatchMessage(Handler.java:99)
E/AndroidRuntime(22817): 	at android.os.Looper.loopOnce(Looper.java:205)
E/AndroidRuntime(22817): 	at android.os.Looper.loop(Looper.java:294)
E/AndroidRuntime(22817): 	at android.app.ActivityThread.main(ActivityThread.java:8177)
E/AndroidRuntime(22817): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(22817): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:552)
E/AndroidRuntime(22817): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:971)
[...]

PR #785, opened 2023-10-13
fix: Increase camera quality (cherry-pick)

navaronbracke:
This PR is a cherry pick of https://github.com/juliansteenbakker/mobile_scanner/pull/760

with some additional changes:
- `androidResolution` has been renamed to `cameraResolution`, for future-proofing
- documentation adjustment for `cameraResolution`
- a warning about a method that could be private was fixed
- adjusts the use of the deprecated display property for Android 30+
- fixes a leak of the display change listener by only making one on demand and disposing it in `stop()`

Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/674
Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/752
Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/749

Credit goes to @EArminjon for most of the work, I just did the extra fixes.

PR #800, opened 2023-10-18
fix: Null exception when getting activity display on Android API >= 30

EArminjon:
Fix https://github.com/juliansteenbakker/mobile_scanner/issues/799 introduced by https://github.com/juliansteenbakker/mobile_scanner/pull/785

When using cameraResolution, actual manner of getting display object lead to app crash.

Tested on Android >= 30 and <= 30.


navaronbracke:
Ah, I didn't read the documentation carefully enough.

```
Get the display this context is associated with. Applications should use this method with Activity or a context associated with a Display via createDisplayContext(Display) to get a display object associated with a Context, or android.hardware.display.DisplayManager.getDisplay to get a display object by id.
```

Good catch!

navaronbracke (approved):

Issue #801, opened 2023-10-19
iOS - Camera got better scan detection in landscape than in portrait


EArminjon:
My users complain about iOS scan detection which is not very performant and this requires my users to be close to the barcodes. They also complain about the fact that in landscape they could have better scan result.

After analysing returned image by Android and iOS, I see that I got a rotated image for iOS. I think this is the source of the issue.
Maybe we give to ML Kit on iOS a rotated image which lead to bad scan detection.

Screenshots which display the image inside the `onDetect` (quality is not same, it's normal) :

```dart
onDetect: (BarcodeCapture barcode) {
  showModalBottomSheet(
    context: context,
    builder: (context) => AlertDialog(
      title: Text("Image result"),
      content: Image.memory(barcode.image!),
    ),
);
```
Android:
![image](https://github.com/juliansteenbakker/mobile_scanner/assets/37028599/33d8cc1e-5ab2-4826-8bb4-b4252c26868d)

iOS : 
![image](https://github.com/juliansteenbakker/mobile_scanner/assets/37028599/0a7bed35-6c90-42df-bf37-133a59286995)

Flutter 3.13.7
Master Branch (commit: 76bc6cc88d021444efc6cbbc0b6ed7733fd4919f)

EArminjon:
> @EArminjon Perhaps the issue is with the use of this function?
> 
> https://github.com/juliansteenbakker/mobile_scanner/blob/master/ios/Classes/MobileScanner.swift#L369-L388

I tried to disabled it but I didn't notice any change.

navaronbracke:
I will try to have a look myself. If you find anything in the meantime, I'll be happy to review the fix.

EArminjon:
Maybe coming from MobileScanerUtilities.swift. An extension create a getter to call .image
This getter put image in .left position.

Maybe it could be better to try read the original orientation and if not possible fallback to the .up position.


Current code : 
```swift
var image: UIImage {
        let ciImage = CIImage(cvPixelBuffer: self)
        let cgImage = CIContext().createCGImage(ciImage, from: ciImage.extent)
       return UIImage(cgImage: cgImage!, scale: 1.0, orientation: UIImage.Orientation.left)
    }
```

navaronbracke:
That could be it. Could you test this locally and open a PR if it fixes it?

PR #802, opened 2023-10-19
fix: iOS Orientation change from .left to .up

EArminjon:
This PR fix the weird orientation seen in https://github.com/juliansteenbakker/mobile_scanner/issues/801.
I didn't find a way to determine the original orientation from the buffer, but it's definitly weird to fallback to the `left`.

### DIFF
--- a/ios/Classes/MobileScannerUtilities.swift
+++ b/ios/Classes/MobileScannerUtilities.swift
@@ -21,7 +21,7 @@ extension CVBuffer {
     var image: UIImage {
         let ciImage = CIImage(cvPixelBuffer: self)
         let cgImage = CIContext().createCGImage(ciImage, from: ciImage.extent)
-        return UIImage(cgImage: cgImage!, scale: 1.0, orientation: UIImage.Orientation.left)
+        return UIImage(cgImage: cgImage!)
     }
     
     var image1: UIImage {


EArminjon:
Issue introduced by : https://github.com/juliansteenbakker/mobile_scanner/commit/5af3a81e6ca285ba3b88b4581b861f55f2e792bc#diff-f2d77367e450abec39e299c6e9b1a59a026b57f02c278c9b310b1d4514bef646

ios/Classes/MobileScannerUtilities.swift
```diff
@@ -21,7 +21,7 @@ extension CVBuffer {
     var image: UIImage {
         let ciImage = CIImage(cvPixelBuffer: self)
         let cgImage = CIContext().createCGImage(ciImage, from: ciImage.extent)
-        return UIImage(cgImage: cgImage!, scale: 1.0, orientation: UIImage.Orientation.left)
+        return UIImage(cgImage: cgImage!, scale: 1.0, orientation: UIImage.Orientation.up)
```
navaronbracke, line 24: @EArminjon in the original code, `orientation` was not given (see the commit you linked in the description). Is it broken if we do `UIImage(cgImage: cgImage!, scale: 1.0)` ?

EArminjon: I also need to remove the scale property. If not I got a compile error. By removing both, no issue, image in correct orientation.

EArminjon: Done :)

Work well, don't hesitate to test if you have a recent or very old iOS device near you. (iOS simulator don't have camera feature)

PR #810, opened 2023-10-22
feat: Update Android setup

navaronbracke:
This PR updates the Android setup with some updates from the `flutter create` template.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -14,7 +14,7 @@ buildscript {
     }
 }
 
-rootProject.allprojects {
+allprojects {
     repositories {
         google()
         mavenCentral()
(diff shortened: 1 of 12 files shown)


example/android/settings.gradle
```diff
@@ -1,11 +1,16 @@
-include ':app'
+pluginManagement {
```
navaronbracke, line 1: The pluginManagement block is new. In the template it also added a redundant apply plugins block. However, Bartek Pacia has already fixed that upstream in the template. I removed it here as well.

android/build.gradle
```diff
@@ -38,13 +42,24 @@ android {
...
         main.java.srcDirs += 'src/main/kotlin'
+        test.java.srcDirs += 'src/test/kotlin'
     }
 
     defaultConfig {
         minSdkVersion 21
     }
 
-    namespace 'dev.steenbakker.mobile_scanner'
+    testOptions {
```
navaronbracke, line 52: New test options, for when we have native tests one day :)

PR #815, opened 2023-10-23
fix: fix typo in `codebar` constant

navaronbracke:
This PR fixes a typo in the `BarcodeFormat.codebar` constant.

The actual name of this format is `Codabar`, with an "a" rather than an "e".
See https://en.wikipedia.org/wiki/Codabar

For now the old constant just points to the new one, so both will work.
I added a comment to point users to the new constant.

We can deprecate the old constant in a follow-up release, and remove it in the release after that one.


lib/src/web/zxing.dart
```diff
@@ -107,7 +107,7 @@ extension ZXingBarcodeFormat on BarcodeFormat {
     switch (this) {
       case BarcodeFormat.aztec:
         return 0;
-      case BarcodeFormat.codebar:
+      case BarcodeFormat.codabar:
```
navaronbracke, line 110: Adding `case BarcodeFormat.codebar:` is not needed, since then we get a warning bout duplicate cases.

PR #816, opened 2023-10-23
fix: use rawValue on Android as well for formats

navaronbracke:
This PR adjusts the Android implementation to accept the raw value of the format, to bring it in line with other platforms.

It also removes an unused argument, formats a Kotlin file, and moves the `cameraResolution` argument out of a platform check. If the `cameraResolution` is not supported by a platform, it is ignored in the implementation anyway.

Lastly, all the enums now use their `rawValue` instead of their `index` as serialized value to the platform side.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -120,7 +120,7 @@ class MobileScannerHandler(
             "stop" -> stop(result)
             "analyzeImage" -> analyzeImage(call, result)
             "setScale" -> setScale(call, result)
-            "resetScale" -> resetScale(call, result)
+            "resetScale" -> resetScale(result)
             "updateScanWindow" -> updateScanWindow(call)
             else -> result.notImplemented()
         }
@@ -250,7 +250,7 @@ class MobileScannerHandler(
         }
     }
 
-    private fun resetScale(call: MethodCall, result: MethodChannel.Result) {
+    private fun resetScale(result: MethodChannel.Result) {
         try {
             mobileScanner!!.resetScale()
             result.success(null)
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -144,8 +144,8 @@ class MobileScannerHandler(
         var barcodeScannerOptions: BarcodeScannerOptions? = null
         if (formats != null) {
             val formatsList: MutableList<Int> = mutableListOf()
-            for (index in formats) {
-                formatsList.add(BarcodeFormats.values()[index].intValue)
+            for (formatValue in formats) {
+                formatsList.add(BarcodeFormats.fromRawValue(formatValue).intValue)
             }
             barcodeScannerOptions = if (formatsList.size == 1) {
                 BarcodeScannerOptions.Builder().setBarcodeFormats(formatsList.first())
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/BarcodeFormats.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/BarcodeFormats.kt
@@ -2,17 +2,41 @@ package dev.steenbakker.mobile_scanner.objects
 
 enum class BarcodeFormats(val intValue: Int) {
     UNKNOWN(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_UNKNOWN),
-    ALL_FORMATS(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_ALL_FORMATS), CODE_128(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_CODE_128), CODE_39(
-        com.google.mlkit.vision.barcode.common.Barcode.FORMAT_CODE_39
-    ),
-    CODE_93(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_CODE_93), CODABAR(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_CODABAR), DATA_MATRIX(
-        com.google.mlkit.vision.barcode.common.Barcode.FORMAT_DATA_MATRIX
-    ),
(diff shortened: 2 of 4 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -250,7 +250,7 @@ class MobileScannerHandler(
         }
     }
 
-    private fun resetScale(call: MethodCall, result: MethodChannel.Result) {
+    private fun resetScale(result: MethodChannel.Result) {
```
navaronbracke, line 253: `call` was unused

android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/BarcodeFormats.kt
```diff
@@ -2,17 +2,41 @@ package dev.steenbakker.mobile_scanner.objects
...
-        com.google.mlkit.vision.barcode.common.Barcode.FORMAT_DATA_MATRIX
-    ),
-    EAN_13(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_EAN_13), EAN_8(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_EAN_8), ITF(
-        com.google.mlkit.vision.barcode.common.Barcode.FORMAT_ITF
-    ),
-    QR_CODE(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_QR_CODE), UPC_A(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_UPC_A), UPC_E(
-        com.google.mlkit.vision.barcode.common.Barcode.FORMAT_UPC_E
-    ),
-    PDF417(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_PDF417), AZTEC(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_AZTEC);
+    ALL_FORMATS(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_ALL_FORMATS),
```
navaronbracke, line 5: Formatted the enum values to each be on their own line.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/BarcodeFormats.kt
```diff
@@ -2,17 +2,41 @@ package dev.steenbakker.mobile_scanner.objects
...
+    EAN_8(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_EAN_8),
+    ITF(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_ITF),
+    QR_CODE(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_QR_CODE),
+    UPC_A(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_UPC_A),
+    UPC_E(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_UPC_E),
+    PDF417(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_PDF417),
+    AZTEC(com.google.mlkit.vision.barcode.common.Barcode.FORMAT_AZTEC);
+
+    companion object {
+        fun fromRawValue(rawValue: Int): BarcodeFormats {
```
navaronbracke, line 21: This is a conversion similar to what happens on the Dart side.

Having a secondary constructor didn't work, since the primary one already uses the same arguments list.

lib/src/mobile_scanner_controller.dart
```diff
@@ -147,19 +148,18 @@ class MobileScannerController {
...
-        arguments['formats'] = formats!.map((e) => e.index).toList();
-        if (cameraResolution != null) {
-          arguments['cameraResolution'] = <int>[
-            cameraResolution!.width.toInt(),
-            cameraResolution!.height.toInt(),
-          ];
-        }
       }
     }
-    arguments['returnImage'] = returnImage;
```
navaronbracke, line 162: The line `arguments['returnImage'] = returnImage;` was moved next to the other ones, for readability. I might turn the argument map into a map literal later.

lib/mobile_scanner_web_plugin.dart
```diff
@@ -139,7 +139,7 @@ class MobileScannerWebPlugin {
               'rawBytes': code.rawBytes,
               'format': code.format.rawValue,
               'displayValue': code.displayValue,
-              'type': code.type.index,
+              'type': code.type.rawValue,
```
navaronbracke, line 142: The raw value is the same as the index in the `BarcodeType` enum.

lib/src/mobile_scanner_controller.dart
```diff
@@ -132,10 +132,11 @@ class MobileScannerController {
     final Map<String, dynamic> arguments = {};
 
     cameraFacingState.value = cameraFacingOverride ?? facing;
-    arguments['facing'] = cameraFacingState.value.index;
+    arguments['facing'] = cameraFacingState.value.rawValue;
```
navaronbracke, line 135: The raw value is the same as the index in the `CameraFacing` enum.

lib/src/mobile_scanner_controller.dart
```diff
@@ -132,10 +132,11 @@ class MobileScannerController {
...
     arguments['torch'] = torchEnabled;
-    arguments['speed'] = detectionSpeed.index;
+    arguments['speed'] = detectionSpeed.rawValue;
```
navaronbracke, line 137: The raw value is the same as the index in the `DetectionSpeed` enum.

lib/src/mobile_scanner_controller.dart
```diff
@@ -147,19 +148,18 @@ class MobileScannerController {
     } */
 
     if (formats != null) {
-      if (kIsWeb || Platform.isIOS || Platform.isMacOS) {
+      if (kIsWeb || Platform.isIOS || Platform.isMacOS || Platform.isAndroid) {
```
navaronbracke, line 151: All supported platforms now use the `rawValue` for the formats.

lib/src/mobile_scanner_controller.dart
```diff
@@ -147,19 +148,18 @@ class MobileScannerController {
...
-          arguments['cameraResolution'] = <int>[
-            cameraResolution!.width.toInt(),
-            cameraResolution!.height.toInt(),
-          ];
-        }
       }
     }
-    arguments['returnImage'] = returnImage;
+
+    if (cameraResolution != null) {
```
navaronbracke, line 156: This block was previously only set for Android. If other platforms don't need the argument, their implementations can just ignore it.

However, if a platform wants to support it, then the argument is already set up for consumption.

PR #818, opened 2023-10-23
feat: Move classes into their own files

navaronbracke:
The classes in `src/objects/barcode.dart` have been moved to their own files.
The API for these classes stays the same unless otherwise noted below.
 
- The `type` attribute of the `Address`, `Email`, `Phone` is now non-null.
   - The enums now specify `unknown` as their value, instead of null.
- The `phoneNumber` of an SMS is now non-null.
   - This was an assumption that I made. The empty string is used in place of null.
- The `latitude` and `longitude` of a `GeoPoint` are now non-null.
   - This was an assumption that I made. A valid point should have both latitude and longitude anyway
   - If either is null, the `GeoPoint.fromNative` method returns `0,0` as fallback.
- The `url` of a `UrlBookmark` is now non-null.
   - This was an assumption that I made. The empty string is used in place of null.
- The `phones` and `urls` of `ContactInfo` are now non-null.
   - empty lists are used if the actual value was null.
- The `BarcodeCapture` class now exposes a `size`.
   - eventually we can deprecate the `width / height`, but keep the `size`
   - Size.zero is used as default if width or height is null
- The `width` and `height` of `BarcodeCapture` are now non-null.
  - These attributes are now getters, which expose the width/height of the `size`
- Cleaned up a Platform check in `barcode_scanner_window.dart`
  - Previously it did `Platform.isIOS` twice, now it only does it once
  - I also added a check for `kIsWeb` because `Platform.isIOS` throws on web
- The `corners` of `BarcodeCapture` are now non-null. If there are no corners, the list is empty.
- The `fromNative()` methods on the data classes now accept `Map<Object?, Object?>` instead of `Map<dynamic, dynamic>`
   - This change is by definition a **breaking change**, but to an internal-only API (consumers _should_ be using the regular constructor, if they ever want to create these classes)

### DIFF
--- a/example/lib/barcode_scanner_window.dart
+++ b/example/lib/barcode_scanner_window.dart
@@ -151,7 +151,7 @@ class BarcodeOverlay extends CustomPainter {
 
   @override
   void paint(Canvas canvas, Size size) {
-    if (barcode.corners == null) {
+    if (barcode.corners.isEmpty) {
       return;
     }
 
@@ -183,7 +183,7 @@ class BarcodeOverlay extends CustomPainter {
     }
 
     final List<Offset> adjustedOffset = [];
-    for (final offset in barcode.corners!) {
+    for (final offset in barcode.corners) {
       adjustedOffset.add(
         Offset(
           offset.dx / ratioWidth + horizontalPadding,
--- a/example/lib/barcode_scanner_window.dart
+++ b/example/lib/barcode_scanner_window.dart
@@ -1,5 +1,6 @@
 import 'dart:io';
 
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
 import 'package:mobile_scanner/mobile_scanner.dart';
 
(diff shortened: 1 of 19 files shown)


example/lib/barcode_scanner_window.dart
```diff
@@ -167,15 +171,19 @@ class BarcodeOverlay extends CustomPainter {
...
-    final ratioWidth =
-        (Platform.isIOS ? capture.width! : arguments.size.width) /
-            adjustedSize.destination.width;
-    final ratioHeight =
-        (Platform.isIOS ? capture.height! : arguments.size.height) /
-            adjustedSize.destination.height;
+    final double ratioWidth;
+    final double ratioHeight;
+
+    if (!kIsWeb && Platform.isIOS) {
```
navaronbracke, line 177: Instead of having 2x a Platform check, I rewrote this to only one. I also added a check for web, since Platform.isIOS throw otherwise.

PR #838, opened 2023-10-29
fix: Handle no cameras on Android

### LINKED ISSUE #719: NPE - Attempt to invoke virtual method 'android.hardware.camera2.CaptureRequest android.hardware.camera2.impl.CameraDeviceImpl$CaptureCallbackHolder.getRequest(int)' on a null object reference
From the @openfoodfacts Sentry system: 
NPE - Attempt to invoke virtual method 'android.hardware.camera2.CaptureRequest android.hardware.camera2.impl.CameraDeviceImpl$CaptureCallbackHolder.getRequest(int)' on a null object reference

Sentry Issue: [SMOOTHIE-2EC](https://openfoodfacts.sentry.io/issues/4275004795/?referrer=github_integration)

```
NullPointerException: Attempt to invoke virtual method 'android.hardware.camera2.CaptureRequest android.hardware.camera2.impl.CameraDeviceImpl$CaptureCallbackHolder.getRequest(int)' on a null object reference
    at android.hardware.camera2.impl.CameraDeviceImpl$CameraDeviceCallbacks.onResultReceived(CameraDeviceImpl.java:2070)
    at android.hardware.camera2.legacy.CameraDeviceUserShim$CameraCallbackThread$CallbackHandler.handleMessage(CameraDeviceUserShim.java:325)
    at android.os.Handler.dispatchMessage(Handler.java:106)
    at android.os.Looper.loop(Looper.java:193)
    at android.os.HandlerThread.run(HandlerThread.java:65)
```

### LINKED ISSUE #762: Crash when no cameras are available on Android
Using the mobile_scanner package makes no camera Android devices not compatible. 

Here is the solution: https://developer.android.com/guide/topics/manifest/uses-permission-element

navaronbracke:
This PR fixes some camera issues on Android by updating how we work with the camera:

- Only do safe access of the camera/cameraProvider using `?.`, instead of `!!.` which forced it.
- Added the feature flag to the manifest to denote that the camera is optional (since we now handle "no camera" properly)
- Updated the error handling for the camera provider to emit errors through a callback
   - This was not working, because the errors are thrown in the camera's executor.
- Removed the unused `TorchError` class
- Updated the toggle torch code to only do this when the device has a flash unit
- Fixes two typos that checked for the wrong error type

Fixes #762 
Fixes #719

### DIFF
--- a/android/src/main/AndroidManifest.xml
+++ b/android/src/main/AndroidManifest.xml
@@ -2,4 +2,5 @@
   package="dev.steenbakker.mobile_scanner">
 
     <uses-permission android:name="android.permission.CAMERA" />
+    <uses-feature android:name="android.hardware.camera" android:required="false" />
 </manifest>
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -235,10 +235,12 @@ class MobileScanner(
 
         cameraProviderFuture.addListener({
             cameraProvider = cameraProviderFuture.get()
+
             if (cameraProvider == null) {
                 throw CameraError()
             }
-            cameraProvider!!.unbindAll()
+
+            cameraProvider?.unbindAll()
             textureEntry = textureRegistry.createSurfaceTexture()
 
             // Preview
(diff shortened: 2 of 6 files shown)


navaronbracke:
@opame-app or @teolemon Do you guys have a physical device without cameras to test on? I do not, and setting the emulator's cameras to `None` in the advanced settings results in a non-fatal error for the lens facing validation in CameraX, because `The expected camera is not available on the device`. I think it has to do with the emulator itself?

The example app runs fine, and reports the no camera error case in the UI.

Issue #750, opened 2023-09-12
Setting DetectionSpeed to noDuplicates prevent the scanner from working after pop and push the view again 


MiBSchmidt:
Hello
to reproduce the error you have to do the following steps:
1. Open the file `barcode_scanner_controller.dart`  from the example app
2. Set the detectionSpeed to noDuplicates
```

   MobileScannerController controller = MobileScannerController(
    torchEnabled: true,
    // formats: [BarcodeFormat.qrCode]
    // facing: CameraFacing.front,
    detectionSpeed: DetectionSpeed.noDuplicates
    // detectionTimeoutMs: 1000,
    // returnImage: false,
  );
````
3. Start the example app
4. Tap on the button "MobileScanner with Controller"
5. Scann a Barcode 
6. Return to main menu
7. Select the button "MobileScanner with Controller"

Now the barcode will not be detected anymore.  If the detectionSpeed is set to normal, everythink works fine.

I guess this issue is related to #619 and #589

I am using a Google Pixel 7 Pro with Android 13 

Greetings 

Michael

navaronbracke:
This is fixed in version 3.5.2

Juanchote:
I am having the same issue with version 3.5.5
Even if I dispose the controller, it doesn't scan the barcodes anymore.

I have Samsung S10+ with android 12.

PR #839, opened 2023-10-30
fix: with the "noDuplicates", the same code is no longer detected after stopping and restarting the scanner

pgeof:
This PR proposes a fix for the issue reported here #750, which I also encountered.

To summarize, we reset the variable that holds the list of recently scanned codes every time the mobile scanner controller is restarted. I also sorted the list to avoid false comparisons of lists with the same codes but in a different order.

Additionally, on iOS, I had a code that was being spammed even though the "noDuplicates" speed was enabled. Upon reviewing the code, I noticed that this occurred when an empty code list was returned by the scanner, which happens as soon as there are no more codes in the camera's field of view. So I added a check to ignore empty lists. In fact, this check was already present in the Android code.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -70,7 +70,7 @@ class MobileScanner(
         scanner.process(inputImage)
             .addOnSuccessListener { barcodes ->
                 if (detectionSpeed == DetectionSpeed.NO_DUPLICATES) {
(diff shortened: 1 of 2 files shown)


navaronbracke (approved):
Do we need the same fix for MacOS ? Since that uses Apple's Vision API, the behavior might be different.

navaronbracke:
Just tested this with the repro steps from the issue and MacOS works fine with `noDuplicates`.

pgeof:
> Just tested this with the repro steps from the issue and MacOS works fine with `noDuplicates`.

I must admit I did not think about macOS and the Vision API, but thank you for validating that!

PR #870, opened 2023-11-21
fix: iOS Timeout is broken because of Double miss conversion

EArminjon:
A Double conversion is broken leading to an Int round result. 
(You can quickly check that here : https://online.swiftplayground.run/)

```swift
// Default conversion will use int
print(500 / 1000) // 0 
// Wrong conversion to Double
print(Double(500 / 1000)) // 0.0 
// Good conversion to Double
print(500 / Double(1000)) // 0.5
```

Important bug :).


EArminjon:
Little wish : Can you after merging it releasing it on pub.dev ? <3 <3

navaronbracke (approved):

navaronbracke:
I'll release a new version with this fix tomorrow.

navaronbracke:
@EArminjon Version 3.5.3 is now available on pub.dev. Thank you for your contribution!

PR #872, opened 2023-11-22
fix: iOS Timeout is broken because of Double miss conversion - additional

EArminjon:
I apology, I didn't test enough and the latest change no longer compile. Here is the fix. 
As timeout is an Int, division seems to not accept `Int / Double` but it accept obviously `Double / Double`. 

This will fix :
https://github.com/juliansteenbakker/mobile_scanner/issues/873

Sorry again for this fail.

### DIFF
--- a/ios/Classes/MobileScannerPlugin.swift
+++ b/ios/Classes/MobileScannerPlugin.swift
@@ -103,7 +103,7 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin {
         let returnImage: Bool = (call.arguments as! Dictionary<String, Any?>)["returnImage"] as? Bool ?? false
         let speed: Int = (call.arguments as! Dictionary<String, Any?>)["speed"] as? Int ?? 0
         let timeoutMs: Int = (call.arguments as! Dictionary<String, Any?>)["timeout"] as? Int ?? 0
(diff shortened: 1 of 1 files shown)


memoriasIT (approved):

navaronbracke:
@EArminjon and @juliansteenbakker Apologies, I should have independently validated this as well. I was a bit time-constrained this morning so I didn't have the time.

I plan on adding integration tests, to catch this kind of thing in the future.

Version 3.5.3 has been retracted, to stop affecting existing customers. Will update to version 3.5.4 to rectify this.

And yes, this time, I did validate the fix myself as well.

navaronbracke:
As I independently validated the fix, I went ahead and merged #877 (which also includes a version bump).

EArminjon:
> @EArminjon and @juliansteenbakker Apologies, I should have independently validated this as well. I was a bit time-constrained this morning so I didn't have the time.
> 
> I plan on adding integration tests, to catch this kind of thing in the future.
> 
> Version 3.5.3 has been retracted, to stop affecting existing customers. Will update to version 3.5.4 to rectify this.
> 
> And yes, this time, I did validate the fix myself as well.

Totally my bad. I apology again, sorry.

PR #880, opened 2023-11-27
Fix android permission

### LINKED ISSUE #860: [Android] QR Scanner not working on second attempt of granting permissions
**Issue Summary:**
The QR scanner in the 'MobileScanner with Controller' example app fails to activate on an Android physical device after permissions are granted on the second attempt.

**Steps to Reproduce:**
1. Open the example app on an Android physical device.
2. Tap on the 'MobileScanner with Controller' button.
3. When prompted for camera permissions, choose "Deny".
4. Go back to the main screen of the app.
5. Tap on the 'MobileScanner with Controller' button again.
6. This time, when prompted for permissions, select "Allow while using the app".

**Expected Behavior:**
The QR scanner should activate and be ready to scan QR codes after the permissions are granted.

**Actual Behavior:**
After granting the permissions on the second attempt, the QR scanner does not activate.

**Additional Information:**
- Device Model: Pixel 4a
- Android Version: 13

navaronbracke:
This PR fixes a bug for Android permissions.

- Android does not have an undetermined permission state
- The check for the permission state wrongly returned `undetermined` for a denied state
- There was also a bug with the permission listener not being unset after it finished

Fixes #860

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
@@ -70,6 +70,7 @@ class MobileScannerPermissions {
                 object: ResultCallback {
                     override fun onResult(errorCode: String?, errorDescription: String?) {
                         ongoing = false
+                        listener = null
                         callback.onResult(errorCode, errorDescription)
                     }
                 }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
@@ -45,7 +45,7 @@ class MobileScannerPermissions {
         return if (hasPermission) {
             1
         } else {
-            0
+            2
         }
     }
 
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -206,32 +206,33 @@ class MobileScannerController {
       }
 
       switch (state) {
+        // Android does not have an undetermined permission state.
+        // So if the permission state is denied, just request it now.
         case MobileScannerState.undetermined:
-          bool result = false;
-
+        case MobileScannerState.denied:
           try {
-            result =
+            final bool granted =
                 await _methodChannel.invokeMethod('request') as bool? ?? false;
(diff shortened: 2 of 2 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
```diff
@@ -45,7 +45,7 @@ class MobileScannerPermissions {
         return if (hasPermission) {
             1
         } else {
-            0
+            2
```
navaronbracke, line 48: The integer 2 corresponds to `denied`. As Android only has `PERMISSION_GRANTED` and `PERMISSION_DENIED` as statuses, this value was wrong before.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
```diff
@@ -70,6 +70,7 @@ class MobileScannerPermissions {
                 object: ResultCallback {
                     override fun onResult(errorCode: String?, errorDescription: String?) {
                         ongoing = false
+                        listener = null
```
navaronbracke, line 73: Not having this would now result in a `permission request pending` for the second attempt.

lib/src/mobile_scanner_controller.dart
```diff
@@ -189,8 +189,9 @@ class MobileScannerController {
       final MobileScannerState state;
 
       try {
-        state = MobileScannerState
-            .values[await _methodChannel.invokeMethod('state') as int? ?? 0];
+        state = MobileScannerState.fromRawValue(
```
navaronbracke, line 192: This is cleanup

lib/src/mobile_scanner_controller.dart
```diff
@@ -205,32 +206,33 @@ class MobileScannerController {
       }
 
       switch (state) {
+        // Android does not have an undetermined permission state.
+        // So if the permission state is denied, just request it now.
         case MobileScannerState.undetermined:
-          bool result = false;
-
+        case MobileScannerState.denied:
```
navaronbracke, line 212: Since Android does not have an undetermined state, just request the permission again if it was denied, and then use the result of that call as the final state.

mcdev84:
Thanks @navaronbracke. As today 7/30/2024, using 5.1.1 version, it is still an issue.

PR #899, opened 2023-12-12
fix: Switch to iOS 13, use Tasks to handle threads and add error handling 

juliansteenbakker:
Continuation of #726


yeras-is:
Any updates on this PR ?  @juliansteenbakker

juliansteenbakker:
Since this PR needs a minimum of iOS 13, i am working on a implementation that also enabled iOS 12 to work. @yeras-is

navaronbracke:
I'm not that familiar with Tasks in Swift, but a quick search yields the following comment as a good summary:

https://developer.apple.com/forums/thread/712303?answerId=724093022#724093022

Since we are only using the task for the `start()` method, it's up to you to decide if we want it or not.
I'm fine with either option.

PR #913, opened 2023-12-27
fix: return type 'int' doesn't allow null for zxingBarcodeFormat method

### LINKED ISSUE #912: [flutter web] zxingBarcodeFormat: return type 'int' doesn't allow null
Hi with the current version of the package (`3.5.5`) app does not get executed on flutter web.

Flutter doctor
```
[✓] Flutter (Channel stable, 3.3.10, on macOS 13.6 22G120 darwin-arm, locale en-ES)
[✓] Android toolchain - develop for Android devices (Android SDK version 31.0.0-rc5)
[✓] Xcode - develop for iOS and macOS (Xcode 15.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2022.1)
[✓] Connected device (3 available)
[✓] HTTP Host Availability
```

Console shows the following error:
```
Launching lib/main_development.dart on Chrome in debug mode...
Waiting for connection from debug service on Chrome...
../../../../.pub-cache/hosted/pub.dartlang.org/mobile_scanner-3.5.5/lib/src/web/zxing.dart:106:11: Error: A non-null value must be returned since the return type 'int' doesn't allow null.
  int get zxingBarcodeFormat {
          ^
Failed to compile application.
```

I traced the error and realized that the switch case inside that method has not a `default` statement, so Dart thinks that a `null` value might be returned.

[Here is the PR](https://github.com/juliansteenbakker/mobile_scanner/pull/913) that fix this, i'll be using my forked version meanwhile.

hazzo:
fixes #912

### DIFF
--- a/lib/src/web/zxing.dart
+++ b/lib/src/web/zxing.dart
@@ -133,7 +133,7 @@ extension ZXingBarcodeFormat on BarcodeFormat {
         return 15;
       case BarcodeFormat.unknown:
       case BarcodeFormat.all:
-      case default:
+      default:
         return -1;
     }
   }
--- a/lib/src/web/zxing.dart
+++ b/lib/src/web/zxing.dart
@@ -133,6 +133,7 @@ extension ZXingBarcodeFormat on BarcodeFormat {
         return 15;
       case BarcodeFormat.unknown:
       case BarcodeFormat.all:
+      case default:
         return -1;
     }
   }


navaronbracke:
Thanks for your contribution. Makes sense to add a default case, the other conversion also had a default case too.

hazzo:
My pleasure 🙏🏼. Do you know when this will land on pub.dev?

navaronbracke:
I'll try to get this published tomorrow.

Issue #939, opened 2024-01-23
Make onDetect Field Nullable and Add `@Deprecated` Annotation 


utamori:
#925 #916

It is difficult to see that the onDetect field is deprecated.

- make it a nullable field
- Change the usage examples in the README and Example to not use onDetect.
- add  Deprecated Annotation

I'd be happy to help.

navaronbracke:
This is included in the new version 5.0.0 beta release.

PR #916, opened 2023-12-29
feat: Mobile scanner platform interface

navaronbracke:
Update: the issue with the constructor was fixed at HEAD in the Dart SDK. We have to wait until it gets cherry-picked into stable before landing this.

This PR refactors the mobile_scanner package to use the new `plugin_platform_interface`.

This refactor brings along a bunch of improvements, but also some breaking changes that make the public API easier to work with.

Breaking changes:
- The plugin no longer manages the lifecycle of the controller. This avoids many of the reported issues when the MobileScanner widget is reparented in the widget tree. As a result the related API for starting the scanner has been removed. This includes the API for `onPermissionSet`, `autoStart`, `onStart`, `onScannerStarted` and `startDelay`.
- A MobileScannerController is now required for the MobileScanner widget, per the above change.
- The `onDetect` method has been removed. Instead, listen to the `barcodes` stream of the controller.
- The `overlay` widget of the `MobileScanner` has been replaced by a new property, `overlayBuilder`, which provides the constraints for the overlay.
- The `hasTorch` getter has been removed. Instead, the `TorchState` now has a new value for the unavailable state.
- The `MobileScannerState` enum has been renamed to `MobileScannerAuthorizationState`, as the name `MobileScannerState` is now reserved for the model of the ChangeNotifier. See the entry in the improvements.
- The `formats` attribute of the `MobileScannerController` is now non-null. The new default value is an empty list, which is treated as all possible formats.
- The analyzeImage method now returns an instance of `BarcodeCapture?`, instead of a boolean.
- The `cameraFacingOverride` argument for the `MobileScannerController.start()` method has been renamed to `cameraDirection` to better reflect its use.
- The class `MobileScannerArguments` has been removed. Instead, use the properties in `MobileScannerState`.
- The `raw` property of a `BarcodeCapture` is now `Object?` instead of `dynamic`, so that it participates in type promotion (whereas the dynamic type did not)
- The `width` and `height` of `BarcodeCapture` have been removed, in favor of the `size` property.
- The web implementation now uses ZXing by default. Adding a newer version of JSQr can be added in the future.

Improvements:
- The `MobileScannerController` is now a ChangeNotifier, with `MobileScannerState` as its model. This consolidates all the different value notifiers into one place.
- Flutter 3.16.0 is now required.
- The web implementation has been refactored away from `dart:html` and now uses the new JS static interop / package:web. This refactor enables some extra features for the web in the future (Dart extension types, support for zoom scale / camera selection)
[...]

PR #940, opened 2024-01-23
feat: make onDetect nullable field

utamori:
relate #939

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -37,7 +37,7 @@ class MobileScanner extends StatefulWidget {
   final BoxFit fit;
 
   /// The function that signals when new codes were detected by the [controller].
-  final void Function(BarcodeCapture barcodes) onDetect;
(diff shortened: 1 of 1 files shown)


lib/src/mobile_scanner.dart
```diff
@@ -37,7 +37,7 @@ class MobileScanner extends StatefulWidget {
   final BoxFit fit;
 
   /// The function that signals when new codes were detected by the [controller].
-  final void Function(BarcodeCapture barcodes) onDetect;
+  final void Function(BarcodeCapture barcodes)? onDetect;
```
stephg, line 40: I think some hint for the recommended way of listening to detected barcodes should be added here.

Or deprecate the use of this property?

utamori: Yes. This field will be removed in the next version, so it is better to add Deprecated annotation

navaronbracke:
This is no longer needed with #916 , as that PR removes the argument altogether.

Issue #922, opened 2024-01-08
Document new Android minimum requirements in the README (Java version / minSdk / compileSdk)


burekas7:
The package suddenly requires Android SDK version 34 even without updating it.
I'm using version 3.5.2 now which works all the time, but suddenly I start to get this requirement.

Probably after clear the cache.

I can't update my sdk now to 34 because I have a problem with other packages.

How can I solve it?
How can I use mobile_scanner with sdk 33?

Thanks

burekas7:
```

Warning: The plugin mobile_scanner requires Android SDK version 34.
For more information about build configuration, see https://docs.flutter.dev/deployment/android#reviewing-the-gradle-build-configuration.
One or more plugins require a higher Android SDK version.
Fix this issue by adding the following to /Users/aditzoref/vscodeProjects/Deliveright/mobile_xp_driver/android/app/build.gradle:
android {
  compileSdkVersion 34
  ...
}



FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:checkDebugAarMetadata'.
> A failure occurred while executing com.android.build.gradle.internal.tasks.CheckAarMetadataWorkAction
   > One or more issues found when checking AAR metadata values:

     The minCompileSdk (34) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-33).
     Dependency: androidx.camera:camera-core:1.3.1.
     AAR metadata file: /Users/------/.gradle/caches/transforms-3/090e3908a3264c467bd84b2d12b9d4fc/transformed/jetified-camera-core-1.3.1/META-INF/com/android/build/gradle/aar-metadata.properties.

     The minCompileSdk (34) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-33).
     Dependency: androidx.camera:camera-lifecycle:1.3.1.
     AAR metadata file: /Users/------/.gradle/caches/transforms-3/f34c10f8d10da3bc85b5f2fe44189204/transformed/jetified-camera-lifecycle-1.3.1/META-INF/com/android/build/gradle/aar-metadata.properties.

     The minCompileSdk (34) specified in a
     dependency's AAR metadata (META-INF/com/android/build/gradle/aar-metadata.properties)
     is greater than this module's compileSdkVersion (android-33).
[...]

PR #946, opened 2024-01-30
docs: Add note about compileSDK requirements

vbuberen:
Added a note to mark which versions would require Android projects to have compileSDK 34.

The reason for this change is because since `3.5.0` the plugin uses dependencies that require compileSDK to be set to 34 as I have already explained in https://github.com/juliansteenbakker/mobile_scanner/issues/922#issuecomment-1912090475

With this change, hopefully, there will be less confused people as versions before 4.0.0 don't mention such requirement, so, potentially, there might be more issues like #922

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,7 +61,7 @@ Bugs fixed:
 
 ## 3.5.0
 
-**NOTE: Since this release plugin requires your project to have compileSDK 34 (Android 14)**
+**NOTE: From this version onwards, `mobile_scanner` requires Android projects to have a `compileSdk` of 34 (Android 14) or higher**
 
 New Features:
 * Added the option to switch between bundled and unbundled MLKit for Android. (thanks @woolfred !)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,15 +5,11 @@ BREAKING CHANGES:
 
 ## 3.5.7
 
-NOTE: This version requires your project to have compileSDK 34 on Android
-
 Improvements:
 * Updated js dependency together with other dependencies.
 * Reverted compileSdk version to 33 on Android. This update will be released under version 4.0.0.
 
 ## 3.5.6
-
-NOTE: This version requires your project to have compileSDK 34 on Android
 
 Bugs fixed:
 * [web] Fixed a crash with the ZXing barcode format (thanks @hazzo!)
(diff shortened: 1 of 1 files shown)


navaronbracke:
I find it a bit weird to have the note for every 3.5.x release? You could just add it to the 3.5.0 release notes?
It is up to end users to read the changelogs of packages they upgrade in my opinion.

Also, perhaps put the message in ** so it is bold in the Markdown. Then it also shows up bold on pub.dev

vbuberen:
> I find it a bit weird to have the note for every 3.5.x release?

Yes, I also don't like it, but 3.5.0 had to be 4.0.0. I believe there is still a place for confusion for some users, like #922. 

> It is up to end users to read the changelogs of packages they upgrade in my opinion.

You know it well as you are also a maintainer of some package like I am that some users are too lazy and don't read anything.

vbuberen:
> Also, perhaps put the message in ** so it is bold in the Markdown. Then it also shows up bold on pub.dev

Thanks, I know how to do it in Markdown, but I checked how release notes were formatted in this package before and never saw a single bold font usage. This is why I didn't make it bold right away. But I will update it.

navaronbracke:
True, we cannot make any assumptions on what our end users might (not) do.

And, yes we should be making the breaking changes notes bold.
Feel free to correct that as well.

vbuberen:
I have just updated to have only one bold mention in 3.5.0 about compile SDK. Let me know if it looks better now.

CHANGELOG.md
```diff
@@ -60,6 +66,9 @@ Bugs fixed:
 * Fixed messages not being sent on the main thread for Android, iOS and MacOS. (thanks @navaronbracke !)
 
 ## 3.5.0
+
+**NOTE: Since this release plugin requires your project to have compileSDK 34 (Android 14)**
```
navaronbracke, line 64:
```suggestion
**NOTE: From this version onwards, `mobile_scanner` requires Android projects to have a `compileSdk` of 34 (Android 14) or higher**
```

navaronbracke (changes requested):
I would suggest a small change in the wording.

CHANGELOG.md
vbuberen: As you wish

navaronbracke (approved):

navaronbracke:
I will be patching up a small bug on the iOS side of things, as 4.0.1, and then this updated changelog will be available to end users.

PR #956, opened 2024-02-08
refactor: Print Statements

### LINKED ISSUE #780: Clean up the use of print statements in the plugin
The plugin has a bunch of print / debugPrint statements in the code.

These should either be removed, or turned into proper thrown errors (MethodChannel result/ event sink or otherwise)

Erengun:
Resolves #780


navaronbracke:
@Erengun Thanks for your contribution. However, with my work on migrating the plugin to the platform interface (which is still pending a small bugfix), I think we will no longer need this PR, as I have cleaned up the print statement usage there.

navaronbracke:
This is no longer needed, in #916 we removed all print statements (except 2 for logging of issues with the image buffer on MacOS)

PR #1012, opened 2024-04-13
feat: Add setting useUnbundled from the pubspec.yaml file

jawa0919:
in gradle use `org.yaml:snakeyaml`  read `pubspec.yaml` find `useUnbundled ` value. choose implementation


navaronbracke:
Thanks for your contribution.

However, this is not something we are going to accept. First of all, you removed a lot of working code, which doesn't impact your new option. Secondly, adding a new dependency for `org.yaml:snakeyaml` is not needed, since the existing option from Gradle itself worked fine.

navaronbracke:
The Flutter plugin doesn't need to decide which version to use? It is the developer of the consuming _application_ that decides which version needs to be used. The plugin on its own is not executable as an application.  Considering `add2app` use cases, I would expect you to be able to communicate with the team of the host Android project? So I think it is appropriate?

The end result - having a configurable property somewhere in the project - stays the same with both approaches (the old way and your new implementation), but your new implementation adds a new dependency, which might need updating over time, as other people (who do not use add2app, which is the majority) might run into dependency issues with `snakeyaml`.

PR #1032, opened 2024-04-23
feat: change platform os detect pattern

penkzhou:
current code in `lib/src/method_channel/mobile_scanner_method_channel.dart` cannot change the platform in the test code when need to change the platform. according to this thread, maybe we should change the platform os detect pattern into `defaultTargetPlatform `. 
ref:https://stackoverflow.com/questions/55233853/flutter-widget-tests-with-platform-branching

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 ## 5.0.1
 
 Improvements:
-* Change platform OS detection pattern. According to the official Dart documentation, the `dart:io.Platform` object should only be used directly when it's critical to detect the actual current platform, without allowing overrides (for example, when calling native system APIs). With this change, we can override the platform in test code.
+* Adjusted the platform checks to use the defaultTargetPlatform API, so that tests can use the correct platform overrides.
 
 ## 5.0.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 5.0.1
+
+Improvements:
+* Change platform OS detection pattern. According to the official Dart documentation, the `dart:io.Platform` object should only be used directly when it's critical to detect the actual current platform, without allowing overrides (for example, when calling native system APIs). With this change, we can override the platform in test code.
+
 ## 5.0.0
 
 This major release contains all the changes from the 5.0.0 beta releases, along with the following changes:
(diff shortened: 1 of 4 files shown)


lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -54,7 +54,7 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
     final List<Map<Object?, Object?>> barcodes =
         data.cast<Map<Object?, Object?>>();
 
-    if (Platform.isMacOS) {
+    if (defaultTargetPlatform == TargetPlatform.macOS) {
```
navaronbracke, line 57: There is also one usage in https://github.com/juliansteenbakker/mobile_scanner/blob/8c87026a00ab38ca116cb1f51f0ed145738cd2c9/example/lib/barcode_scanner_window.dart#L207

navaronbracke (changes requested):

navaronbracke:
@penkzhou Just to be sure, this is primarily for people that write tests, right?

I had a look at the implementation of `defaultTargetPlatform` and that uses the `Platform.is`  getters under the hood.

navaronbracke:
You are correct, yes. In these cases, we already called the native side and we just parse the result.
There is no "system level API" that we needed to call here.

navaronbracke (approved):
If you want to update the version to `5.0.1` and update the changelog to include a mention of this fix, that would be helpful.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 5.0.1
+
+Improvements:
+* Change platform OS detection pattern. According to the official Dart documentation, the `dart:io.Platform` object should only be used directly when it's critical to detect the actual current platform, without allowing overrides (for example, when calling native system APIs). With this change, we can override the platform in test code.
```
navaronbracke, line 4: I'd prefer to keep the implementation details private. (since they can change) What about

```
Adjusted the platform checks to use the defaultTargetPlatform API, so that tests can use the correct platform overrides.
```

navaronbracke (changes requested):

navaronbracke (approved):

PR #1070, opened 2024-05-17
fix: connecting the useNewCameraSelector option

bswhite1:
useNewCameraSelector is not being sent to the android backend. This forwards it with the other start options.

### DIFF
--- a/lib/src/objects/start_options.dart
+++ b/lib/src/objects/start_options.dart
@@ -14,6 +14,7 @@ class StartOptions {
     required this.formats,
     required this.returnImage,
     required this.torchEnabled,
+    required this.useNewCameraSelector,
   });
 
   /// The direction for the camera.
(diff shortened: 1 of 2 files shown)


lib/src/objects/start_options.dart
```diff
@@ -37,6 +38,10 @@ class StartOptions {
   /// Whether the torch should be turned on when the scanner starts.
   final bool torchEnabled;
 
+  /// Only supported on Android.
```
navaronbracke, line 41:
```suggestion
  /// Whether the new resolution selector should be used.
  ///
  /// This option is only supported on Android. Other platforms will ignore this option.
```

navaronbracke (commented):
Minor nit for the docs.

lib/src/objects/start_options.dart
bswhite1: updated.

PR #1071, opened 2024-05-20
feat: read inverted datamatrix

petri-lipponen-movesense:
Here's a quick stab at adding support for reading inverted data matrixes that the MLKit does not support. The iOS implementation is a bit clumsy (I'm not an iOS coder), but Android one is quite straight forward. Tested and works in Android & iOS. Feel free to modify, include or leave be as you will.

Future idea: add "tryInverted" which duplicates each frame and tries normal and inverted both.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -80,6 +80,9 @@ dependencies {
     implementation 'androidx.camera:camera-lifecycle:1.3.3'
     implementation 'androidx.camera:camera-camera2:1.3.3'
 
+    // opencv
+    implementation 'com.quickbirdstudios:opencv:3.4.15'
+
     testImplementation 'org.jetbrains.kotlin:kotlin-test'
     testImplementation 'org.mockito:mockito-core:5.12.0'    
 }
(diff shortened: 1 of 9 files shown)


navaronbracke:
Interesting. Did you manage to do some performance profiling on this? Inverting each frame might be expensive to do (I think). A bit sad we have to pull in `opencv` for this to work

petri-lipponen-movesense:
> Interesting. Did you manage to do some performance profiling on this? Inverting each frame might be expensive to do (I think). A bit sad we have to pull in `opencv` for this to work

No I didn't. We need it so short time, that it really makes no difference, and it didn't feel slow on phones tested. 
I agree, it really is a bummer that the opencv is needed. There has to be better way to do it on both systems, but I just needed something that works with minimum effort and those came up on stackoverflow 😁 .

RafaRuiz:
I remember doing this for a client of mine, on a native solution, I had to use opencv too.

RafaRuiz:
After creating #1214 to enhance your PR, on Android I've seen that the inverted image is shrink, I'm working on fixing it on my branch and I will create a PR on yours as well @petri-lipponen-movesense 
image

juliansteenbakker:
Thank you for your work, and my apologies for my late reply. I think we will be going forward with #1215 due to no external package being necessary there.

Issue #1078, opened 2024-05-27
Camera is not working


kcs-ravindrapatil:
Hello folks I have facing some issue while opening scanner, when I opening scanner then it will display info icon instead of camera.

![image](https://github.com/juliansteenbakker/mobile_scanner/assets/113008846/d56cf131-f2eb-44a3-9a89-7d1b1f4996ea)

camera is not detect properly and not scanning, am providing some code

### Declaration/Functions/Methods:

```
  @override
  void dispose() {
    cameraController.stop();
    cameraController.dispose();
    WidgetsBinding.instance.removeObserver(this);
    super.dispose();
  }

  @override
  void initState() {
    super.initState();
    WidgetsBinding.instance.addObserver(this);
    startCamera();
  }

  @override
  void didChangeAppLifecycleState(AppLifecycleState state) {
    if (state == AppLifecycleState.inactive || state == AppLifecycleState.paused) {
      cameraController.stop();
      print("Lifecycle stop is call....");
    } else if (state == AppLifecycleState.resumed) {
      startCamera();
      print("Lifecycle resume is call....");
    }
  }

  Future<void> startCamera() async {
    if (cameraController != null) {
      await cameraController.start();
    }
  }

  void resetScanner() async {
    await cameraController.stop();
    await Future.delayed(const Duration(milliseconds: 500)); // Wait for the scanner to stop
    await cameraController.start();
  }

```
### UI: 
```
MobileScanner(
                                fit: BoxFit.fill,
                                controller: MobileScannerController(detectionSpeed: DetectionSpeed.noDuplicates),
                                onDetect: (capture) {
                                  print("-->>" + capture.barcodes.toString());
                                  final List<Barcode> barcodes = capture.barcodes;
                                  for (final barcode in barcodes) {
                                    if (barcode.rawValue != null) {
                                      print("Barcode value: ${barcode.rawValue!}");
                                    // Going to next screen after successfully scanned qr 
                                     } else {
                                      print("Something Wrong");
                                    }
                                  }
                                },
                              ),
```

navaronbracke:
The black screen with (i) icon is the default error builder for the `MobileScanner` widget.
What error details (message + stacktrace) do you get in the `MobileScannerException` ? (you can inspect it using the `MobileScanner.errorBuilder` property)

kcs-ravindrapatil:
@navaronbracke Yes I know this is error builder but it gives the below error 

[...]

Issue #502, opened 2023-02-20
Hot restart not working on iOS (v3.0.0)


zunimc:
I have this simple code to start. I removed the torchlight and front camera buttons for simplicity.

I am testing it on real devices (iPhone 6S and iPhone 12).

```
import 'package:flutter/material.dart';
import 'package:mobile_scanner/mobile_scanner.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  // This widget is the root of your application.
  @override
  Widget build(BuildContext context) {
    MobileScannerController cameraController = MobileScannerController();
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        primarySwatch: Colors.blue,
      ),
      home: Scaffold(
        appBar: AppBar(
          title: const Text('Mobile Scanner'),
        ),
        body: MobileScanner(
          controller: cameraController,
          onDetect: (capture) {
            final List<Barcode> barcodes = capture.barcodes;
            for (final barcode in barcodes) {
              debugPrint('Barcode found! ${barcode.rawValue}');
            }
          },
        ),
      ),
    );
  }
}
```

Problem is as soon as I hit Hot Restart (Shift+R) camera stops working.

`Performing hot restart...                                               
Restarted application in 1.097ms.
flutter: mobile_scanner: MobileScannerException: code genericError, message: Called start() while already started!`

godspeed0310:
I'm having the same issue on Android, were you able to find a solution for this?

zunimc:
> I'm having the same issue on Android, were you able to find a solution for this?

Android? That's weird as that OS worked like a charm on my code. Anyway I ended up replacing this library and found a working one: https://pub.dev/packages/code_scan

With very few adjustments I got the same result on both mobile systems. Hope that helps.

godspeed0310:
Is this the same code that you used for android too? And yeah it does help, thanks for the library recommendation

zunimc:
Yes, same code for both iOS and Android.

`FutureBuilder<double>(
  future: setScreenHeight(), // this functions returns a custom height for each device.
  builder: (context, snapshot) {
    return SizedBox(
      height: snapshot.data, // result of setScreenHeight()
      child: code_scan.CodeScanner(
        onScan: (code, details, controller) {
          setState(() {
            checkBarcodeFormat(code!);
          });
        },
[...]

PR #1086, opened 2024-06-03
fix: Ignore additional calls to start() when the controller is starting

navaronbracke:
Fixes: https://github.com/juliansteenbakker/mobile_scanner/issues/1078
Fixes: https://github.com/juliansteenbakker/mobile_scanner/issues/502

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/BarcodeHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/BarcodeHandler.kt
@@ -18,15 +18,15 @@ class BarcodeHandler(binaryMessenger: BinaryMessenger) : EventChannel.StreamHand
         eventChannel.setStreamHandler(this)
     }
 
-    fun publishEvent(event: Map<String, Any>) {
-        Handler(Looper.getMainLooper()).post {
-            eventSink?.success(event)
-        }
-    }
-
     fun publishError(errorCode: String, errorMessage: String, errorDetails: Any?) {
         Handler(Looper.getMainLooper()).post {
             eventSink?.error(errorCode, errorMessage, errorDetails)
+        }
+    }
+
+    fun publishEvent(event: Map<String, Any>) {
+        Handler(Looper.getMainLooper()).post {
+            eventSink?.success(event)
         }
     }
 
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/BarcodeHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/BarcodeHandler.kt
@@ -24,6 +24,12 @@ class BarcodeHandler(binaryMessenger: BinaryMessenger) : EventChannel.StreamHand
         }
     }
 
+    fun publishError(errorCode: String, errorMessage: String, errorDetails: Any?) {
+        Handler(Looper.getMainLooper()).post {
+            eventSink?.error(errorCode, errorMessage, errorDetails)
+        }
+    }
+
     override fun onListen(event: Any?, eventSink: EventChannel.EventSink?) {
         this.eventSink = eventSink
     }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -67,7 +67,7 @@ class MobileScannerHandler(
 
     private val errorCallback: MobileScannerErrorCallback = {error: String ->
         barcodeHandler.publishEvent(mapOf(
-            "name" to "error",
+            "name" to MobileScannerErrorCodes.BARCODE_ERROR,
             "data" to error,
         ))
     }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -29,7 +29,7 @@ class MobileScannerHandler(
 
     private val analyzeImageErrorCallback: AnalyzerErrorCallback = {
         Handler(Looper.getMainLooper()).post {
-            analyzerResult?.error(MobileScannerErrorCodes.GENERIC_ERROR, it, null)
+            analyzerResult?.error(MobileScannerErrorCodes.BARCODE_ERROR, it, null)
             analyzerResult = null
         }
     }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -10,6 +10,7 @@ import androidx.camera.core.ExperimentalGetImage
 import com.google.mlkit.vision.barcode.BarcodeScannerOptions
 import dev.steenbakker.mobile_scanner.objects.BarcodeFormats
 import dev.steenbakker.mobile_scanner.objects.DetectionSpeed
+import dev.steenbakker.mobile_scanner.objects.MobileScannerErrorCodes
 import io.flutter.plugin.common.BinaryMessenger
 import io.flutter.plugin.common.MethodCall
 import io.flutter.plugin.common.MethodChannel
@@ -28,7 +29,7 @@ class MobileScannerHandler(
 
     private val analyzeImageErrorCallback: AnalyzerErrorCallback = {
         Handler(Looper.getMainLooper()).post {
-            analyzerResult?.error("MobileScanner", it, null)
+            analyzerResult?.error(MobileScannerErrorCodes.GENERIC_ERROR, it, null)
             analyzerResult = null
         }
     }
@@ -106,21 +107,21 @@ class MobileScannerHandler(
 
     @ExperimentalGetImage
     override fun onMethodCall(call: MethodCall, result: MethodChannel.Result) {
-        if (mobileScanner == null) {
-            result.error("MobileScanner", "Called ${call.method} before initializing.", null)
-            return
-        }
         when (call.method) {
             "state" -> result.success(permissions.hasCameraPermission(activity))
             "request" -> permissions.requestPermission(
                 activity,
                 addPermissionListener,
                 object: MobileScannerPermissions.ResultCallback {
-                    override fun onResult(errorCode: String?, errorDescription: String?) {
+                    override fun onResult(errorCode: String?) {
                         when(errorCode) {
                             null -> result.success(true)
-                            MobileScannerPermissions.CAMERA_ACCESS_DENIED -> result.success(false)
-                            else -> result.error(errorCode, errorDescription, null)
+                            MobileScannerErrorCodes.CAMERA_ACCESS_DENIED -> result.success(false)
+                            MobileScannerErrorCodes.CAMERA_PERMISSIONS_REQUEST_ONGOING -> result.error(
+                                MobileScannerErrorCodes.CAMERA_PERMISSIONS_REQUEST_ONGOING,
+                                MobileScannerErrorCodes.CAMERA_PERMISSIONS_REQUEST_ONGOING_MESSAGE, null)
+                            else -> result.error(
+                                MobileScannerErrorCodes.GENERIC_ERROR, MobileScannerErrorCodes.GENERIC_ERROR_MESSAGE, null)
                         }
                     }
                 })
@@ -185,29 +186,29 @@ class MobileScannerHandler(
                     when (it) {
                         is AlreadyStarted -> {
                             result.error(
-                                "MobileScanner",
-                                "Called start() while already started",
+                                MobileScannerErrorCodes.ALREADY_STARTED_ERROR,
+                                MobileScannerErrorCodes.ALREADY_STARTED_ERROR_MESSAGE,
                                 null
                             )
                         }
                         is CameraError -> {
                             result.error(
-                                "MobileScanner",
-                                "Error occurred when setting up camera!",
+                                MobileScannerErrorCodes.CAMERA_ERROR,
+                                MobileScannerErrorCodes.CAMERA_ERROR_MESSAGE,
                                 null
                             )
                         }
                         is NoCamera -> {
                             result.error(
-                                "MobileScanner",
-                                "No camera found or failed to open camera!",
+                                MobileScannerErrorCodes.NO_CAMERA_ERROR,
+                                MobileScannerErrorCodes.NO_CAMERA_ERROR_MESSAGE,
                                 null
                             )
                         }
                         else -> {
                             result.error(
-                                "MobileScanner",
-                                "Unknown error occurred.",
+                                MobileScannerErrorCodes.GENERIC_ERROR,
+                                MobileScannerErrorCodes.GENERIC_ERROR_MESSAGE,
                                 null
                             )
                         }
@@ -252,9 +253,11 @@ class MobileScannerHandler(
             mobileScanner!!.setScale(call.arguments as Double)
             result.success(null)
         } catch (e: ZoomWhenStopped) {
-            result.error("MobileScanner", "Called setScale() while stopped!", null)
+            result.error(
+                MobileScannerErrorCodes.SET_SCALE_WHEN_STOPPED_ERROR, MobileScannerErrorCodes.SET_SCALE_WHEN_STOPPED_ERROR_MESSAGE, null)
         } catch (e: ZoomNotInRange) {
-            result.error("MobileScanner", "Scale should be within 0 and 1", null)
+            result.error(
+                MobileScannerErrorCodes.GENERIC_ERROR, MobileScannerErrorCodes.INVALID_ZOOM_SCALE_ERROR_MESSAGE, null)
         }
     }
 
@@ -263,7 +266,8 @@ class MobileScannerHandler(
             mobileScanner!!.resetScale()
             result.success(null)
         } catch (e: ZoomWhenStopped) {
-            result.error("MobileScanner", "Called resetScale() while stopped!", null)
+            result.error(
+                MobileScannerErrorCodes.SET_SCALE_WHEN_STOPPED_ERROR, MobileScannerErrorCodes.SET_SCALE_WHEN_STOPPED_ERROR_MESSAGE, null)
         }
     }
 
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -66,10 +66,7 @@ class MobileScannerHandler(
     }
 
     private val errorCallback: MobileScannerErrorCallback = {error: String ->
-        barcodeHandler.publishEvent(mapOf(
-            "name" to MobileScannerErrorCodes.BARCODE_ERROR,
-            "data" to error,
-        ))
+        barcodeHandler.publishError(MobileScannerErrorCodes.BARCODE_ERROR, error, null)
     }
 
     private var methodChannel: MethodChannel? = null
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
@@ -5,6 +5,7 @@ import android.app.Activity
 import android.content.pm.PackageManager
 import androidx.core.app.ActivityCompat
 import androidx.core.content.ContextCompat
+import dev.steenbakker.mobile_scanner.objects.MobileScannerErrorCodes
 import io.flutter.plugin.common.PluginRegistry.RequestPermissionsResultListener
 
 /**
@@ -12,11 +13,6 @@ import io.flutter.plugin.common.PluginRegistry.RequestPermissionsResultListener
  */
 class MobileScannerPermissions {
     companion object {
-        const val CAMERA_ACCESS_DENIED = "CameraAccessDenied"
-        const val CAMERA_ACCESS_DENIED_MESSAGE = "Camera access permission was denied."
-        const val CAMERA_PERMISSIONS_REQUEST_ONGOING = "CameraPermissionsRequestOngoing"
-        const val CAMERA_PERMISSIONS_REQUEST_ONGOING_MESSAGE = "Another request is ongoing and multiple requests cannot be handled at once."
-
         /**
          * When the application's activity is [androidx.fragment.app.FragmentActivity], requestCode can only use the lower 16 bits.
          * @see androidx.fragment.app.FragmentActivity.validateRequestPermissionsRequestCode
@@ -25,7 +21,7 @@ class MobileScannerPermissions {
     }
 
     interface ResultCallback {
-        fun onResult(errorCode: String?, errorDescription: String?)
+        fun onResult(errorCode: String?)
     }
 
     private var listener: RequestPermissionsResultListener? = null
@@ -53,14 +49,13 @@ class MobileScannerPermissions {
                           addPermissionListener: (RequestPermissionsResultListener) -> Unit,
                           callback: ResultCallback) {
         if (ongoing) {
-            callback.onResult(
-                CAMERA_PERMISSIONS_REQUEST_ONGOING, CAMERA_PERMISSIONS_REQUEST_ONGOING_MESSAGE)
+            callback.onResult(MobileScannerErrorCodes.CAMERA_PERMISSIONS_REQUEST_ONGOING)
             return
         }
 
         if(hasCameraPermission(activity) == 1) {
             // Permissions already exist. Call the callback with success.
-            callback.onResult(null, null)
+            callback.onResult(null)
             return
         }
 
@@ -68,10 +63,10 @@ class MobileScannerPermissions {
             // Keep track of the listener, so that it can be unregistered later.
             listener = MobileScannerPermissionsListener(
                 object: ResultCallback {
-                    override fun onResult(errorCode: String?, errorDescription: String?) {
+                    override fun onResult(errorCode: String?) {
                         ongoing = false
                         listener = null
-                        callback.onResult(errorCode, errorDescription)
+                        callback.onResult(errorCode)
                     }
                 }
             )
--- a/example/.gitignore
+++ b/example/.gitignore
@@ -5,9 +5,11 @@
 *.swp
 .DS_Store
 .atom/
+.build/
 .buildlog/
 .history
 .svn/
+.swiftpm/
 migrate_working_dir/
 
 # IntelliJ related
--- a/example/ios/Runner.xcodeproj/project.pbxproj
+++ b/example/ios/Runner.xcodeproj/project.pbxproj
@@ -488,14 +488,14 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = 75Y2P2WSQQ;
+				DEVELOPMENT_TEAM = "";
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Runner/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "com.example.mobile-scanner";
+				PRODUCT_BUNDLE_IDENTIFIER = "com.example.mobile-scanner-example";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SWIFT_OBJC_BRIDGING_HEADER = "Runner/Runner-Bridging-Header.h";
@@ -670,14 +670,14 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = 75Y2P2WSQQ;
+				DEVELOPMENT_TEAM = "";
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Runner/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = "com.example.mobile-scanner";
+				PRODUCT_BUNDLE_IDENTIFIER = "com.example.mobile-scanner-example";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				SWIFT_OBJC_BRIDGING_HEADER = "Runner/Runner-Bridging-Header.h";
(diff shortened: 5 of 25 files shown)


simplenotezy:
Nice!

You probably need to update the examples folder as well, as I had to now manually call start() after setting the controller in initState().

navaronbracke:
@simplenotezy Ah right, I'll probably have to tweak the automatic start in the samples.

Just to be clear, you want automatic start to be _enabled_ for the samples? Or not?

simplenotezy:
Sounds good! After trying your version, I needed to call start manually. Otherwise the example code would not work.

```diff
class _TicketScannerScreenState extends State<TicketScannerScreen> {
  late MobileScannerController controller;

  @override
  void initState() {
    super.initState();
    controller = MobileScannerController(
      formats: const [BarcodeFormat.qrCode],
    );

+    controller.start();
  }

  @override
  Future<void> dispose() async {
    super.dispose();
    await controller.dispose();
  }

  @override
  Widget build(BuildContext context) {
    final scanWindow = Rect.fromCenter(
      center: MediaQuery.sizeOf(context).center(Offset.zero),
      width: 400,
      height: 200,
    );

**Comment by simplenotezy** [comment]
Hmm - something strange is going on in the repository code. Same issue I had before, but this time, there's just no error. It just wont work - black screen.

Image description



```dart
import 'package:auto_route/auto_route.dart';

@RoutePage()
class TicketScannerScreen extends StatefulWidget {
  const TicketScannerScreen({super.key});

  @override
  State<TicketScannerScreen> createState() => _TicketScannerScreenState();
}

class _TicketScannerScreenState extends State<TicketScannerScreen> {
  late MobileScannerController controller;

  @override
  void initState() {
    super.initState();
    controller = MobileScannerController(
      formats: const [BarcodeFormat.qrCode],
    );

    controller.start();
  }

  @override
  Future<void> dispose() async {
    super.dispose();
    await controller.dispose();
  }

  @override
    final scanWindow = Rect.fromCenter(
      center: MediaQuery.sizeOf(context).center(Offset.zero),
      width: 400,
      height: 200,
    );

    return Scaffold(
      backgroundColor: Colors.black,
      appBar: AppBar(
        title: Text('Scan'.hardcoded),
      ),
      body: Stack(
        fit: StackFit.expand,
        children: [
          Center(
            child: MobileScanner(
              fit: BoxFit.contain,
              controller: controller,
              scanWindow: scanWindow,
              errorBuilder: (context, error, child) {
                return ScannerErrorWidget(error: error);
              },
              overlayBuilder: (context, constraints) {
                return Padding(
                  padding: const EdgeInsets.all(16.0),
                  child: Align(
                    alignment: Alignment.bottomCenter,
                    child: ScannedBarcodeLabel(barcodes: controller.barcodes),
                  ),
                );
              },
            ),
          ),
          ValueListenableBuilder(
            valueListenable: controller,
            builder: (context, value, child) {
              if (!value.isInitialized ||
                  !value.isRunning ||
                  value.error != null) {
                return const SizedBox();
              }

              return CustomPaint(
                painter: ScannerOverlay(scanWindow: scanWindow),
              );
            },
          ),
          Align(
            alignment: Alignment.bottomCenter,
            child: Padding(
              padding: const EdgeInsets.all(16.0),
              child: Row(
                mainAxisAlignment: MainAxisAlignment.spaceEvenly,
                children: [
                  ToggleFlashlightButton(controller: controller),
                  SwitchCameraButton(controller: controller),
                ],
              ),
            ),
          ),
        ],
      ),
    );
  }
}

class ScannerOverlay extends CustomPainter {
  const ScannerOverlay({
    required this.scanWindow,
    this.borderRadius = 12.0,
  });

  final Rect scanWindow;
  final double borderRadius;

  @override
  void paint(Canvas canvas, Size size) {
    // TODO: use `Offset.zero & size` instead of Rect.largest
    // we need to pass the size to the custom paint widget
    final backgroundPath = Path()..addRect(Rect.largest);

    final cutoutPath = Path()
      ..addRRect(
        RRect.fromRectAndCorners(
          scanWindow,
          topLeft: Radius.circular(borderRadius),
          topRight: Radius.circular(borderRadius),
          bottomLeft: Radius.circular(borderRadius),
          bottomRight: Radius.circular(borderRadius),
        ),
      );

    final backgroundPaint = Paint()
      ..color = Colors.black.withOpacity(0.3)
      ..style = PaintingStyle.fill
      ..blendMode = BlendMode.dstOut;

    final backgroundWithCutout = Path.combine(
      PathOperation.difference,
      backgroundPath,
      cutoutPath,
    );

    final borderPaint = Paint()
      ..color = Colors.white
      ..style = PaintingStyle.stroke
      ..strokeWidth = 4.0;

    final borderRect = RRect.fromRectAndCorners(
      scanWindow,
      topLeft: Radius.circular(borderRadius),
      topRight: Radius.circular(borderRadius),
      bottomLeft: Radius.circular(borderRadius),
      bottomRight: Radius.circular(borderRadius),
    );

    // First, draw the background,
    // with a cutout area that is a bit larger than the scan window.
    // Finally, draw the scan window itself.
    canvas.drawPath(backgroundWithCutout, backgroundPaint);
    canvas.drawRRect(borderRect, borderPaint);
  }

  @override
  bool shouldRepaint(ScannerOverlay oldDelegate) {
    return scanWindow != oldDelegate.scanWindow ||
        borderRadius != oldDelegate.borderRadius;
  }
}

class ScannerErrorWidget extends StatelessWidget {
  const ScannerErrorWidget({super.key, required this.error});

  final MobileScannerException error;

  @override
    String errorMessage;

    switch (error.errorCode) {
      case MobileScannerErrorCode.controllerUninitialized:
        errorMessage = 'Controller not ready.';
      case MobileScannerErrorCode.permissionDenied:
        errorMessage = 'Permission denied';
      case MobileScannerErrorCode.unsupported:
        errorMessage = 'Scanning is unsupported on this device';
      default:
        errorMessage = 'Generic Error';
        break;
    }

    return ColoredBox(
      color: Colors.black,
      child: Center(
        child: Column(
          mainAxisSize: MainAxisSize.min,
          children: [
            const Padding(
              padding: EdgeInsets.only(bottom: 16),
              child: Icon(Icons.error, color: Colors.white),
            ),
            Text(
              errorMessage,
              style: const TextStyle(color: Colors.white),
            ),
            Text(
              error.errorDetails?.message ?? '',
              style: const TextStyle(color: Colors.white),
            ),
          ],
        ),
      ),
    );
  }
}

class AnalyzeImageFromGalleryButton extends StatelessWidget {
  const AnalyzeImageFromGalleryButton({required this.controller, super.key});

  final MobileScannerController controller;

  @override
    return IconButton(
      color: Colors.white,
      icon: const Icon(Icons.image),
      iconSize: 32.0,
      onPressed: () async {
        // final ImagePicker picker = ImagePicker();

        // final XFile? image = await picker.pickImage(
        //   source: ImageSource.gallery,
        // );

        // if (image == null) {
        //   return;
        // }

        // final BarcodeCapture? barcodes = await controller.analyzeImage(
        //   image.path,
        // );

        // if (!context.mounted) {
        //   return;
        // }

        // final SnackBar snackbar = barcodes != null
        //     ? const SnackBar(
        //         content: Text('Barcode found!'),
        //         backgroundColor: Colors.green,
        //       )
        //     : const SnackBar(
        //         content: Text('No barcode found!'),
        //         backgroundColor: Colors.red,
        //       );

        // ScaffoldMessenger.of(context).showSnackBar(snackbar);
      },
    );
  }
}

class StartStopMobileScannerButton extends StatelessWidget {
  const StartStopMobileScannerButton({required this.controller, super.key});

  final MobileScannerController controller;

  @override
    return ValueListenableBuilder(
      valueListenable: controller,
      builder: (context, state, child) {
        if (!state.isInitialized || !state.isRunning) {
          return IconButton(
            color: Colors.white,
            icon: const Icon(Icons.play_arrow),
            iconSize: 32.0,
            onPressed: () async {
              await controller.start();
            },
          );
        }

        return IconButton(
          color: Colors.white,
          icon: const Icon(Icons.stop),
          iconSize: 32.0,
          onPressed: () async {
            await controller.stop();
          },
        );
      },
    );
  }
}

class SwitchCameraButton extends StatelessWidget {
  const SwitchCameraButton({required this.controller, super.key});

  final MobileScannerController controller;

  @override
    return ValueListenableBuilder(
      valueListenable: controller,
      builder: (context, state, child) {
        if (!state.isInitialized || !state.isRunning) {
          return const SizedBox.shrink();
        }

        final int? availableCameras = state.availableCameras;

        if (availableCameras != null && availableCameras < 2) {
          return const SizedBox.shrink();
        }

        final Widget icon;

        switch (state.cameraDirection) {
          case CameraFacing.front:
            icon = const Icon(Icons.camera_front);
          case CameraFacing.back:
            icon = const Icon(Icons.camera_rear);
        }

        return IconButton(
          iconSize: 32.0,
          icon: icon,
          onPressed: () async {
            await controller.switchCamera();
          },
        );
      },
    );
  }
}

class ToggleFlashlightButton extends StatelessWidget {
  const ToggleFlashlightButton({required this.controller, super.key});

  final MobileScannerController controller;

  @override
    return ValueListenableBuilder(
      valueListenable: controller,
      builder: (context, state, child) {
        if (!state.isInitialized || !state.isRunning) {
          return const SizedBox.shrink();
        }

        switch (state.torchState) {
          case TorchState.auto:
            return IconButton(
              color: Colors.white,
              iconSize: 32.0,
              icon: const Icon(Icons.flash_auto),
              onPressed: () async {
                await controller.toggleTorch();
              },
            );
          case TorchState.off:
            return IconButton(
              color: Colors.white,
              iconSize: 32.0,
              icon: const Icon(Icons.flash_off),
              onPressed: () async {
                await controller.toggleTorch();
              },
            );
          case TorchState.on:
            return IconButton(
              color: Colors.white,
              iconSize: 32.0,
              icon: const Icon(Icons.flash_on),
              onPressed: () async {
                await controller.toggleTorch();
              },
            );
          case TorchState.unavailable:
            return const Icon(
              Icons.no_flash,
              color: Colors.grey,
            );
        }
      },
    );
  }
}

class ScannedBarcodeLabel extends StatelessWidget {
  const ScannedBarcodeLabel({
    super.key,
    required this.barcodes,
  });

  final Stream<BarcodeCapture> barcodes;

  @override
    return StreamBuilder(
      stream: barcodes,
      builder: (context, snapshot) {
        final scannedBarcodes = snapshot.data?.barcodes ?? [];

        if (scannedBarcodes.isEmpty) {
          return const Text(
            'Scan something!',
            overflow: TextOverflow.fade,
            style: TextStyle(color: Colors.white),
          );
        }

        return Text(
          scannedBarcodes.first.displayValue ?? 'No display value.',
          overflow: TextOverflow.fade,
          style: const TextStyle(color: Colors.white),
        );
      },
    );
  }
}

**Comment by navaronbracke** [comment]
Hmm, what device is this on? Do you see anything in XCode / Logcat / browser logs?

**Comment by simplenotezy** [comment]
It's a physical device. I see the following exception being thrown:

image

**Comment by simplenotezy** [comment]
And it works initially. If I start a new debug session it will work just fine. If I then hot reload/restart the app, and go to the route, it won't work. Which smells like something is off in the underlying APIs.

**Comment by navaronbracke** [comment]
Hmm, it's precisely that error that we should ignore when already starting. Is `autoStart` false in your case?

Also, do you have a stack trace? I can test with the sample above to see where it occurs, though.

**Comment by simplenotezy** [comment]
The error is this: when the widget is exposed, the underlying service is not being stopped. When you restart your application, it is technically still running, so when you go to the route that starts the controller, it won't work.

**Comment by simplenotezy** [comment]
Here's a video showing the issue:


https://github.com/juliansteenbakker/mobile_scanner/assets/3786627/c4699717-6ae9-4d9d-82b2-a9350eba182b

**Comment by navaronbracke** [comment]
>  when the widget is disposed*, the underlying service is not being stopped
Could you clarify that? We _should_ be disposing the underlying camera.

https://github.com/navaronbracke/mobile_scanner/blob/fix_already_started_bug/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt#L398

Perhaps there is a case where this isn't true?

**Comment by navaronbracke** [comment]
So this exception only happens when you press `r` (not `R`) in the terminal? In that case, I think it has to do with the plugin not handling hot restart properly. We have an open bug for that as well, although I would have hoped that this fix would have fixed that issue as well.

I'll have to specifically address the use case for hot reload then, perhaps by just short circuiting a return value instead of an error, if we do not have hot reload callbacks.

Another idea would be to see if stopping the controller in `reassemble()` works around this issue?

File: analysis_options.yaml [hunk 1]
```diff
@@ -4,4 +4,5 @@ linter:
   rules:
     - combinators_ordering
     - require_trailing_commas
-    - unnecessary_library_directive
\ No newline at end of file
+    - unnecessary_library_directive
+    - prefer_single_quotes
```
navaronbracke, line 8: We had one violation for this rule. I do personally prefer single quotes, so I turned this on.

lib/src/web/mobile_scanner_web.dart
```diff
@@ -38,12 +38,6 @@ class MobileScannerWeb extends MobileScannerPlatform {
   /// The container div element for the camera view.
   late HTMLDivElement _divElement;
 
-  /// The flag that keeps track of whether a permission request is in progress.
-  ///
-  /// On the web, a permission request triggers a dialog, that in turn triggers a lifecycle change.
-  /// While the permission request is in progress, any attempts at (re)starting the camera should be ignored.
-  bool _permissionRequestInProgress = false;
```
navaronbracke, line 45: This flag is no longer needed. We check if the scanner != null and abort any double start attempts instead.

lib/src/web/mobile_scanner_web.dart
```diff
@@ -256,11 +245,13 @@ class MobileScannerWeb extends MobileScannerPlatform {
 
   @override
   Future<MobileScannerViewAttributes> start(StartOptions startOptions) async {
-    // If the permission request has not yet completed,
-    // the camera view is not ready yet.
-    // Prevent the permission popup from triggering a restart of the scanner.
-    if (_permissionRequestInProgress) {
-      throw PermissionRequestPendingException();
+    if (_barcodeReader != null) {
```
navaronbracke, line 248: Moved up to replace the old check

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerPermissions.kt
```diff
@@ -25,7 +21,7 @@ class MobileScannerPermissions {
     }
 
     interface ResultCallback {
-        fun onResult(errorCode: String?, errorDescription: String?)
+        fun onResult(errorCode: String?)
```
navaronbracke, line 24: The second argument was unused, since we return a bool to the interface

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -28,7 +29,7 @@ class MobileScannerHandler(
 
     private val analyzeImageErrorCallback: AnalyzerErrorCallback = {
         Handler(Looper.getMainLooper()).post {
-            analyzerResult?.error("MobileScanner", it, null)
+            analyzerResult?.error(MobileScannerErrorCodes.BARCODE_ERROR, it, null)
```
navaronbracke, line 32: To be consistent with the results from the scanning stream, I used the same barcode error constant

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -103,21 +104,21 @@ class MobileScannerHandler(
...
-            return
-        }
         when (call.method) {
             "state" -> result.success(permissions.hasCameraPermission(activity))
             "request" -> permissions.requestPermission(
                 activity,
                 addPermissionListener,
                 object: MobileScannerPermissions.ResultCallback {
-                    override fun onResult(errorCode: String?, errorDescription: String?) {
+                    override fun onResult(errorCode: String?) {
```
navaronbracke, line 113: Plumbing through the new error codes

ios/Classes/MobileScanner.swift
```diff
@@ -127,7 +124,6 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     /// Gets called when a new image is added to the buffer
     public func captureOutput(_ output: AVCaptureOutput, didOutput sampleBuffer: CMSampleBuffer, from connection: AVCaptureConnection) {
         guard let imageBuffer = CMSampleBufferGetImageBuffer(sampleBuffer) else {
-            print("Failed to get image buffer from sample buffer.")
```
navaronbracke, line 130: The print isn't helpful to users :)

ios/Classes/MobileScanner.swift
```diff
@@ -25,9 +25,6 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     /// Barcode scanner for results
     var scanner = BarcodeScanner.barcodeScanner()
 
-    /// Return image buffer with the Barcode event
-    var returnImage: Bool = false
```
navaronbracke, line 29: This field was moved to fix a bug

ios/Classes/MobileScannerPlugin.swift
```diff
@@ -37,23 +41,47 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin {
...
-            if barcodes != nil {
-                let barcodesMap: [Any?] = barcodes!.compactMap { barcode in
-                    if (MobileScannerPlugin.scanWindow != nil) {
-                        if (MobileScannerPlugin.isBarcodeInScanWindow(barcode: barcode, imageSize: image.size)) {
-                            return barcode.data
-                        } else {
-                            return nil
-                        }
-                    } else {
+            if (error != nil) {
```
navaronbracke, line 44: Small refactor to use early returns for readability + plumbing of the new error codes

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -90,55 +90,45 @@ class MobileScanner(
...
                 for (barcode in barcodes) {
-                    if (scanWindow != null) {
-                        val match = isBarcodeInScanWindow(scanWindow!!, barcode, imageProxy)
-                        if (!match) {
-                            continue
-                        } else {
-                            barcodeMap.add(barcode.data)
-                        }
-                    } else {
+                    if (scanWindow == null) {
```
navaronbracke, line 93: Refactoring with early returns

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -133,9 +148,13 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
 
   @override
   Stream<BarcodeCapture?> get barcodesStream {
-    return eventsStream
-        .where((event) => event['name'] == 'barcode')
-        .map((event) => _parseBarcode(event));
+    // Handle both incoming barcode events and barcode error events.
```
navaronbracke, line 151: This change forwards barcode error events, which were previously dropped

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -154,13 +173,21 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
-    );
+    try {
+      final Map<String, Object?>? result =
+          await methodChannel.invokeMapMethod<String, Object?>(
+        'analyzeImage',
+        path,
+      );
+
+      return _parseBarcode(result);
+    } on PlatformException catch (exception) {
```
navaronbracke, line 184: The normal scanning flow uses events from the stream sink. `analyzeImage` uses a method call result.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -241,17 +290,13 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
-        startResult['size'] as Map<Object?, Object?>?;
-    final double? width = sizeInfo?['width'] as double?;
-    final double? height = sizeInfo?['height'] as double?;
-
     final Size size;
 
-    if (width == null || height == null) {
-      size = Size.zero;
-    } else {
+    if (startResult['size']
```
navaronbracke, line 295: if-case to extract the size

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -198,14 +224,37 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
     try {
       startResult = await methodChannel.invokeMapMethod<String, Object?>(
         'start',
         startOptions.toMap(),
       );
     } on PlatformException catch (error) {
+      // Map the error code to a MobileScannerErrorCode.
+      // The error code strings should be kept in sync with their native counterparts.
+      // Any error code that is not mapped will be treated as a generic error.
+      switch (error.code) {
```
navaronbracke, line 238: This improved error code mapping helps with detecting what kind of error happened during the start call.

lib/src/mobile_scanner_controller.dart
```diff
@@ -246,13 +263,6 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
       );
     }
 
-    // Permission was denied, do nothing.
-    // When the controller is stopped,
-    // the error is reset so the permission can be requested again if possible.
-    if (value.error?.errorCode == MobileScannerErrorCode.permissionDenied) {
```
navaronbracke, line 252: This was part of the workaround, which is no longer needed.

lib/src/mobile_scanner_controller.dart
```diff
@@ -306,8 +323,6 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
           zoomScale: 1.0,
         );
       }
-    } on PermissionRequestPendingException catch (_) {
```
navaronbracke, line 309: This was part of the workaround, which is no longer needed.

lib/src/mobile_scanner_controller.dart
```diff
@@ -121,14 +124,25 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+    _barcodesSubscription =
+        MobileScannerPlatform.instance.barcodesStream.listen(
+      (BarcodeCapture? barcode) {
+        if (_barcodesController.isClosed || barcode == null) {
+          return;
+        }
+
+        _barcodesController.add(barcode);
+      },
+      onError: (Object error) {
```
navaronbracke, line 136: Now we forward the barcode error events to end-users.

macos/Classes/MobileScannerPlugin.swift
```diff
@@ -106,7 +103,6 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
             return
         }
         guard let imageBuffer = CMSampleBufferGetImageBuffer(sampleBuffer) else {
-            print("Failed to get image buffer from sample buffer.")
```
navaronbracke, line 109: This print is not helpful for end users

macos/Classes/MobileScannerPlugin.swift
```diff
@@ -148,27 +144,28 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
-                                    //                                        let image = CIImage(image: source)
-                                    //                                        image?.cropping(to: barcode.boundingBox)
-                                    //                                        self.qrCodeDescriptor(qrCode: barcode, qrCodeImage: image!)
-                                    //                                    }
                                 }
                             }
                         } else {
                             DispatchQueue.main.async {
-                                self?.sink?(FlutterError(code: "MobileScanner", message: error?.localizedDescription, details: nil))
+                                self?.sink?([
```
navaronbracke, line 151: Plumbing the new error codes + removing commented out code. I have an open point for implementing analyze image on MacOS

navaronbracke:
@simplenotezy Unfortunately this PR reintroduces a bug with permissions. TLDR: I do have a fix for that problem.

In our current guidance for lifecycle changes, we show in the examples that you can start the controller when the app resumes and the controller is initialized. However, there is one problem with that statement: On Android, this would cause a loop of permission requests.

The reason for that is that each permission dialog triggers a lifecycle change. Combining that with our guidance to restart the controller, issues another start request. This is fine if you already have permission. However, if you don't then a new permission popup will trigger, restarting the entire cycle (hence the infinite loop bug).

Due to the fact that iOS, MacOS and web only allow the permission dialog once, they sidestep this problem, so they are not affected.

My solution is simple: In the lifecycle guidance, also check if the scanner state has the permission denied error code and abort the start request. I.e.

```dart
  @override
    if (!controller.value.isInitialized ||
        controller.value.error?.errorCode == MobileScannerErrorCode.permissionDenied) {
      return;
    }

    switch (state) {
      case AppLifecycleState.detached:
      case AppLifecycleState.hidden:
      case AppLifecycleState.paused:
        return;
      case AppLifecycleState.resumed:
        _subscription = controller.barcodes.listen(_handleBarcode);

        unawaited(controller.start());
      case AppLifecycleState.inactive:
        unawaited(_subscription?.cancel());
        _subscription = null;
        unawaited(controller.stop());
    }
  }
```

The second part of the `||` is the new part. However, to ease the transition for our users, I'd like to move this extra check in `isInitialized`, but that feels weird, since you now have a divergent meaning of `isInitialized` and the `error` property.

Do you have any ideas? Perhaps a new getter that combines both checks? But I don't know what to name it.

navaronbracke:
@tanvir93 You should be able to use this patch to see what is going on on Android 14. I added the reporting of errors from MLKit in this PR. You'll have to handle the errors from the `barcodes` stream. Those include the errors from the MLKit error callbacks.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -197,14 +224,37 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
+        // In case the scanner was already started, report the right error code.
+        // If the scanner is already starting,
+        // this error code is a signal to the controller to just ignore the attempt.
+        case 'MOBILE_SCANNER_ALREADY_STARTED_ERROR':
```
navaronbracke, line 242: Can we avoid the magic strings here?

fulstadev:
> ```dart
> controller.value.error?.errorCode == MobileScannerErrorCode.permissionDenied
> ```

@navaronbracke you may want to change the [example](https://github.com/juliansteenbakker/mobile_scanner/blob/master/example/lib/barcode_scanner_controller.dart)'s lifecycle handler init check, as there is no `controller.value.hasCameraPermission` property, but I guess that's exactly what you were aiming to do in that example?

navaronbracke:
> > ```dart
> > controller.value.error?.errorCode == MobileScannerErrorCode.permissionDenied
> > ```
> 
> @navaronbracke you may want to change the [example](https://github.com/juliansteenbakker/mobile_scanner/blob/master/example/lib/barcode_scanner_controller.dart)'s lifecycle handler init check, as there is no `controller.value.hasCameraPermission` property, but I guess that's exactly what you were aiming to do in that example?

That is a new getter that I added in https://github.com/juliansteenbakker/mobile_scanner/commit/1ab7f5a753f9e27c021e780bfa3ac76b8d3cc7b0, which does exactly this check.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/BarcodeHandler.kt
```diff
@@ -24,6 +24,12 @@ class BarcodeHandler(binaryMessenger: BinaryMessenger) : EventChannel.StreamHand
         }
     }
 
+    fun publishError(errorCode: String, errorMessage: String, errorDetails: Any?) {
```
navaronbracke: Like `publishEvent` above, but for `result.error()`

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -65,10 +66,7 @@ class MobileScannerHandler(
     }
 
     private val errorCallback: MobileScannerErrorCallback = {error: String ->
-        barcodeHandler.publishEvent(mapOf(
-            "name" to "error",
-            "data" to error,
-        ))
+        barcodeHandler.publishError(MobileScannerErrorCodes.BARCODE_ERROR, error, null)
```
navaronbracke, line 69: This is an actual error event now.

ios/Classes/MobileScannerPlugin.swift
```diff
@@ -42,7 +42,10 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin {
     init(barcodeHandler: BarcodeHandler, registry: FlutterTextureRegistry) {
         self.mobileScanner = MobileScanner(registry: registry, mobileScannerCallback: { barcodes, error, image in
             if error != nil {
-                barcodeHandler.publishEvent(["name": "error", "data": error!.localizedDescription])
+                barcodeHandler.publishError(
```
navaronbracke, line 45: This is an actual error event now

lib/src/web/zxing/zxing_exception.dart
```diff
@@ -0,0 +1,10 @@
+import 'dart:js_interop';
+
+/// The JS static interop class for the Result class in the ZXing library.
+///
+/// See also: https://github.com/zxing-js/library/blob/master/src/core/Exception.ts
+@JS('ZXing.Exception')
```
navaronbracke, line 6: ZXing had an error type, so I prefer to use it, so we can forward the message.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -145,21 +164,30 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
+              : [
+                  for (final BarcodeFormat format in formats)
+                    if (format != BarcodeFormat.unknown) format.rawValue,
+                ],
+        },
+      );
 
-    return _parseBarcode(result);
+      return _parseBarcode(result);
+    } on PlatformException catch (error) {
```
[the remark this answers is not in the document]
navaronbracke: As this handles `PlatformException` directly in the catch clause filter, this can stay. `_handleBarcodeError()` does an extra type check that would be redundant here.

navaronbracke:
I have started testing this patch and it seems to fix the underlying issue.

I still need to test it on iOS and I'm having a bit of trouble testing MacOS hot reload/hot restart due to a tool bug,
but once those have been tested I will land this patch.

example/.gitignore
```diff
@@ -5,9 +5,11 @@
 *.swp
 .DS_Store
 .atom/
+.build/
 .buildlog/
 .history
 .svn/
+.swiftpm/
```
navaronbracke, line 12: This was added by a project migrator when running on the master branch of Flutter. This is probably needed for SwiftPM stuff.

example/web/index.html
```diff
@@ -21,7 +21,7 @@
   <meta name="description" content="Demonstrates how to use the mobile_scanner plugin.">
 
   <!-- iOS meta tags & icons -->
-  <meta name="apple-mobile-web-app-capable" content="yes">
+  <meta name="mobile-web-app-capable" content="yes">
```
navaronbracke, line 24: The old tag is deprecated

lib/src/web/zxing/zxing_barcode_reader.dart
```diff
@@ -98,16 +106,24 @@ final class ZXingBarcodeReader extends BarcodeReader {
...
             return;
           }
 
-          controller.add(
-            BarcodeCapture(
-              barcodes: [result.toBarcode],
-            ),
-          );
+          // Skip the event if no code was detected.
+          if (error != null && error.message != kNoCodeDetectedErrorMessage) {
```
navaronbracke, line 115: When introducing the forwarding of the errors on web, this was apparently breaking our usage. Added a check to fix it.

example/macos/Runner/DebugProfile.entitlements
```diff
@@ -10,5 +10,7 @@
 	<true/>
 	<key>com.apple.security.network.server</key>
 	<true/>
+  	<key>com.apple.security.files.user-selected.read-only</key>
```
navaronbracke, line 13: Not having this broke the example app on MacOS for analyze image.

This was documented in `image_picker`

example/macos/Runner/AppDelegate.swift
```diff
@@ -6,4 +6,8 @@ class AppDelegate: FlutterAppDelegate {
   override func applicationShouldTerminateAfterLastWindowClosed(_ sender: NSApplication) -> Bool {
     return true
   }
+
+  override func applicationSupportsSecureRestorableState(_ app: NSApplication) -> Bool {
```
navaronbracke, line 10: This fixes a warning for MacOS

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -488,14 +488,14 @@
 				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = "$(FLUTTER_BUILD_NUMBER)";
-				DEVELOPMENT_TEAM = 75Y2P2WSQQ;
+				DEVELOPMENT_TEAM = "";
```
navaronbracke, line 491: When trying to test the iOS example app, XCode kept complaining about the development team & bundle identifier.

Presumably the old development team was Julian's personal team or something.
Likewise, the bundle identifier was already registered (which really should not be the case for an exampel app, that isn't intended for publishing)

I updated the bundle identifier to something that is not registered, by adding the `-example` string at the end.
I changed the development team to `None` in XCode, which is better than a team that isn't yours in my opinion.

navaronbracke:
This fix was now verified for iOS, MacOS, Android and the web. With this fix, during hot restart, the app restarts fully (as expected) and with hot reload the camera continues to work as if the hot reload never happened (since hot reload is only for Dart code on the client, not the plugin)

This fix will be available for the next release, which I plan to release later today.

PR #1095, opened 2024-06-12
info.plist is updated with NSCameraUsageDescription and NSPhotoLibraryUsageDescription, Readme has modified accordingly

gopalkumr:
info.plist has updated with
`
 <key>NSCameraUsageDescription</key>
    <string>Camera permission is required for barcode scanning.</string>
    <key>NSPhotoLibraryUsageDescription</key>
    <string>Photo library permission is required for barcode scanning.</string>
`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,19 +60,10 @@ dev.steenbakker.mobile_scanner.useUnbundled=true
 ```
 
 ### iOS
-**Add the following keys to your Info.plist file, located in <project root>/ios/Runner/Info.plist:**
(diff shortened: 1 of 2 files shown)


navaronbracke:
This PR seems redundant? The example app merely includes the permission labels in the Info.plist to make the example run out of the box. It is up to the user to provide these labels for their own app (and translate them accordingly as well, using InfoPlist.strings)

The previous README documentation stated this properly.
Also, this PR replaces the old Info.plist instructions with a `flutter run` command, which is something that the flutter tool teaches you to use.

Issue #1103, opened 2024-06-19
[Bug] iOS background queue has the wrong QoS priority


chenjiangmin:
I attempted to introduce this plugin by merely including it in the pubspec.yaml.

I did not use any of its functionalities.

However, the “flutter_baidu_mapapi_map” that I use in my program can no longer work. When opening the map, the thread gets stuck.

Thread running at QOS_CLASS_USER_INTERACTIVE waiting on a lower QoS thread running at QOS_CLASS_DEFAULT.

navaronbracke:
Could you provide more information? Are you running on MacOS or iOS? What version of mobile_scanner are you using?

chenjiangmin:
Yes, I am using the iOS system to debug the system version, which is iOS 17


```
[✓] Flutter (Channel stable, 3.19.6, on macOS 14.5 23F79 darwin-arm64, locale zh-Hans-CN)
    • Flutter version 3.19.6 on channel stable at /Users/chenjiangmin/fvm/versions/3.19.6
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision 54e66469a9 (10 周前), 2024-04-17 13:08:03 -0700
    • Engine revision c4cd48e186
    • Dart version 3.3.4
    • DevTools version 2.31.1

[✓] Android toolchain - develop for Android devices (Android SDK version 33.0.0)
    • Android SDK at /Users/chenjiangmin/Library/Android/sdk
    • Platform android-34, build-tools 33.0.0
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)
    • All Android licenses accepted.

[✓] Xcode - develop for iOS and macOS (Xcode 15.4)
    • Xcode at /Applications/Xcode.app/Contents/Developer
    • Build 15F31d
    • CocoaPods version 1.15.2

[✓] Android Studio (version 2023.2)
    • Android Studio at /Applications/Android Studio.app/Contents
    • Flutter plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/9212-flutter
    • Dart plugin can be installed from:
      🔨 https://plugins.jetbrains.com/plugin/6351-dart
    • Java version OpenJDK Runtime Environment (build 17.0.9+0-17.0.9b1087.7-11185874)

[...]

PR #1110, opened 2024-07-01
fix: Adjust iOS thread QoS value

navaronbracke:
Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/1103


navaronbracke:
It turns out this might be a Flutter issue with Platform Views, see https://github.com/flutter/flutter/issues/142841

Issue #988, opened 2024-03-18
StrictMode DiskReadViolation blocking main thread


radivojeostojic:
During testing of our Android application in StrictMode, we noticed increased number of DiskReadViolation of our MainThread related to mobile_scanner.

Sample log:

```
03-15 14:07:39.411 29151 29161 D StrictMode: StrictMode policy violation: android.os.strictmode.LeakedClosableViolation: A resource was acquired at attached stack trace but never released. See java.io.Closeable for information on avoiding resource leaks.
03-15 14:07:39.411 29151 29161 D StrictMode: 	at android.os.StrictMode$AndroidCloseGuardReporter.report(StrictMode.java:1786)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at dalvik.system.CloseGuard.warnIfOpen(CloseGuard.java:264)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at android.view.Surface.finalize(Surface.java:216)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at java.lang.Daemons$FinalizerDaemon.doFinalize(Daemons.java:250)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at java.lang.Daemons$FinalizerDaemon.runInternal(Daemons.java:237)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at java.lang.Daemons$Daemon.run(Daemons.java:103)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at java.lang.Thread.run(Thread.java:764)
03-15 14:07:39.411 29151 29161 D StrictMode: Caused by: java.lang.Throwable: Explicit termination method 'release' not called
03-15 14:07:39.411 29151 29161 D StrictMode: 	at dalvik.system.CloseGuard.open(CloseGuard.java:221)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at android.view.Surface.setNativeObjectLocked(Surface.java:618)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at android.view.Surface.<init>(Surface.java:201)
03-15 14:07:39.411 29151 29161 D StrictMode: 	at dev.steenbakker.mobile_scanner.MobileScanner.start$lambda$12$lambda$6(MobileScanner.kt:265)
[...]

Issue #1134, opened 2024-07-30
Cannot round NaN excetion crushing the app


mkaladelfos:
FATAL EXCEPTION: main
Cannot round NaN

FATAL EXCEPTION: main
E/AndroidRuntime( 8980): Process: com.example.app_name, PID: 8980
E/AndroidRuntime( 8980): java.lang.IllegalArgumentException: Cannot round NaN value.
E/AndroidRuntime( 8980): 	at kotlin.math.MathKt__MathJVMKt.roundToInt(MathJVM.kt:1165)
E/AndroidRuntime( 8980): 	at dev.steenbakker.mobile_scanner.MobileScanner.isBarcodeInScanWindow(MobileScanner.kt:177)
E/AndroidRuntime( 8980): 	at dev.steenbakker.mobile_scanner.MobileScanner.captureOutput$lambda$6$lambda$1(MobileScanner.kt:94)
E/AndroidRuntime( 8980): 	at dev.steenbakker.mobile_scanner.MobileScanner.$r8$lambda$WHeCA6JglANxIjeKXlTGAl6H0VI(Unknown Source:0)
E/AndroidRuntime( 8980): 	at dev.steenbakker.mobile_scanner.MobileScanner$$ExternalSyntheticLambda3.invoke(Unknown Source:8)
E/AndroidRuntime( 8980): 	at dev.steenbakker.mobile_scanner.MobileScanner.captureOutput$lambda$6$lambda$2(MobileScanner.kt:80)
E/AndroidRuntime( 8980): 	at dev.steenbakker.mobile_scanner.MobileScanner.$r8$lambda$FNQbW8g2ym_6Y4EZ1gdRVqQ6VqA(Unknown Source:0)
E/AndroidRuntime( 8980): 	at dev.steenbakker.mobile_scanner.MobileScanner$$ExternalSyntheticLambda14.onSuccess(Unknown Source:2)
E/AndroidRuntime( 8980): 	at com.google.android.gms.tasks.zzm.run(com.google.android.gms:play-services-tasks@@18.0.2:1)
E/AndroidRuntime( 8980): 	at android.os.Handler.handleCallback(Handler.java:942)
E/AndroidRuntime( 8980): 	at android.os.Handler.dispatchMessage(Handler.java:99)
E/AndroidRuntime( 8980): 	at android.os.Looper.loopOnce(Looper.java:226)
[...]

Issue #1069, opened 2024-05-17
Issue with MobileScannerController: Recognizes Unspecified Formats


DamianoPedoniMintFlavour:
**Description**:
I am experiencing an issue with the MobileScannerController where it recognizes barcode formats that are not specified in the formats list.

**Code Snippet**:

```
formats = [BarcodeFormat.qrCode];

final MobileScannerController controller = MobileScannerController(
  formats: formats
);

final BarcodeCapture? barcodes = await controller.analyzeImage(f.path);
```

**Expected Behavior**:
The scanner should only recognize QR codes as specified in the formats list.

**Actual Behavior**:
The scanner recognizes other barcode formats, such as 1D barcodes (e.g., Code 128), even though only QR codes are specified.

**Steps to Reproduce**:

Set up the MobileScannerController with formats set to [BarcodeFormat.qrCode].
Use the controller to analyze an image containing a 1D barcode.
Observe that the 1D barcode is recognized, despite not being specified in the formats list.
Environment:

mobile_scanner version: 5.1.1
Device: IOS device

navaronbracke:
I think I see the problem here.

The library erroneously sets up the scanner without options as the _default value_. And we only set up the scanner with options when you call `start()`.

PR https://github.com/juliansteenbakker/mobile_scanner/pull/991 partially fixes that issue (although I'll have to get rid of the assert when the scanner is null)

Once that PR is merged, we should take the formats from the controller and use them when creating a short lived scanner instance.

DamianoPedoniMintFlavour:
Thank you for the quick response.

For the time being, I've used this workaround.

```
if (barcodes != null) {
  for (Barcode barcode in barcodes.barcodes) {
    // check if the format is usable
    if (!formats.contains(barcode.format)) {
      print("We should not use this");
      continue;
    }
    
    [..]
  }
}
```

PR #1139, opened 2024-08-08
fix: fix barcode scanner leak on Android

navaronbracke:
This PR adds the following changes:
- Fixes a possible leak on Android relating to keeping an orphaned barcode scanner around (the one that is created when the scanner variable is declared, but not yet used)
- Fixed another possible leak where not all camera observers were released
- Reflows some part of the Android code to use early returns
- Updates `analyzeImage()` to take configurable barcode scanner options & use a short lived scanner
- Adds a way to mock the barcode scanner for testing (will be used for the NaN error fix)

For future readers, iOS does not have an equivalent release method for the scanner in the MLKit library.
It's possible that uses an autorelease pool or destructor.

Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/988
Part of https://github.com/juliansteenbakker/mobile_scanner/issues/1069
Part of https://github.com/juliansteenbakker/mobile_scanner/issues/1134

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## NEXT
+* Fixed a leak of the barcode scanner on Android.
+
 ## 5.1.1
 * This release fixes an issue with automatic starts in the examples.
 
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -472,4 +472,11 @@ class MobileScanner(
         camera?.cameraControl?.setZoomRatio(1f)
     }
 
+    /**
+     * Dispose of this scanner instance.
+     */
+    fun dispose() {
+        scanner?.close()
+        scanner = null
+    }
 }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -43,6 +43,7 @@ class MobileScanner(
     private val textureRegistry: TextureRegistry,
     private val mobileScannerCallback: MobileScannerCallback,
     private val mobileScannerErrorCallback: MobileScannerErrorCallback,
+    private val barcodeScannerFactory: (options: BarcodeScannerOptions?) -> BarcodeScanner = ::defaultBarcodeScannerFactory,
 ) {
 
     /// Internal variables
@@ -60,6 +61,15 @@ class MobileScanner(
     private var detectionSpeed: DetectionSpeed = DetectionSpeed.NO_DUPLICATES
     private var detectionTimeout: Long = 250
     private var returnImage = false
+
+    companion object {
+        /**
+         * Create a barcode scanner from the given options.
+         */
+        fun defaultBarcodeScannerFactory(options: BarcodeScannerOptions?) : BarcodeScanner {
+            return if (options == null) BarcodeScanning.getClient() else BarcodeScanning.getClient(options)
+        }
+    }
 
     /**
      * callback for the camera. Every frame is passed through this function.
@@ -248,7 +258,7 @@ class MobileScanner(
         }
 
         lastScanned = null
-        scanner = createBarcodeScanner(barcodeScannerOptions)
+        scanner = barcodeScannerFactory(barcodeScannerOptions)
 
         val cameraProviderFuture = ProcessCameraProvider.getInstance(activity)
         val executor = ContextCompat.getMainExecutor(activity)
@@ -451,7 +461,7 @@ class MobileScanner(
         val inputImage = InputImage.fromFilePath(activity, image)
 
         // Use a short lived scanner instance, which is closed when the analysis is done.
-        val barcodeScanner: BarcodeScanner = createBarcodeScanner(scannerOptions)
+        val barcodeScanner: BarcodeScanner = barcodeScannerFactory(scannerOptions)
 
         barcodeScanner.process(inputImage).addOnSuccessListener { barcodes ->
             val barcodeMap = barcodes.map { barcode -> barcode.data }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -20,12 +20,12 @@ import androidx.camera.core.ImageAnalysis
 import androidx.camera.core.ImageProxy
 import androidx.camera.core.Preview
 import androidx.camera.core.TorchState
-import androidx.camera.core.resolutionselector.AspectRatioStrategy
 import androidx.camera.core.resolutionselector.ResolutionSelector
 import androidx.camera.core.resolutionselector.ResolutionStrategy
 import androidx.camera.lifecycle.ProcessCameraProvider
 import androidx.core.content.ContextCompat
 import androidx.lifecycle.LifecycleOwner
+import com.google.mlkit.vision.barcode.BarcodeScanner
 import com.google.mlkit.vision.barcode.BarcodeScannerOptions
 import com.google.mlkit.vision.barcode.BarcodeScanning
 import com.google.mlkit.vision.barcode.common.Barcode
@@ -36,7 +36,6 @@ import dev.steenbakker.mobile_scanner.utils.YuvToRgbConverter
 import io.flutter.view.TextureRegistry
 import java.io.ByteArrayOutputStream
 import kotlin.math.roundToInt
-
 
 class MobileScanner(
     private val activity: Activity,
@@ -50,7 +49,7 @@ class MobileScanner(
     private var camera: Camera? = null
     private var preview: Preview? = null
     private var textureEntry: TextureRegistry.SurfaceTextureEntry? = null
-    private var scanner = BarcodeScanning.getClient()
+    private var scanner: BarcodeScanner? = null
     private var lastScanned: List<String?>? = null
     private var scannerTimeout = false
     private var displayListener: DisplayManager.DisplayListener? = null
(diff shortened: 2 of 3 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -13,19 +13,20 @@ import android.os.Looper
...
 import android.view.WindowManager
+import androidx.annotation.VisibleForTesting
 import androidx.camera.core.Camera
 import androidx.camera.core.CameraSelector
 import androidx.camera.core.ExperimentalGetImage
 import androidx.camera.core.ImageAnalysis
 import androidx.camera.core.ImageProxy
 import androidx.camera.core.Preview
 import androidx.camera.core.TorchState
-import androidx.camera.core.resolutionselector.AspectRatioStrategy
```
navaronbracke, line 23: Unused import

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -37,20 +38,20 @@ import io.flutter.view.TextureRegistry
...
 import kotlin.math.roundToInt
 
-
 class MobileScanner(
     private val activity: Activity,
     private val textureRegistry: TextureRegistry,
     private val mobileScannerCallback: MobileScannerCallback,
-    private val mobileScannerErrorCallback: MobileScannerErrorCallback
+    private val mobileScannerErrorCallback: MobileScannerErrorCallback,
+    private val barcodeScannerFactory: (options: BarcodeScannerOptions?) -> BarcodeScanner = ::defaultBarcodeScannerFactory,
```
navaronbracke, line 46: This new callback is mainly for tests. I'm adding a new test soon that needs this, in a follow up PR for the rounding bug https://github.com/juliansteenbakker/mobile_scanner/issues/1134

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -37,20 +38,20 @@ import io.flutter.view.TextureRegistry
...
 ) {
 
     /// Internal variables
     private var cameraProvider: ProcessCameraProvider? = null
     private var camera: Camera? = null
     private var preview: Preview? = null
     private var textureEntry: TextureRegistry.SurfaceTextureEntry? = null
-    private var scanner = BarcodeScanning.getClient()
+    private var scanner: BarcodeScanner? = null
```
navaronbracke, line 54: This client was never disposed

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -61,6 +62,15 @@ class MobileScanner(
     private var detectionTimeout: Long = 250
     private var returnImage = false
 
+    companion object {
+        /**
+         * Create a barcode scanner from the given options.
+         */
+        fun defaultBarcodeScannerFactory(options: BarcodeScannerOptions?) : BarcodeScanner {
```
navaronbracke, line 69: Default for the barcode scanner creator function

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -76,76 +86,75 @@ class MobileScanner(
             scannerTimeout = true
         }
 
-        scanner.process(inputImage)
-            .addOnSuccessListener { barcodes ->
+        scanner?.let {
```
navaronbracke, line 89: Scanner is nullable now, so used a let block

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -76,76 +86,75 @@ class MobileScanner(
...
+            it.process(inputImage).addOnSuccessListener { barcodes ->
                 if (detectionSpeed == DetectionSpeed.NO_DUPLICATES) {
-                    val newScannedBarcodes = barcodes.mapNotNull { barcode -> barcode.rawValue }.sorted()
+                    val newScannedBarcodes = barcodes.mapNotNull {
```
navaronbracke, line 92: Formatting of mapNotNull to use braces

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -76,76 +86,75 @@ class MobileScanner(
...
+                        barcode -> barcode.rawValue
+                    }.sorted()
+
                     if (newScannedBarcodes == lastScanned) {
                         // New scanned is duplicate, returning
                         return@addOnSuccessListener
                     }
-                    if (newScannedBarcodes.isNotEmpty()) lastScanned = newScannedBarcodes
+                    if (newScannedBarcodes.isNotEmpty()) {
```
navaronbracke, line 100: Same formatting change here

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -76,76 +86,75 @@ class MobileScanner(
...
                 for (barcode in barcodes) {
-                    if (scanWindow != null) {
-                        val match = isBarcodeInScanWindow(scanWindow!!, barcode, imageProxy)
-                        if (!match) {
-                            continue
-                        } else {
-                            barcodeMap.add(barcode.data)
-                        }
-                    } else {
+                    if (scanWindow == null) {
```
navaronbracke, line 108: Inverting the check here made more sense

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -76,76 +86,75 @@ class MobileScanner(
...
-                }
-
-
-                if (barcodeMap.isNotEmpty()) {
-                    if (returnImage) {
 
-                        val bitmap = Bitmap.createBitmap(mediaImage.width, mediaImage.height, Bitmap.Config.ARGB_8888)
-
-                        val imageFormat = YuvToRgbConverter(activity.applicationContext)
+                    if (isBarcodeInScanWindow(scanWindow!!, barcode, imageProxy)) {
```
navaronbracke, line 113: Also invert the check here

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -76,76 +86,75 @@ class MobileScanner(
...
+                        barcodeMap.add(barcode.data)
+                    }
+                }
 
-                        imageFormat.yuvToRgb(mediaImage, bitmap)
+                if (barcodeMap.isEmpty()) {
```
navaronbracke, line 118: Early return 2x

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -408,14 +412,27 @@ class MobileScanner(
         }
 
         val owner = activity as LifecycleOwner
-        camera?.cameraInfo?.torchState?.removeObservers(owner)
+        // Release the camera observers first.
+        camera?.cameraInfo?.let {
```
navaronbracke, line 416: This was a bug, we didn't remove all observers. `removeObservers` does nothing if the owner is not a listener, so calling this is safe for all three LiveData instances.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -408,14 +412,27 @@ class MobileScanner(
...
+            it.torchState.removeObservers(owner)
+            it.zoomState.removeObservers(owner)
+            it.cameraState.removeObservers(owner)
+        }
+        // Unbind the camera use cases, the preview is a use case.
+        // The camera will be closed when the last use case is unbound.
         cameraProvider?.unbindAll()
-        textureEntry?.release()
-
+        cameraProvider = null
```
navaronbracke, line 424: The following is just reflowing code so that the relevant things are together.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -408,14 +412,27 @@ class MobileScanner(
...
         camera = null
         preview = null
+
+        // Release the texture for the preview.
+        textureEntry?.release()
         textureEntry = null
-        cameraProvider = null
+
+        // Release the scanner.
+        scanner?.close()
```
navaronbracke, line 433: This is new, we release the scanner when the users stops the camera. No need to keep it around when the preview / camera is gone.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -408,14 +412,27 @@ class MobileScanner(
...
+        scanner = null
+        lastScanned = null
```
navaronbracke, line 435: Setting `lastScanned` to null is to keep things consistent.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -439,22 +456,29 @@ class MobileScanner(
     /**
      * Analyze a single image.
      */
-    fun analyzeImage(image: Uri, onSuccess: AnalyzerSuccessCallback, onError: AnalyzerErrorCallback) {
+    fun analyzeImage(
```
navaronbracke, line 459: Analyze image now takes configurable options and uses a short lived scanner.

Using the short lived scanner is part of the fix for the retain cycle thing and allowing options is part of the fix for analyze image allowing all formats.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -92,6 +92,7 @@ class MobileScannerHandler(
     fun dispose(activityPluginBinding: ActivityPluginBinding) {
         methodChannel?.setMethodCallHandler(null)
         methodChannel = null
+        mobileScanner?.dispose()
```
navaronbracke, line 95: When the user calls `MobileScannerController.dispose()` we should also dispose the underlying camera / scanner. There is no guarantee that `stop()` is called before `dispose()`.

We might get only `dispose()` when the user leaves a scanning page for example.

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## NEXT
+* Fixed a leak of the barcode scanner on Android.
```
navaronbracke, line 2: I'm planning on fixing the NaN error & analyze image format issue in the same release, so I'm using the NEXT tag here for a moment.

PR #1148, opened 2024-08-13
fix: Export StartOptions

gregorscholz:
To create an Instance of the StartOptions, they need to be exported aswell.


navaronbracke:
What is the user case of exporting the StartOptions? This is meant to be a private API.

gregorscholz:
I am trying to use the scanner without using the widget.
While looking through the code i saw that the export is missing. I thought this was unintentional.

gregorscholz:
Okay, thank you for your time.

PR #1157, opened 2024-08-20
chore: Bump version to 5.2.0

navaronbracke:
This PR bumps the version of `mobile_scanner` to 5.2.0

I used this version number, to not break users that have not migrated to Flutter 3.22.0

This version includes the bump for package:web and some bugfixes.

For the next version, I plan to fix the NV21 format bug and finish the Impeller migration on Android.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-## NEXT
+## 5.2.0
 * This release requires Flutter 3.22.0 and Dart 3.4.
 
 * [Android] Fixed a leak of the barcode scanner.
--- a/ios/mobile_scanner.podspec
+++ b/ios/mobile_scanner.podspec
@@ -4,7 +4,7 @@
 #
 Pod::Spec.new do |s|
   s.name             = 'mobile_scanner'
-  s.version          = '5.1.1'
+  s.version          = '5.2.0'
   s.summary          = 'An universal scanner for Flutter based on MLKit.'
   s.description      = <<-DESC
 An universal scanner for Flutter based on MLKit.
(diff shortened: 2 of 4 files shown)


SofienTouati-ProxymIT:
the new provided version is not compatible with auto_route 8.2.0
here is the log
```Resolving dependencies...
Because auto_route ^8.2.0 depends on web ^0.5.1 and mobile_scanner >=5.2.0 depends on web ^1.0.0, auto_route ^8.2.0 is incompatible with mobile_scanner >=5.2.0.
So, because dukhan_flutter depends on both auto_route ^8.2.0 and mobile_scanner ^5.2.0, version solving failed.


You can try one of the following suggestions to make the pubspec resolve:
* Consider downgrading your constraint on auto_route: flutter pub add auto_route:^8.1.4
* Consider downgrading your constraint on mobile_scanner: flutter pub add mobile_scanner:^5.1.1
```

juliansteenbakker:
@navaronbracke This should be a major version release, not a minor one, because flutter pub upgrade will fail for people not on the latest version of flutter. Is it possible to revert this release, and re release it under 6.0.0?

juliansteenbakker:
@navaronbracke I've looked into the changes, but as i can see the actual web code hasn't changed. Because of that, we use `web: ">=0.5.1 <2.0.0"` in order to maintain compatibility with older versions and we won't need a major version release. The flutter version can be determined by the client app itself.

navaronbracke:
@juliansteenbakker I do agree, we should probably be using a version range here. I'll publish version 5.2.1 with the change.

juliansteenbakker:
Thank you for the quick response!

navaronbracke:
Opened https://github.com/juliansteenbakker/mobile_scanner/pull/1158 to fix this.

PR #1158, opened 2024-08-20
chore: Use a version range for package:web

navaronbracke:


navaronbracke:
@juliansteenbakker PTAL

PR #1175, opened 2024-09-09
fix: Handle unknown barcode data enum types & disable media controls

### LINKED ISSUE #1172: Exception when scanning WIFI QR code with encryption type set to SAE (iOS only)
I encountered a problem when scanning a wifi QR code with the encryption type set to "SAE". First of all, I want to point out that "SAE" is not officially defined as an encryption type. To my knowledge it's only "nopass", "WEP", and "WPA".

That aside, when scanning a wifi QR code with "SAE" as encryption type, then the underlying ML Kit returns an otherwise undocumented value of 4 on iOS. As such, when converting the raw value to its enum value, an argument error is raised.

https://github.com/juliansteenbakker/mobile_scanner/blob/65bc14bb3930e27b48d917d3d645d0bef9bb1bb0/lib/src/enums/encryption_type.dart#L17-L30

On Android, however, ML Kit returns a value of 1 (which I think is sufficient in this case).

According to the documentation of ML Kit, there is no official value defined for 4. The same is true for the Android docs.

https://developers.google.com/ml-kit/reference/ios/mlkitbarcodescanning/api/reference/Constants

```objectivec
static const MLKBarcodeWiFiEncryptionType MLKBarcodeWiFiEncryptionTypeUnknown = 0
static const MLKBarcodeWiFiEncryptionType MLKBarcodeWiFiEncryptionTypeOpen = 1
static const MLKBarcodeWiFiEncryptionType MLKBarcodeWiFiEncryptionTypeWPA = 2
static const MLKBarcodeWiFiEncryptionType MLKBarcodeWiFiEncryptionTypeWEP = 3
```

I have experimented a little, and it seems that 4 is only returned for "SAE" and no other letter combinations, e.g. "XXX".

**WEP**

`WIFI:T:WEP;S:SSID;P:PASSWORD;;`

**WPA**

`WIFI:T:WPA;S:SSID;P:PASSWORD;;`

**SAE**

`WIFI:T:SAE;S:SSID;P:PASSWORD;;`

**XXX**

`WIFI:T:XXX;S:SSID;P:PASSWORD;;`

I got the following results on Android and on iOS:

| Raw | Android  | iOS |
| --- | --- | --- |
| WEP | 3 | 3 |
| WPA | 2 | 2 |
| SAE | 1 | 4 |
| XXX | 1 | 1 |

I am not sure how to solve this problem, but currently there is no way to catch the error from the Flutter side other than listing for fatal error. Maybe open an issue for the ML Kit development team? Otherwise, I wouldn't mind having 4 converted to none/unknown.

Lastly, I wonder if there are more undocumented values beyond 4...

### LINKED ISSUE #1174: mobile_scanner overlay on iOS webview
Description:
I'm experiencing an issue where the mobile_scanner widget (version 3.4.1) displays an unexpected overlay when used within an iOS webview integrated into a native iOS app.
This overlay contains controls such as pause/play, mute, a cross button, and a "Live" label at the bottom.
The website works correctly on Safari, Chrome, and Android webviews, suggesting the issue is specific to iOS webviews.

Steps to Reproduce:
Integrate a website containing a mobile_scanner widget into an iOS native app using a webview.
Run the iOS app and attempt to use the mobile_scanner widget.

Expected Behavior:
The mobile_scanner widget should function normally within the webview without any overlay.

Actual Behavior:
An overlay with multimedia controls appears on top of the mobile_scanner widget, interfering with its functionality.
Additional Information:
The issue persists even after upgrading to the latest versions of mobile_scanner.
The problem only occurs within iOS webviews, not on Safari, Chrome, or Android webviews.

The below shown screen is what  I am experiencing
Image_20240909_123616_972

The below shown screen is the expected 
Image_20240909_123617_032

navaronbracke:
Fixes #1172 
Fixes #1174

### DIFF
--- a/lib/src/enums/encryption_type.dart
+++ b/lib/src/enums/encryption_type.dart
@@ -14,7 +14,9 @@ enum EncryptionType {
 
   const EncryptionType(this.rawValue);
 
-  @Deprecated('EncryptionType.none is deprecated. Use EncryptionType.unknown instead.')
+  @Deprecated(
+    'EncryptionType.none is deprecated. Use EncryptionType.unknown instead.',
+  )
   static const EncryptionType none = EncryptionType.unknown;
 
   factory EncryptionType.fromRawValue(int value) {
--- a/lib/src/enums/encryption_type.dart
+++ b/lib/src/enums/encryption_type.dart
@@ -1,7 +1,7 @@
 /// Wifi encryption type constants.
 enum EncryptionType {
   /// Unknown encryption type.
-  none(0),
+  unknown(0),
 
   /// Not encrypted.
   open(1),
@@ -14,10 +14,13 @@ enum EncryptionType {
 
   const EncryptionType(this.rawValue);
 
+  @Deprecated('EncryptionType.none is deprecated. Use EncryptionType.unknown instead.')
+  static const EncryptionType none = EncryptionType.unknown;
+
   factory EncryptionType.fromRawValue(int value) {
     switch (value) {
       case 0:
-        return EncryptionType.none;
+        return EncryptionType.unknown;
       case 1:
         return EncryptionType.open;
       case 2:
@@ -25,7 +28,7 @@ enum EncryptionType {
       case 3:
         return EncryptionType.wep;
       default:
-        throw ArgumentError.value(value, 'value', 'Invalid raw value.');
+        return EncryptionType.unknown;
     }
   }
 
--- a/lib/src/web/mobile_scanner_web.dart
+++ b/lib/src/web/mobile_scanner_web.dart
@@ -88,6 +88,18 @@ class MobileScannerWeb extends MobileScannerPlatform {
       ..transformOrigin = 'center'
       ..pointerEvents = 'none';
 
+    // Do not show the media controls, as this is a preview element.
+    // Also prevent play/pause events from changing the media controls.
+    videoElement.controls = false;
+
+    videoElement.onplay = (JSAny _) {
+      videoElement.controls = false;
+    }.toJS;
+
+    videoElement.onpause = (JSAny _) {
+      videoElement.controls = false;
+    }.toJS;
+
     // Attach the video element to its parent container
     // and setup the PlatformView factory for this `textureId`.
     _divElement = HTMLDivElement()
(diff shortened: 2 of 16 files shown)


lib/src/enums/encryption_type.dart
```diff
@@ -1,7 +1,7 @@
 /// Wifi encryption type constants.
 enum EncryptionType {
   /// Unknown encryption type.
-  none(0),
+  unknown(0),
```
navaronbracke, line 4: As noted in https://github.com/juliansteenbakker/mobile_scanner/issues/1172

Android: never reported 'none/unknown' 
iOS: reported `MLKBarcodeWiFiEncryptionTypeUnknown` as integer 0

So this constant's name is wrong, but the documentation is correct. I opted into deprecating the old name (providing a static const for users to migrate), to match the new, correct behavior.

lib/src/enums/encryption_type.dart
```diff
@@ -14,18 +14,23 @@ enum EncryptionType {
...
-        return EncryptionType.none;
+        return EncryptionType.unknown;
       case 1:
         return EncryptionType.open;
       case 2:
         return EncryptionType.wpa;
       case 3:
         return EncryptionType.wep;
       default:
-        throw ArgumentError.value(value, 'value', 'Invalid raw value.');
```
navaronbracke, line 28: This fixes the strange integer 4 that was reported in https://github.com/juliansteenbakker/mobile_scanner/issues/1172

lib/src/enums/encryption_type.dart
```diff
@@ -14,18 +14,23 @@ enum EncryptionType {
 
   const EncryptionType(this.rawValue);
 
+  @Deprecated(
+    'EncryptionType.none is deprecated. Use EncryptionType.unknown instead.',
+  )
+  static const EncryptionType none = EncryptionType.unknown;
```
navaronbracke, line 20: For backwards compatibility, Included a static const alias that points to the old name.

We can remove this constant in version 6.0.0

lib/src/web/mobile_scanner_web.dart
```diff
@@ -88,6 +88,18 @@ class MobileScannerWeb extends MobileScannerPlatform {
       ..transformOrigin = 'center'
       ..pointerEvents = 'none';
 
+    // Do not show the media controls, as this is a preview element.
+    // Also prevent play/pause events from changing the media controls.
+    videoElement.controls = false;
```
navaronbracke, line 93: Apparently we did not unset the controls for the video element. iOS WKWebviews might add the controls to the video, unless prevented.

PR #1180, opened 2024-09-18
fix: Clean up error codes

navaronbracke:
Part of #1086 but split off to make the other diff smaller

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -5,3 +5,4 @@ linter:
     - combinators_ordering
     - require_trailing_commas
     - unnecessary_library_directive
+    - prefer_single_quotes
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -120,12 +120,7 @@ class MobileScanner(
                 }
 
                 if (!returnImage) {
-                    mobileScannerCallback(
-                        barcodeMap,
-                        null,
-                        null,
-                        null
-                    )
+                    mobileScannerCallback(barcodeMap, null, null, null)
                     return@addOnSuccessListener
                 }
 
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/MobileScannerErrorCodes.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/MobileScannerErrorCodes.kt
@@ -1,0 +1,25 @@
+package dev.steenbakker.mobile_scanner.objects
+
+class MobileScannerErrorCodes {
+    companion object {
+        const val ALREADY_STARTED_ERROR = "MOBILE_SCANNER_ALREADY_STARTED_ERROR"
+        const val ALREADY_STARTED_ERROR_MESSAGE = "The scanner was already started."
+        // The error code 'BARCODE_ERROR' does not have an error message,
+        // because it uses the error message from the underlying error.
+        const val BARCODE_ERROR = "MOBILE_SCANNER_BARCODE_ERROR"
+        // The error code 'CAMERA_ACCESS_DENIED' does not have an error message,
+        // because it is used for a boolean result.
+        const val CAMERA_ACCESS_DENIED = "MOBILE_SCANNER_CAMERA_PERMISSION_DENIED"
+        const val CAMERA_ERROR = "MOBILE_SCANNER_CAMERA_ERROR"
+        const val CAMERA_ERROR_MESSAGE = "An error occurred when opening the camera."
+        const val CAMERA_PERMISSIONS_REQUEST_ONGOING = "MOBILE_SCANNER_CAMERA_PERMISSION_REQUEST_PENDING"
+        const val CAMERA_PERMISSIONS_REQUEST_ONGOING_MESSAGE = "Another request is ongoing and multiple requests cannot be handled at once."
+        const val GENERIC_ERROR = "MOBILE_SCANNER_GENERIC_ERROR"
+        const val GENERIC_ERROR_MESSAGE = "An unknown error occurred."
+        const val INVALID_ZOOM_SCALE_ERROR_MESSAGE = "The zoom scale should be between 0 and 1 (both inclusive)"
+        const val NO_CAMERA_ERROR = "MOBILE_SCANNER_NO_CAMERA_ERROR"
+        const val NO_CAMERA_ERROR_MESSAGE = "No cameras available."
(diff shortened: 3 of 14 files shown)


analysis_options.yaml
```diff
@@ -4,4 +4,5 @@ linter:
   rules:
     - combinators_ordering
     - require_trailing_commas
-    - unnecessary_library_directive
\ No newline at end of file
+    - unnecessary_library_directive
+    - prefer_single_quotes
```
navaronbracke, line 8: I do prefer single quotes, and we had one violation. So I turned on the lint.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -120,12 +120,7 @@ class MobileScanner(
                 }
 
                 if (!returnImage) {
-                    mobileScannerCallback(
-                        barcodeMap,
-                        null,
-                        null,
-                        null
-                    )
+                    mobileScannerCallback(barcodeMap, null, null, null)
```
navaronbracke, line 123: formatting

example/lib/barcode_scanner_controller.dart
```diff
@@ -60,7 +60,7 @@ class _BarcodeScannerWithControllerState
 
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
-    if (!controller.value.isInitialized) {
+    if (!controller.value.hasCameraPermission) {
```
navaronbracke, line 63: If the scanner has no permission, prevent app lifecycle changes from changing the controller state.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -240,17 +239,13 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
-        startResult['size'] as Map<Object?, Object?>?;
-    final double? width = sizeInfo?['width'] as double?;
-    final double? height = sizeInfo?['height'] as double?;
-
     final Size size;
 
-    if (width == null || height == null) {
-      size = Size.zero;
-    } else {
+    if (startResult['size']
```
navaronbracke, line 244: If-case to extract the size

lib/src/mobile_scanner.dart
```diff
@@ -281,8 +281,7 @@ class _MobileScannerState extends State<MobileScanner>
 
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
-    if (widget.controller != null) return;
-    if (!controller.value.isInitialized) {
+    if (widget.controller != null || !controller.value.hasCameraPermission) {
```
navaronbracke, line 284: Using the new getter to fix a bug in the sample.

lib/src/enums/mobile_scanner_error_code.dart
```diff
@@ -24,5 +25,25 @@ enum MobileScannerErrorCode {
   permissionDenied,
 
   /// Scanning is unsupported on the current device.
-  unsupported,
+  unsupported;
+
+  /// Convert the given [PlatformException.code] to a [MobileScannerErrorCode].
+  factory MobileScannerErrorCode.fromPlatformException(
```
navaronbracke, line 31: To avoid magic strings, I added a factory constructor for this.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/MobileScannerErrorCodes.kt
```diff
@@ -0,0 +1,25 @@
+package dev.steenbakker.mobile_scanner.objects
+
+class MobileScannerErrorCodes {
```
navaronbracke, line 3: Consolidating the error codes in one place. MacOS and iOS also have an equivalent. Will be used by the new fix.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -204,7 +203,7 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
       );
     } on PlatformException catch (error) {
       throw MobileScannerException(
-        errorCode: MobileScannerErrorCode.genericError,
+        errorCode: MobileScannerErrorCode.fromPlatformException(error),
```
navaronbracke, line 206: Convert the incoming error into the right code. The Android/iOS/macOS implementations will be reporting different error codes, depending on the situation.

lib/src/objects/mobile_scanner_state.dart
```diff
@@ -60,6 +62,12 @@ class MobileScannerState {
   /// The current zoom scale of the camera.
   final double zoomScale;
 
+  /// Whether permission to access the camera was granted.
+  bool get hasCameraPermission {
```
navaronbracke, line 66: This new getter is used to fix a bug in our lifecycle handling.

lib/src/mobile_scanner_exception.dart
```diff
@@ -46,3 +46,22 @@ class MobileScannerErrorDetails {
 /// This exception type is only used internally,
 /// and is not part of the public API.
 class PermissionRequestPendingException implements Exception {}
+
+/// This class represents an exception thrown by the [MobileScannerController]
+/// when a barcode scanning error occurs when processing an input frame.
+class MobileScannerBarcodeException implements Exception {
```
navaronbracke, line 52: This error will be used by `analyzeImage()` & `start()` to report barcode scanning errors from image frames.

That will be added in the actual fix pull request.

lib/src/mobile_scanner_exception.dart
```diff
@@ -1,6 +1,6 @@
 import 'package:mobile_scanner/src/enums/mobile_scanner_error_code.dart';
 
-/// This class represents an exception thrown by the mobile scanner.
+/// This class represents an exception thrown by the [MobileScannerController].
```
navaronbracke, line 3: I cannot yet use `@docImport` but hopefully, we'll be able to address that in the future. (maybe with a lint?)

macos/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -459,20 +457,14 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         let symbologies:[VNBarcodeSymbology] = argReader.toSymbology()
         
         guard let filePath: String = argReader.string(key: "filePath") else {
-            // TODO: fix error code
-            result(FlutterError(code: "MobileScanner",
-                                message: "No image found in analyzeImage!",
-                                details: nil))
+            result(nil)
```
navaronbracke, line 460: Like on iOS, use nil

PR #1186, opened 2024-09-20
fix: clean up example button widgets

IanVS:
I noticed when experimenting with the overlay example, that the camera switching button had no color defined (and was invisible when I tried it in my app initially), and that the flash-unavailable icon did not take as much space as the flash icon button (and was a different size).  This attempts to fix those issues.

Before:

https://github.com/user-attachments/assets/090825e0-337a-4029-80b2-d3a0d14bd5f1

After:

https://github.com/user-attachments/assets/414fdff9-95e9-4968-a354-d7294533e689

### DIFF
--- a/example/lib/scanner_button_widgets.dart
+++ b/example/lib/scanner_button_widgets.dart
@@ -112,6 +112,7 @@ class SwitchCameraButton extends StatelessWidget {
         }
 
         return IconButton(
+          color: Colors.white,
           iconSize: 32.0,
           icon: icon,
           onPressed: () async {
@@ -166,10 +167,13 @@ class ToggleFlashlightButton extends StatelessWidget {
               },
             );
           case TorchState.unavailable:
-            return const Icon(
-              Icons.no_flash,
-              color: Colors.grey,
-            );
+            return const SizedBox(
+                width: 48.0,
+                child: Icon(
+                  Icons.no_flash,
+                  size: 32.0,
+                  color: Colors.grey,
+                ));
         }
       },
     );
--- a/example/lib/scanner_button_widgets.dart
+++ b/example/lib/scanner_button_widgets.dart
@@ -168,12 +168,15 @@ class ToggleFlashlightButton extends StatelessWidget {
             );
           case TorchState.unavailable:
             return const SizedBox(
-                width: 48.0,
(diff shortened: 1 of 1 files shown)


example/lib/scanner_button_widgets.dart
```diff
@@ -166,10 +167,13 @@ class ToggleFlashlightButton extends StatelessWidget {
               },
             );
           case TorchState.unavailable:
-            return const Icon(
-              Icons.no_flash,
-              color: Colors.grey,
-            );
+            return const SizedBox(
```
navaronbracke, line 170: Could you add the missing trailing comma and wrap this in a Center widget?

```suggestion
            return const SizedBox(
               width: 48.0,
               child: Center(
                  child: Icon(
                     Icons.no_flash,
                     size: 32.0,
                     color: Colors.grey,
                  ),
               ),
            );
```

navaronbracke (changes requested):

navaronbracke (approved):

navaronbracke:
Hopefully one of many! If you find the bug for the scanner overlay, fell free to send another fix in my direction.

IanVS:
Whoops, I took your suggestion without testing it, and now I see that it centers on the wrong axis.

image

IanVS:
Oh, I think it's because I didn't specify a `height` for the `SizedBox`, I'll push a fix.

IanVS:
If I can ask,  @navaronbracke, what is the purpose of wrapping the icon in `Center`?  It seems to have no effect, and `SizedBox` appears to center the child anyway, and in fact I see suggestions to use it as a way to apply padding around a widget, which is essentially what I'm doing.

IanVS:
So in this case, since there was an `size` specified on the icon, it overrides the height/width that the SizedBox is imparting, so there's no visual difference?  I need to find a good way to visualize layouts in flutter.  With CSS I'd just start giving various elements a background color (or more likely, use the inspect tool to see their bounding boxes).  Wonder if there's anything like that for this…

navaronbracke:
> I need to find a good way to visualize layouts in flutter.

I do suggest you start using the Flutter DevTools, which includes a layout explorer.

Issue #1183, opened 2024-09-20
MobileScannerState.size is not equal to BarcodeCapture.size on iOS


jsroest:
Hi,

I would like to implement a picklist mode. 

This is a scanning mode known from manufacturers like Zebra and Honeywell, where the user is able to mark and scan precisely a single barcode from a barcode sheet (a paper sheet containing a large number of barcodes). With these hardware devices, a laser crosshair is projected by the barcodescanner and only the barcode that intersects with that crosshair is scanned.

To have something similar, I project a red crosshair above the mobile_scanner live-camera feed. Touching the screen temporarily ignores any detected barcodes, to give the user the opportunity to move to the right barcode to scan. The crosshair is drawn in a green color, to give the user feedback that the scanning is temporarily disabled.

The idea is that once the user touches a single barcode with the (active) crosshair, that single barcode is scanned.

For this to work, I need the following:
1. The corners of the detected barcodes
2. The size of the image to have a reference for the corners

The corners for each barcode are passed in the BarcodeCapture.barcode.corners, which is great.

I do have issues finding the size of the image. 

I preferably do not want to receive the image in the BarcodeCapture, because I expect unnecessary performance issues with passing the image to the dart side only for having the size as a reference.

Here some samples where I place a barcode as much as possible to the lower-right corner of the camera feed.

Android 9 on a Galaxy S8
MobileScannerState.size = Size(480.0, 640.0)
BarcodeCapture.size = Size(480.0, 640.0)
BarcodeCapture.barcodes[0].corners =
[Offset(272.0, 570.0), Offset(479.0, 566.0), Offset(486.0, 631.0), Offset(280.0, 638.0)]

iOS 18, on a iPhone 15
MobileScannerState.size = Size(3024.0, 4032.0)
BarcodeCapture.size = Size(1178.0, 1572.0)
BarcodeCapture.barcodes[0].corners =
[Offset(731.0, 1409.0), Offset(1110.0, 1407.0), Offset(1115.0, 1540.0), Offset(736.0, 1547.0)]

My question boils down to the following:
1. Is it a bug that MobileScannerState.size is not equal to BarcodeCapture.size on iOS?
2. If it's not a bug, how can I get the size of the image where the barcode.corners are placed on, without setting returnImage to true in the BarcodeController.

jsroest:
I've added my implementation of the 'picklist mode' to the example app, which you can find here. 

[...]

PR #1188, opened 2024-09-23
feat: Add picklist mode sample to the example app

jsroest:
This PR adds a picklist mode sample to the sample app.

This sample is not fully optimised, because of this issue:
[MobileScannerState.size is not equal to BarcodeCapture.size on iOS](https://github.com/juliansteenbakker/mobile_scanner/issues/1183)


maxfrees:
image
Can I do a demo in this form

maxfrees:
image
Can you provide a demo like this ？

jsroest:
The latest version of the picklist sample uses the fact that the image size is now always available, also when returnImage is false. It still relies on two workarounds 'fixPortraitLandscape' and 'fixCorners'.


https://github.com/user-attachments/assets/0f1fc2e8-7f3c-4f41-b7c7-2bb45bd46229

juliansteenbakker:
Hi all, sorry for my absence, but i've tested your code and it looks like the example runs fine on android and ios! I do think we also need to solve the issues mentioned in #1183, but since this code is only an example, and it also contains a workaround for now, i think it is save to merge this into our current release, and try to fix #1183 in a next release.

PR #1189, opened 2024-09-23
fix: Constrain torch unavailable example icon

IanVS:
Fixes a bug introduced in https://github.com/juliansteenbakker/mobile_scanner/pull/1186, which caused the icon to center vertically unconstrained.

I also removed the `Center` widget, because it does not appear to have any effect inside a `SizedBox`.  But I'm happy to re-add if it's necessary.

### DIFF
--- a/example/lib/scanner_button_widgets.dart
+++ b/example/lib/scanner_button_widgets.dart
@@ -167,14 +167,12 @@ class ToggleFlashlightButton extends StatelessWidget {
               },
             );
           case TorchState.unavailable:
-            return const SizedBox(
-              width: 48.0,
-              child: Center(
-                child: Icon(
-                  Icons.no_flash,
-                  size: 32.0,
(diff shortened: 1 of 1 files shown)


example/lib/scanner_button_widgets.dart
```diff
@@ -167,14 +167,12 @@ class ToggleFlashlightButton extends StatelessWidget {
...
           case TorchState.unavailable:
-            return const SizedBox(
-              width: 48.0,
-              child: Center(
-                child: Icon(
-                  Icons.no_flash,
-                  size: 32.0,
-                  color: Colors.grey,
-                ),
+            return const SizedBox.square(
```
navaronbracke, line 170: Right, this should have been on both axes. And yes an Icon is just a SizedBox and a Text widget, so the Center was probably not needed.

navaronbracke (approved):

PR #1192, opened 2024-09-24
feat: add macos privacy manifest

### LINKED ISSUE #1178: Add empty privacy manifest for MacOS
We should add an empty privacy manifest for MacOS.

This manifest would need an empty dict for the accessed API types.

NSPrivacyTracking* and NSPrivacyCollectedDataTypes keys are required, but the values are blank. macOS explicitly does not need the NSPrivacyAccessedAPITypes

navaronbracke:
This PR adds the privacy manifest for MacOS. Built the example app to verify that the privacy manifest is included in the archive.

Fixes #1178

### DIFF
--- a/macos/mobile_scanner/Package.swift
+++ b/macos/mobile_scanner/Package.swift
@@ -17,8 +17,7 @@ let package = Package(
             name: "mobile_scanner",
             dependencies: [],
             resources: [
-                // To add other resources, see the instructions at
-                // https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package
+                .process("Resources"),
             ]
         )
     ]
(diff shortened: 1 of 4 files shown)


macos/mobile_scanner/Package.swift
```diff
@@ -17,8 +17,7 @@ let package = Package(
             name: "mobile_scanner",
             dependencies: [],
             resources: [
-                // To add other resources, see the instructions at
-                // https://developer.apple.com/documentation/xcode/bundling-resources-with-a-swift-package
+                .process("Resources"),
```
navaronbracke, line 20: Instead of having the link in here, I opted to use the `Resources` folder, like on iOS.

PR #1202, opened 2024-10-05
feat: Always return capture size

navaronbracke:
This PR updates the behavior of `returnImage` to only refer to the bytes of the output image, not the size.

The size can always be returned, to allow users to do calculations with it (for example, together with the `Barcode.size`)

Part of https://github.com/juliansteenbakker/mobile_scanner/issues/1183

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -120,7 +120,11 @@ class MobileScanner(
                 }
 
                 if (!returnImage) {
-                    mobileScannerCallback(barcodeMap, null, null, null)
+                    mobileScannerCallback(
+                        barcodeMap,
+                        null,
+                        mediaImage.width,
+                        mediaImage.height)
                     return@addOnSuccessListener
                 }
 
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -47,22 +47,17 @@ class MobileScannerHandler(
     private var analyzerResult: MethodChannel.Result? = null
 
     private val callback: MobileScannerCallback = { barcodes: List<Map<String, Any?>>, image: ByteArray?, width: Int?, height: Int? ->
-        if (image != null) {
-            barcodeHandler.publishEvent(mapOf(
-                "name" to "barcode",
-                "data" to barcodes,
-                "image" to mapOf(
-                    "bytes" to image,
-                    "width" to width?.toDouble(),
-                    "height" to height?.toDouble(),
-                )
-            ))
-        } else {
-            barcodeHandler.publishEvent(mapOf(
-                "name" to "barcode",
-                "data" to barcodes
-            ))
-        }
+        barcodeHandler.publishEvent(mapOf(
+            "name" to "barcode",
+            "data" to barcodes,
+            // The image dimensions are always provided.
+            // The image bytes are only non-null when `returnImage` is true.
+            "image" to mapOf(
+                "bytes" to image,
+                "width" to width?.toDouble(),
+                "height" to height?.toDouble(),
+            )
+        ))
     }
 
     private val errorCallback: MobileScannerErrorCallback = {error: String ->
--- a/ios/Classes/MobileScannerPlugin.swift
+++ b/ios/Classes/MobileScannerPlugin.swift
@@ -69,22 +69,18 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin {
                 return
             }
             
-            if (!MobileScannerPlugin.returnImage) {
-                barcodeHandler.publishEvent([
-                    "name": "barcode",
-                    "data": barcodesMap,
-                ])
-                return
-            }
+            // The image dimensions are always provided.
+            // The image bytes are only non-null when `returnImage` is true.
+            let imageData: [String: Any?] = [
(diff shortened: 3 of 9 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -120,7 +120,11 @@ class MobileScanner(
                 }
 
                 if (!returnImage) {
-                    mobileScannerCallback(barcodeMap, null, null, null)
+                    mobileScannerCallback(
+                        barcodeMap,
```
navaronbracke, line 124: When not returning the created bitmap, use the available dimensions from the input image, which are the same.

lib/src/mobile_scanner_controller.dart
```diff
@@ -75,8 +75,7 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
   /// If this is empty, all supported formats are detected.
   final List<BarcodeFormat> formats;
 
-  /// Whether scanned barcodes should contain the image
-  /// that is embedded into the barcode.
+  /// Whether the [BarcodeCapture.image] bytes should be provided.
```
navaronbracke, line 78: This wording was incorrect, so I adjusted it. There is no support for returning images that are embedded in the center of some specific QR codes. (i.e. little square logo's)

lib/src/objects/barcode_capture.dart
```diff
@@ -32,8 +32,6 @@ class BarcodeCapture {
   /// This is the data that was used to detect the available [barcodes], the input [image] and the [size].
   final Object? raw;
 
-  /// The size of the input [image].
-  ///
-  /// If [image] is null, this will be [Size.zero].
+  /// The size of the camera input [image].
```
navaronbracke, line 35: Adjusted the wording to match the new behavior.

lib/src/web/zxing/zxing_barcode_reader.dart
```diff
@@ -121,6 +121,7 @@ final class ZXingBarcodeReader extends BarcodeReader {
             controller.add(
               BarcodeCapture(
                 barcodes: [result.toBarcode],
+                size: videoSize,
```
navaronbracke, line 124: We already had this information in the getter implementation, so I just forwarded it. This returns `Size.zero` if the camera output is not available.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -47,22 +47,17 @@ class MobileScannerHandler(
...
-                "data" to barcodes
-            ))
-        }
+        barcodeHandler.publishEvent(mapOf(
+            "name" to "barcode",
+            "data" to barcodes,
+            // The image dimensions are always provided.
+            // The image bytes are only non-null when `returnImage` is true.
+            "image" to mapOf(
+                "bytes" to image,
```
navaronbracke, line 56: The width / height were nullable, which could have been changed. However, the Dart code already handles null, so I just left it as-is, to reduce the diff here.

lib/src/web/mobile_scanner_web.dart
```diff
@@ -143,6 +143,7 @@ class MobileScannerWeb extends MobileScannerPlatform {
     final JSArray<JSString>? facingModes = capabilities.facingModeNullable;
 
     // TODO: this is an empty array on MacOS Chrome, where there is no facing mode, but one, user facing camera.
+    // We might be able to add a workaround, using the label of the video track.
```
navaronbracke, line 146: I'm going to revisit the broken check for the transform on the web in the future. Just left a breadcrumb for myself here.

ios/Classes/MobileScannerPlugin.swift
```diff
@@ -69,22 +69,18 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin {
...
-            if (!MobileScannerPlugin.returnImage) {
-                barcodeHandler.publishEvent([
-                    "name": "barcode",
-                    "data": barcodesMap,
-                ])
-                return
-            }
+            // The image dimensions are always provided.
+            // The image bytes are only non-null when `returnImage` is true.
+            let imageData: [String: Any?] = [
```
navaronbracke, line 74: Making the bytes conditional started giving me issues with nullable Any, do I moved it to a separate dict here to fix that.

macos/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -156,22 +156,26 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
                                     "name": "barcode",
                                     "data": barcodes.map({ $0.toMap() }),
                                 ])
                                 return
                             }
-                                                        
+                            
+                            // The image dimensions are always provided.
+                            // The image bytes are only non-null when `returnImage` is true.
+                            let imageData: [String: Any?] = [
```
navaronbracke, line 169: Like on iOS, I moved the dict to a separate var to fix a warning.

macos/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -156,22 +156,26 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                         })
                         
                         DispatchQueue.main.async {
-                            if (!MobileScannerPlugin.returnImage) {
+                            guard let image = cgImage else {
```
navaronbracke, line 159: Using a guard-let to avoid `!` in the cgImage below, makes things look better

PR #1206, opened 2024-10-07
fix: bump MLKit to 7.0.0

### LINKED ISSUE #1165: GoogleDataTransport version conflict with firebase_crashlytics
After update firebase_* packages ios build was broken.

<details>
<summary>Pod error log</summary>

```log
[!] CocoaPods could not find compatible versions for pod "GoogleDataTransport":
  In Podfile:
    firebase_crashlytics (from '.symlinks/plugins/firebase_crashlytics/ios') was resolved to 4.1.0, which depends on
      Firebase/Crashlytics (= 11.0.0) was resolved to 11.0.0, which depends on
        FirebaseCrashlytics (~> 11.0.0) was resolved to 11.0.0, which depends on
          GoogleDataTransport (~> 10.0)

    mobile_scanner (from '.symlinks/plugins/mobile_scanner/ios') was resolved to 5.2.1, which depends on
      GoogleMLKit/BarcodeScanning (~> 6.0.0) was resolved to 6.0.0, which depends on
        MLKitBarcodeScanning (~> 5.0.0) was resolved to 5.0.0, which depends on
          MLKitCommon (~> 11.0) was resolved to 11.0.0, which depends on
            GoogleDataTransport (< 10.0, >= 9.4.1)
``` 

</details>

navaronbracke:
This PR updates MLKit to the latest release, which fixes dependency issues with other dependencies, such as Firebase.

However, due to MLKit requirements, the iOS minimum deployment target has been changed from 12.0 to 15.5.0.

See https://developers.google.com/ml-kit/release-notes#improvements for context.

Fixes #1165


navaronbracke:
cc @juliansteenbakker for second thought on the version number. We do need to update the minimum iOS version here.

juliansteenbakker:
I would prefer a new major version, since this is a breaking change. People that have pinned older versions of firebase could still use the current version without problems. So if we publish this as 5.2.4 or 5.3.0, we would break those instances.

On the other hand, lots of builds are failing right now, and if we publish this as a major version, a regular `flutter pub upgrade` would not fix this. So im not really sure what to do..

juliansteenbakker:
Because this only affects people using firebase_ packages, lets stick to a new major version 6.0.0. People without firebase_ packages won't be affected by the issue, and this update could break that.

navaronbracke:
@juliansteenbakker Agreed, updated the version to 6.0.0 and also added a note about the minimum XCode version in the changelog. (is mentioned in the MLKit release notes)

juliansteenbakker:
Great! LGTM!

juliansteenbakker (approved):

creativecreatorormaybenot:
@navaronbracke Thanks for your work! Will the new version be published to Pub today?

navaronbracke:
Currently it is almost 10 PM local time here, but it is first on my list for tomorrow morning! (8 AM UTC+1)

navaronbracke:
Version 6.0.0 has been published to Pub.dev

Issue #1210, opened 2024-10-10
onDetect does not handle barcode errors


AhmadFikriAS:
I get this error when using any emulator and the emulator stops when starting the mobile scanner but for my original device there are no problems. Why is that? this happened after updating to version 6.0.0 from version 5.2.0

Mobile_scanner ^6.0.0

<img width="621" alt="image" src="https://github.com/user-attachments/assets/56166684-214b-4815-8e49-806c2f3f4b67">

![cscxxcxc](https://github.com/user-attachments/assets/e106f617-977f-495c-9f55-c7fc7d1421ee)

navaronbracke:
`MOBILE_SCANNER_BARCODE_ERROR` is reported when an error occurred when something when wrong during the processing of a barcode. This is a change that we introduced in version 6.0.0, to align the error codes. (these used to be strings that all had `MobileScanner` as error code, which was not helpful)

I see you are using `onDetect`, which is documented to not handle these kinds of errors.
I do also see that the `onDetect` usage does not handle the errors, but instead leaves them as unhandled, which is undesired. That is a bug.

See https://github.com/juliansteenbakker/mobile_scanner/blob/master/lib/src/mobile_scanner.dart#L36-L44

navaronbracke:
I will release a fix for this in a moment.

PR #1213, opened 2024-10-11
fix: handle errors in onDetect; exclude armv7 for iOS

### LINKED ISSUE #1209: xArchive broken on iOS
This lib broke my iOS archive on Flutter 3.24.3.

When it is removed, I can archive normally. 

I'll try to do the changes proposed by [ml_kid](https://pub.dev/packages/google_ml_kit) doc. If it works, I can open a PR here to improve docs.

Otherwise you guys should do it, since there is nothing in the changelog proposing iOS changes.

Thanks for your time.

navaronbracke:
Fixes #1209 
Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/1210

### DIFF
--- a/example/ios/Podfile
+++ b/example/ios/Podfile
@@ -40,8 +40,8 @@ end
 post_install do |installer|
   installer.pods_project.targets.each do |target|
     flutter_additional_ios_build_settings(target)
-        target.build_configurations.each do |config|
-          config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '15.5.0'
-        end
+    target.build_configurations.each do |config|
+      config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '15.5.0'
+    end
   end
 end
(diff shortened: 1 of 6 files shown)


example/ios/Podfile
```diff
@@ -40,8 +40,8 @@ end
 post_install do |installer|
   installer.pods_project.targets.each do |target|
     flutter_additional_ios_build_settings(target)
-        target.build_configurations.each do |config|
-          config.build_settings['IPHONEOS_DEPLOYMENT_TARGET'] = '15.5.0'
-        end
+    target.build_configurations.each do |config|
```
navaronbracke, line 43: This was badly indented

lib/src/mobile_scanner.dart
```diff
@@ -34,15 +35,17 @@ class MobileScanner extends StatefulWidget {
   final MobileScannerController? controller;
 
   /// The function that signals when new codes were detected by the [controller].
-  /// If null, use the controller.barcodes stream directly to capture barcodes.
```
navaronbracke, line 37: This guidance is obsolete, as we now can handle errors in the case.

PR #1214, opened 2024-10-11
feat: read inverted datamatrix (improve #1071)

RafaRuiz:
Quoting petri-lipponen-movesense:
> Here's a quick stab at adding support for reading inverted data matrixes that the MLKit does not support. The iOS implementation is a bit clumsy (I'm not an iOS coder), but Android one is quite straight forward. Tested and works in Android & iOS. Feel free to modify, include or leave be as you will.
> 
> Future idea: add "tryInverted" which duplicates each frame and tries normal and inverted both.

And based on a old solution I had to do for an old client, sometimes codes could come either normal or inverted, so we had to scan both situations.

In my PR, which is almost [exact as petri-lipponen-movesense's](https://github.com/juliansteenbakker/mobile_scanner/pull/1071), I just modify the bool to add to invert the image into intervals.
This means that every image in the stream will change from inverted to normal.
In this case, we could either scan any of these two codes:

image

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -80,6 +80,9 @@ dependencies {
     implementation 'androidx.camera:camera-lifecycle:1.3.3'
     implementation 'androidx.camera:camera-camera2:1.3.3'
 
+    // opencv
+    implementation 'com.quickbirdstudios:opencv:3.4.15'
+
     testImplementation 'org.jetbrains.kotlin:kotlin-test'
     testImplementation 'org.mockito:mockito-core:5.12.0'    
 }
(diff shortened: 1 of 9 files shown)


navaronbracke:
I like the idea of using the invert interval, as it makes scanning both types of codes easier.

I'll try to give this a review when I find some time for it.
In the meantime can you try fixing up the conflict? Probably this PR was made off of the old PR, that wasn't updated to match the latest commit?

RafaRuiz:
> I like the idea of using the invert interval, as it makes scanning both types of codes easier.
> 
> I'll try to give this a review when I find some time for it. In the meantime can you try fixing up the conflict? Probably this PR was made off of the old PR, that wasn't updated to match the latest commit?

I'm actually investigating this issue when inverting it. iOS is good but Android is buggy.
Will fix the conflicts though
https://github.com/juliansteenbakker/mobile_scanner/pull/1071#issuecomment-2408001067

RafaRuiz:
@juliansteenbakker I didn't realise I forked an old version, as I forked @petri-lipponen-movesense's PR
So I decided to bring it down, fork it from `master` and enhance it. I was able to remove OpenCV's dependencies.

#1215

Issue #1142, opened 2024-08-11
Deprecated RenderScript APIs in Mobile Scanner Plugin


doctypepro:
_Flutter Version: 3.24.0 (Channel stable)
Kotlin Version: 2.0.10_

The mobile_scanner plugin version 5.1.1 includes usage of deprecated RenderScript APIs in its Android Kotlin codebase. The following issues are identified:

### Deprecated Imports:

```
android.renderscript.Allocation
android.renderscript.Element
android.renderscript.RenderScript
android.renderscript.ScriptIntrinsicYuvToRGB
android.renderscript.Type
```

### Deprecated API Usage:

```
RenderScript.create(context)
ScriptIntrinsicYuvToRGB.create(rs, Element.U8_4(rs))
Allocation.createTyped()
Type.Builder methods and constructors (e.g., Type.Builder(rs, Element.U8(rs)))
Allocation.USAGE_SCRIPT
```

### Warning Details:

Various warnings indicate that classes and methods from the android.renderscript package are deprecated and will be removed in future Android versions. This includes allocation, element, script, and type handling classes and methods. These deprecations suggest that the code may not be compatible with future versions of Android and could lead to runtime issues or require significant updates.

### Here are the screenshots of the issue.

<img width="607" alt="Deprecated RenderScript imports in mobile_scanner plugin." src="https://github.com/user-attachments/assets/0ae9c60f-7e73-4bd4-9a3f-97e2f389785e">
<img width="614" alt="Deprecated API usage in mobile_scanner plugin." src="https://github.com/user-attachments/assets/db3f90a6-4b17-48ee-8ca7-c5d172b936f9">

navaronbracke:
If these are all relating to the YUV to RGB converter utility that we have, I'm planing on removing that utility entirely with the switch to NV21. See also https://github.com/juliansteenbakker/mobile_scanner/issues/1116

Issue #1116, opened 2024-07-05
Android should always use NV21 image format


navaronbracke:
On Android, mobile_scanner currently has a bug, where it defaults to using YUV, instead of NV21.

Per the documentation in https://developer.android.com/reference/android/hardware/Camera.Parameters#setPreviewFormat(int)
the NV21 format is always supported. So we should always be using NV21 on Android.

Per the docs in https://developer.android.com/reference/android/graphics/ImageFormat#NV21 it seems that NV21 is the default format when no other format is requested.

The discussion in https://github.com/juliansteenbakker/mobile_scanner/issues/698 also mentions that the affected device only supports NV21.

Supersedes https://github.com/juliansteenbakker/mobile_scanner/issues/698  https://github.com/juliansteenbakker/mobile_scanner/issues/1124 and https://github.com/juliansteenbakker/mobile_scanner/issues/1077

navaronbracke:
I'm currently a bit stretched on some other work, but I can start implementing this fix at the end of this week.

I don't expect this to take too long to implement, though. We'll have to specify that we use NV21 for the camera preview and drop the conversion to YUV entirely. I don't expect it to be much else, in terms of a fix.

nijs9:
@navaronbracke  Today we encountered this issue on a Galaxy A23 (we are using mobile_scanner 5.2.3). Could you let us know if your proposed fix is already included in the latest version? Thanks!

navaronbracke:
The reporter issue is for Samsung A23 devices. I do not yet have an update, although I want this fixed for v5.3.0

alanlanglois:
Also happening on a Samsung A54 5G Android 14.

@navaronbracke thanks for the hard work! Do you have any ETA for the 5.3.0?

pgiacomo69:
@navaronbracke for me it's critical too, if you. have a working version want I can test it in the field, distributing a Beta App to affected users.

navaronbracke:
I am currently looking into landing a fix for a different issue, but after that, fixing the NV21 formats bug is next on my list.

navaronbracke:
I had another look at this issue and also had a look at what the `camera` plugin does.

CameraX always sends back `YUV_420_888` image data, which cannot be changed.
However, the Android implementation of `camera` always converts `YUV_420_888` to NV21, for sending back to Dart (in case people want to process frames on the Dart side). Conveniently, MLKit also requires NV21 as the format for the input images.

[...]

PR #1215, opened 2024-10-12
feat: Ability to scan both normal codes and inverted codes

RafaRuiz:
Started on #1214 to improve #1071 but with a few enhancements:
- No usage of OpenCV for Android, MLKit would do the inversion.
- Set `intervalInvertImage` instead of choosing whether to invert the image constantly or not. This way we can read both types, for example:
<details>
<summary>Both types of codes (example)</summary>

image

</details>

- Kept to latest version 6.0.1 instead from the first one which was taken from 5.0.1


lib/src/objects/start_options.dart
```diff
@@ -23,6 +24,9 @@ class StartOptions {
   /// The desired camera resolution for the scanner.
   final Size? cameraResolution;
 
+  /// Whether the scanner should work also with inverted images.
```
navaronbracke, line 27: This doc comment can be improved.
```suggestion
  /// Whether the scanner should try to detect color-inverted barcodes in every other frame.
```

lib/src/mobile_scanner_platform_interface.dart
```diff
@@ -111,6 +111,11 @@ abstract class MobileScannerPlatform extends PlatformInterface {
     throw UnimplementedError('updateScanWindow() has not been implemented.');
   }
 
+  /// Set inverting image colors in intervals (for negative Data Matrices).
+  Future<void> setIntervalInvertImage(bool intervalInvertImage) {
```
navaronbracke, line 115: I don't think a new method on the public interface is needed?

Unless you have a specific use case for it?

lib/src/mobile_scanner_controller.dart
```diff
@@ -82,6 +83,11 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
   /// Defaults to false, and is only supported on iOS, MacOS and Android.
   final bool returnImage;
 
+  /// Whether the image should be inverted in intervals (original - inverted - original…)
```
navaronbracke, line 86:
```suggestion
  /// Whether the scanner should try to detect color-inverted barcodes in every other frame.
  ///
  /// When this option is enabled, every odd frame from the camera preview has its colors inverted before processing.
  /// This is useful if barcodes can be both black-on-white (the most common) and white-on-black (less common).
  /// Usage of this parameter can incur a performance cost, as some frames need to be altered further during processing.
  ///
  /// Defaults to false and is only supported on Android and iOS.
```

We should provide users with a detailed description of what this does.

lib/src/mobile_scanner_controller.dart
```diff
@@ -82,6 +83,11 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+  ///
+  /// Defaults to false.
+  final bool intervalInvertImage;
```
navaronbracke, line 89: I find the name `intervalInvertImage` a bit confusing for the public API.

What about `scanInverted` ? Open to better name suggestions.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -77,7 +81,17 @@ class MobileScanner(
     @ExperimentalGetImage
     val captureOutput = ImageAnalysis.Analyzer { imageProxy -> // YUV_420_888 format
         val mediaImage = imageProxy.image ?: return@Analyzer
-        val inputImage = InputImage.fromMediaImage(mediaImage, imageProxy.imageInfo.rotationDegrees)
+
+        // Inversion
```
navaronbracke, line 85:
```suggestion
        // Invert every other frame.
```

The comment next to the assignment can be removed. We can just keep the comment at the top.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -58,6 +60,8 @@ class MobileScanner(
 
     /// Configurable variables
     var scanWindow: List<Float>? = null
+    var intervalInvertImage: Boolean = false
+    private var invertImage: Boolean = false
```
navaronbracke, line 64:
```suggestion
    private var invertCurrentImage: Boolean = false
```

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -462,6 +478,54 @@ class MobileScanner(
...
+        val nv21 = ByteArray(ySize + uSize + vSize)
+
+        yBuffer.get(nv21, 0, ySize)
+        vBuffer.get(nv21, ySize, vSize)
+        uBuffer.get(nv21, ySize + vSize, uSize)
+
+        return nv21
+    }
+
+    // Helper function to invert image data
```
navaronbracke, line 524:
```suggestion
    private fun invertBytes(bytes: ByteArray): ByteArray {}
```

We can omit the comment if the method says what it does :)

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -462,6 +478,54 @@ class MobileScanner(
...
+     */
+    @SuppressLint("UnsafeOptInUsageError")
+    fun invertInputImage(imageProxy: ImageProxy): InputImage {
+        // Extract Image from ImageProxy
+        val image = imageProxy.image ?: throw IllegalArgumentException("Image is null")
+
+        // Convert YUV_420_888 image to NV21 format
+        val imageByteArray = yuv420888toNV21(image)
+
+        // Invert the cropped image
```
navaronbracke, line 492: The image isn't really cropped? So this comment is misleading.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -462,6 +478,54 @@ class MobileScanner(
...
+            invertedBytes,
+            image.width,
+            image.height,
+            imageProxy.imageInfo.rotationDegrees,
+            InputImage.IMAGE_FORMAT_NV21
+        )
+    }
+
+    // Helper function to convert YUV_420_888 to NV21
+    private fun yuv420888toNV21(image: Image): ByteArray {
```
navaronbracke, line 506: We have an open bug for avoiding the use of `yuv420_888`: https://github.com/juliansteenbakker/mobile_scanner/issues/1116

I'd rather fix that bug first, instead of adding a new dependency on YUV here.
If you don't want to fix that bug, I would suggest we move this conversion from YUV to the
utils folder at https://github.com/juliansteenbakker/mobile_scanner/tree/master/android/src/main/kotlin/dev/steenbakker/mobile_scanner/utils

When we fix that other formats bug, we can just delete all the YUV related utils.

ios/Classes/MobileScannerPlugin.swift
```diff
@@ -117,17 +117,28 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin {
...
         case "updateScanWindow":
             updateScanWindow(call, result)
+        case "setIntervalInvertImage":
+            setIntervalInvertImage(call, result)
         default:
             result(FlutterMethodNotImplemented)
         }
     }
 
+    func convertCIImageToCGImage(inputImage: CIImage) -> CGImage? {
```
navaronbracke, line 127: This function seems to be unused? It is a copy of the one in MobileScanner.swift

ios/Classes/MobileScanner.swift
```diff
@@ -120,6 +124,14 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     func requestPermission(_ result: @escaping FlutterResult) {
         AVCaptureDevice.requestAccess(for: .video, completionHandler: { result($0) })
     }
+
+    func convertCIImageToCGImage(inputImage: CIImage) -> CGImage? {
```
navaronbracke, line 128: Can this be private?

ios/Classes/MobileScanner.swift
```diff
@@ -136,10 +148,20 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
...
-            let ciImage = latestBuffer.image
 
-            let image = VisionImage(image: ciImage)
+            // Inversion
+            let uiImage : UIImage
+            if (intervalInvertImage) {
+               invertImage = !invertImage
+            }
+            if (invertImage) {
+                let tempImage = self.invertImage(image: latestBuffer.image)
```
navaronbracke, line 158: This temp variable is redundant

ios/Classes/MobileScanner.swift
```diff
@@ -48,6 +48,10 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
 
     var detectionSpeed: DetectionSpeed = DetectionSpeed.noDuplicates
 
+    /// Analyze inverted image intervally to include both inverted and normal images
```
navaronbracke, line 51:
```suggestion
var intervalInvertImage: Bool = false
private var invertCurrentImage: Bool = false
```

Same name suggestion as on Android. The comments can also be removed.

ios/Classes/MobileScanner.swift
```diff
@@ -136,10 +148,20 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
 
             nextScanTime = currentTime + timeoutSeconds
             imagesCurrentlyBeingProcessed = true
-            
-            let ciImage = latestBuffer.image
 
-            let image = VisionImage(image: ciImage)
+            // Inversion
```
navaronbracke, line 152:
```suggestion
            // Invert every other frame.
```

ios/Classes/MobileScanner.swift
```diff
@@ -434,16 +461,30 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     /// Analyze a single image
     func analyzeImage(image: UIImage, position: AVCaptureDevice.Position,
                       barcodeScannerOptions: BarcodeScannerOptions?, callback: @escaping BarcodeScanningCallback) {
-        let image = VisionImage(image: image)
-        image.orientation = imageOrientation(
+        var uiImage = image
+        if (invertImage) {
+            uiImage = self.invertImage(image: uiImage)
+        }
+        let visImage = VisionImage(image: uiImage)
```
navaronbracke, line 468:
```suggestion
        let visionImage = VisionImage(image: uiImage)
```

ios/Classes/MobileScanner.swift
```diff
@@ -434,16 +461,30 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     /// Analyze a single image
     func analyzeImage(image: UIImage, position: AVCaptureDevice.Position,
                       barcodeScannerOptions: BarcodeScannerOptions?, callback: @escaping BarcodeScanningCallback) {
-        let image = VisionImage(image: image)
-        image.orientation = imageOrientation(
+        var uiImage = image
```
navaronbracke, line 464:
```suggestion
let uiImage: UiImage

if (invertImage) {
  uiImage = self.invertImage(image: image)
} else {
  uiImage = image
}
```

Can you also use a `let` here?

ios/Classes/MobileScanner.swift
```diff
@@ -434,16 +461,30 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
...
             position: position
         )
         
         let scanner: BarcodeScanner = barcodeScannerOptions != nil ? BarcodeScanner.barcodeScanner(options: barcodeScannerOptions!) : BarcodeScanner.barcodeScanner()
 
-        scanner.process(image, completion: callback)
+        scanner.process(visImage, completion: callback)
+    }
+
+    func invertImage(image: UIImage) -> UIImage {
```
navaronbracke, line 480: Can this be private?

ios/Classes/MobileScanner.swift
```diff
@@ -434,16 +461,30 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
...
+        let ciImage = CIImage(image: image)
+        let filter = CIFilter(name: "CIColorInvert")
```
navaronbracke, line 482: I would not use a magic string here, you should be using https://developer.apple.com/documentation/coreimage/cifilter/3228292-colorinvert

navaronbracke (changes requested):
Overall, this looks okay, but I have feedback for the implementation.

Will you be looking into color inversion for the Vision API (for MacOS) as well?

lib/src/mobile_scanner_controller.dart
RafaRuiz: `interval` is the word that came to me.
maybe `shouldConsiderInvertedImages: bool`?
`scanInverted` looks like it would work only with inverted colour images

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
RafaRuiz: Nope! I cropped it to avoid the bounds and see results fast. I'll fix it 👍

lib/src/mobile_scanner_controller.dart
navaronbracke: That would be sufficient for a name.

lib/src/mobile_scanner_platform_interface.dart
RafaRuiz: Yes, throughout the comments I've seen that some users generate their own **normal** QR codes and it's at their use, so I thought why allowing inverted images if it's not going to be used for them, it's a waste of energy.

ios/Classes/MobileScanner.swift
RafaRuiz: I kept the private temp var's comment and enhanced it. Let me know if that's alright
image

ios/Classes/MobileScanner.swift
RafaRuiz: @navaronbracke can you give me a hand here?

I'm trying to make it API available, as I've seen there are a few statements for iOS 13 somewhere in the project, but despite setting my project deployment target to 15, I keep seeing this:

image

I've done just a little Swift in my life, it's been more Android

lib/src/mobile_scanner_platform_interface.dart
```diff
@@ -116,6 +116,11 @@ abstract class MobileScannerPlatform extends PlatformInterface {
     throw UnimplementedError('updateScanWindow() has not been implemented.');
   }
 
+  /// Set inverting image colors in intervals (for negative Data Matrices).
```
navaronbracke, line 119:
```suggestion
  /// Enable or disable the inverting of image colors.
  ///
  /// This is useful when working with negative-color Data Matrices.
  /// See also: https://en.wikipedia.org/wiki/Negative_(photography)
```

lib/src/mobile_scanner_controller.dart
```diff
@@ -82,6 +83,17 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
 
+  /// Whether the scanner should try to detect color-inverted barcodes in every other frame.
+  ///
+  /// When this option is enabled, every odd frame from the camera preview has its colors inverted before processing.
+  /// This is useful if barcodes can be both black-on-white (the most common) and white-on-black (less common).
+  /// Usage of this parameter can incur a performance cost, as some frames need to be altered further during processing.
+  ///
+  /// Defaults to false and is only supported on Android and iOS.
+  ///
+  /// Defaults to false.
```
navaronbracke, line 94: The second `Defaults to false.` can be removed.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -291,6 +291,14 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
     );
   }
 
+  @override
+  Future<void> setShouldConsiderInvertedImages(bool shouldConsiderInvertedImages) async {
+    await methodChannel.invokeMethod<void>(
+      'setShouldConsiderInvertedImages',
+      {'shouldConsiderInvertedImages': shouldConsiderInvertedImages},
```
navaronbracke, line 298: Can we just pass in a boolean flag instead of a Map here?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -79,7 +83,17 @@ class MobileScanner(
     @ExperimentalGetImage
     val captureOutput = ImageAnalysis.Analyzer { imageProxy -> // YUV_420_888 format
         val mediaImage = imageProxy.image ?: return@Analyzer
-        val inputImage = InputImage.fromMediaImage(mediaImage, imageProxy.imageInfo.rotationDegrees)
+
+        // Invert every other frame.
```
navaronbracke, line 87:
```suggestion
        // Invert every other frame, by flipping the flag on every input frame.
```

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -79,7 +83,17 @@ class MobileScanner(
...
+        if (shouldConsiderInvertedImages) {
+            invertCurrentImage = !invertCurrentImage // so we jump from one normal to one inverted and viceversa
```
navaronbracke, line 89:
```suggestion
            invertCurrentImage = !invertCurrentImage
```

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -510,6 +526,45 @@ class MobileScanner(
...
+     * Inverts the image colours respecting the alpha channel
+     */
+    @SuppressLint("UnsafeOptInUsageError")
+    fun invertInputImage(imageProxy: ImageProxy): InputImage {
+        val image = imageProxy.image ?: throw IllegalArgumentException("Image is null")
+
+        // Convert YUV_420_888 image to NV21 format
+        // based on our util helper
+        val bitmap = Bitmap.createBitmap(image.width, image.height, Bitmap.Config.ARGB_8888)
+        YuvToRgbConverter(activity).yuvToRgb(image, bitmap)
```
navaronbracke, line 539: This adds another usage of the deprecated RenderScript API's. Can we avoid this?

See #1142

ios/Classes/MobileScannerPlugin.swift
```diff
@@ -177,6 +180,19 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin {
         result(nil)
     }
 
+    /// Sets the zoomScale.
+    private func setShouldConsiderInvertedImages(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
+        let shouldConsiderInvertedImages = call.arguments as? Bool
+        if (shouldConsiderInvertedImages == nil) {
```
navaronbracke, line 186: Can we guard `mobileScanner.setShouldConsiderInvertedImages(shouldConsiderInvertedImages!)` to not do anything if the argument is null ? Otherwise this does not match Android.

ios/Classes/MobileScanner.swift
```diff
@@ -48,6 +48,12 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
 
     var detectionSpeed: DetectionSpeed = DetectionSpeed.noDuplicates
 
+    var shouldConsiderInvertedImages: Bool = false
+    // local variable to invert this image only this time,
```
navaronbracke, line 52: I think this can be a one line comment.

```suggestion
    // This flag is used to determine if the current frame should be color-inverted.
```

navaronbracke (changes requested):
@juliansteenbakker This entire implementation would omit MacOS, even though that should be supported there also, considering the API's we use.

Since this is a draft, can we close this PR and reopen this on the dev branch instead?

juliansteenbakker:
Sounds good to me @navaronbracke. The android implementation can be a lot faster, i will work on it. I'll first merge the changes from master into develop, before fixing the merge conflicts on this PR.

juliansteenbakker:
@navaronbracke do you see any usecase for the `setShouldConsiderInvertedImages` function in the interface? I don't think we would want to set options via functions, it would be best to just let the `MobileScannerController` handle this with `shouldConsiderInvertedImages` parameter.

navaronbracke:
@juliansteenbakker I agree, you'd probably want to set it up on the controller itself. That follows the same pattern for the other options.

juliansteenbakker:
I have added some improvements, and i think this is the best form we will get with this implementation. Yes, android.renderscript is deprecated, but with the current implementation it's the best/fastest way of transforming images. I think we still need to migrate to something new, but that would be a new issue/pr to fix this. #1142

lib/src/mobile_scanner_controller.dart
```diff
@@ -84,6 +85,15 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
   /// Defaults to false, and is only supported on iOS, MacOS and Android.
   final bool returnImage;
 
+  /// Whether the scanner should try to detect color-inverted barcodes in every other frame.
```
navaronbracke, line 88: Is this wording still correct? As it stands now, Android uses the flag, but never for "every other" frame.
When it is enabled, it currently applies the inversion for every frame.

navaronbracke (changes requested):
The change overall is LGTM. I just have some minor feedback.

example/lib/barcode_scanner_controller.dart
```diff
@@ -18,7 +18,7 @@ class _BarcodeScannerWithControllerState
     extends State<BarcodeScannerWithController> with WidgetsBindingObserver {
   final MobileScannerController controller = MobileScannerController(
     autoStart: false,
-    torchEnabled: true,
```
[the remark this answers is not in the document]
juliansteenbakker: After testing many times in the evening i finally got fed up with the flash in my eyes, so it's about time!

navaronbracke:
@RafaRuiz iOS & MacOS use the Vision API. That API should already support color inverted barcodes IIRC.
See also https://developer.apple.com/documentation/vision/vnbarcodeobservation/iscolorinverted

RafaRuiz:
> @RafaRuiz iOS & MacOS use the Vision API. That API should already support color inverted barcodes IIRC. See also https://developer.apple.com/documentation/vision/vnbarcodeobservation/iscolorinverted

I don't understand. Apparently the inverted images would be working on Android with this library and not on iOS because there hasn't been any code bridged to it.

PR #1216, opened 2024-10-12
chore: update gradle and java in example app

aprxty3:
this update for handle gradle issue on new flutter version and on mac device when build on android

### DIFF
--- a/example/android/app/build.gradle
+++ b/example/android/app/build.gradle
@@ -27,12 +27,12 @@ android {
     compileSdk 34
 
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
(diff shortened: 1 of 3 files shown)


example/android/app/build.gradle
```diff
@@ -42,7 +42,7 @@ android {
     defaultConfig {
         // TODO: Specify your own unique Application ID (https://developer.android.com/studio/build/application-id.html).
         applicationId "dev.steenbakker.mobile_scanner_example"
-        minSdkVersion 21
+        minSdkVersion 24
```
navaronbracke, line 45: I assume this is needed for Gradle 8.9 or higher?

aprxty3: yups... when i try the example code, i get an error cant build the app, then i'm improvised the gradle then it work's (can build the app).

PR #1177, opened 2024-09-10
fix: support formats for analyze image

navaronbracke:
This PR adds support for specifying the formats for analyze image.

It also adds the (normalized) barcode size for the barcodes on the web and adds support for analyze image on MacOS.

Fixes #1069

PR #1220, opened 2024-10-14
perf:  named Parameter formats in analyzeImage of Controller

abhiaggoel:
Added Named Parameter formats to analyzeImage method of Controller


navaronbracke:
I had forgotten to add this in https://github.com/juliansteenbakker/mobile_scanner/pull/1177

Good catch!

navaronbracke:
Can you update the patch version in the pubspec and podspecs for iOS/macOS ?
We will also need to add an entry in the changelog.

navaronbracke:
I'll be including a similar fix for the next patch version, so I'm going to close this PR.

Thanks for bringing this to my attention!

Issue #1185, opened 2024-09-20
Overlay example does not apply background opacity correctly


IanVS:
I'm using the overlay example as a guide for building a feature in my app, and I noticed that the background opacity is not being applied outside of the scan window correctly:

<img width="328" alt="Android Studio 2024-09-20 10 50 43" src="https://github.com/user-attachments/assets/88aaa860-63b4-4842-92b9-10d89d3b3413">

I'm a beginner flutter dev, so I don't know what might be causing this.

navaronbracke:
This could be a bug in the example. Although, I think that the example just draws the white rectangle, and nothing else.

IanVS:
From what I can tell, it does _try_ to apply some opacity.  

https://github.com/juliansteenbakker/mobile_scanner/blob/c3271f0cda3df5594a4cebc02696007657785861/example/lib/mobile_scanner_overlay.dart#L147C21-L147C41

RobertCer:
In `ScannerOverlay` example class, changing from:
```dart
final backgroundPaint = Paint()
  ..color = Colors.black.withOpacity(0.5)
  ..style = PaintingStyle.fill
  ..blendMode = BlendMode.dstOut;
```
to
```dart
final backgroundPaint = Paint()
  ..color = Colors.black.withOpacity(0.5)
  ..style = PaintingStyle.fill
  ..blendMode = BlendMode.srcOver; // <- This line
```
will add the background opacity.

IanVS:
@RobertCer I'm not having any luck with that, did you change anything else as well?

giorgiNavdarashvili09:
@IanVS  that will add overlay outside of camera preview as well and might not be desired outcome for people who have scaffold background other than black

PR #1230, opened 2024-11-01
fix: scanner overlay example

IanVS:
Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/1185

I also updated the other usage of Rect.largest in `BarcodeScannerWithScanWindow`, even though it wasn't visually broken previously.  I figured it was better to keep them consistent, though.

|Before | After|
|------|-------|
| image  |   image   |
|  image  | image |


navaronbracke (approved):
Thanks for the fix!

PR #1242, opened 2024-11-20
feat: add improvement for `analyzeImage` function

dwikyhardi:
Add `addBarcode` method. This method is used for adding a barcode to the `_barcodesController` streams. so when using `analyzeImage` the return value can added and be used in `MobileScanner` using `onDetect` callback.

### DIFF
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -439,4 +439,17 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
 
     await MobileScannerPlatform.instance.dispose();
   }
+
+  /// Exposes the [addBarcode] method.
+  ///
+  /// This method is used for adding a barcode to the barcode stream.
+  /// so when using [analyzeImage] the return value
+  /// can be used in [MobileScanner] using [onDetect] callback.
+  void addBarcode(BarcodeCapture? barcode) {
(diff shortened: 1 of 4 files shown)


lib/src/mobile_scanner_controller.dart
```diff
@@ -439,4 +439,17 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
 
     await MobileScannerPlatform.instance.dispose();
   }
+
+  /// Exposes the [addBarcode] method.
+  ///
+  /// This method is used for adding a barcode to the barcode stream.
+  /// so when using [analyzeImage] the return value
+  /// can be used in [MobileScanner] using [onDetect] callback.
+  void addBarcode(BarcodeCapture? barcode) {
```
navaronbracke, line 448: I don't understand the reasoning behind this change? Why would you want these events to be added to the stream of barcodes that are coming from the camera preview?

The `analyzeImage` method is _distinctly_ different from the barcode stream, because the barcodes have a different source. Besides, if you aboslutely want this behavior, you can just listen to the barcodes stream and forward events to your own StreamController.

navaronbracke:
I'm going to close this PR, as it goes against the definition of what `analyzeImage()` actually is.

If people absolutely want this change, they can always forward events from `analyzeImage` to their own stream controller, that also captures events from the barcodes stream.

PR #1272, opened 2025-01-04
fix: remove unused child argument

### LINKED ISSUE #1256: `placeholderBuilder` and `errorBuilder` callbacks for `MobileScanner` have an extra argument which serves no purpose.
This is what the `build` method of the `MobileScanner` widget looks like:

```dart
@override
  return ValueListenableBuilder<MobileScannerState>(
    valueListenable: controller,
    builder: (BuildContext context, MobileScannerState value, Widget? child) {
      if (!value.isInitialized) {
        const Widget defaultPlaceholder = ColoredBox(color: Colors.black);

        return widget.placeholderBuilder?.call(context, child) ??
            defaultPlaceholder;
      }

      final MobileScannerException? error = value.error;

      if (error != null) {
        const Widget defaultError = ColoredBox(
          color: Colors.black,
          child: Center(child: Icon(Icons.error, color: Colors.white)),
        );

        return widget.errorBuilder?.call(context, error, child) ??
            defaultError;
      }

      return LayoutBuilder(
        builder: (context, constraints) {
          _maybeUpdateScanWindow(value, constraints);

          final Widget? overlay =
              widget.overlayBuilder?.call(context, constraints);
          final Size cameraPreviewSize = value.size;

          final Widget scannerWidget = ClipRect(
            child: SizedBox.fromSize(
              size: constraints.biggest,
              child: FittedBox(
                fit: widget.fit,
                child: SizedBox(
                  width: cameraPreviewSize.width,
                  height: cameraPreviewSize.height,
                  child: MobileScannerPlatform.instance.buildCameraView(),
                ),
              ),
            ),
          );

          if (overlay == null) {
            return scannerWidget;
          }

          return Stack(
            alignment: Alignment.center,
            children: <Widget>[
              scannerWidget,
              overlay,
            ],
          );
        },
      );
    },
  );
}
```

Since `child` is not passed to the `ValueListenableBuilder`, it will always be `null`. There is no point in making `placeholderBuilder` and `errorBuilder` callbacks take an extra argument.

navaronbracke:
Fixes #1256

### DIFF
--- a/example/lib/barcode_scanner_controller.dart
+++ b/example/lib/barcode_scanner_controller.dart
@@ -55,7 +55,7 @@ class _BarcodeScannerWithControllerState
         children: [
           MobileScanner(
             controller: controller,
-            errorBuilder: (context, error, child) {
+            errorBuilder: (context, error) {
               return ScannerErrorWidget(error: error);
             },
             fit: BoxFit.contain,
--- a/example/lib/barcode_scanner_controller.dart
+++ b/example/lib/barcode_scanner_controller.dart
@@ -65,7 +65,7 @@ class _BarcodeScannerWithControllerState
             child: Container(
               alignment: Alignment.bottomCenter,
               height: 100,
-              color: Colors.black.withOpacity(0.4),
+              color: const Color.fromRGBO(0, 0, 0, 0.4),
               child: Row(
                 mainAxisAlignment: MainAxisAlignment.spaceEvenly,
                 children: [
--- a/example/lib/scanner_error_widget.dart
+++ b/example/lib/scanner_error_widget.dart
@@ -19,7 +19,6 @@ class ScannerErrorWidget extends StatelessWidget {
         errorMessage = 'Scanning is unsupported on this device';
       default:
         errorMessage = 'Generic Error';
-        break;
     }
 
     return ColoredBox(
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -8,13 +8,6 @@ import 'package:mobile_scanner/src/objects/barcode_capture.dart';
 import 'package:mobile_scanner/src/objects/mobile_scanner_state.dart';
 import 'package:mobile_scanner/src/scan_window_calculation.dart';
 
-/// The function signature for the error builder.
-typedef MobileScannerErrorBuilder = Widget Function(
-  BuildContext,
-  MobileScannerException,
-  Widget?,
-);
-
 /// This widget displays a live camera preview for the barcode scanner.
 class MobileScanner extends StatefulWidget {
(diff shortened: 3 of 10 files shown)


lib/src/mobile_scanner.dart
```diff
@@ -8,13 +8,6 @@ import 'package:mobile_scanner/src/objects/barcode_capture.dart';
 import 'package:mobile_scanner/src/objects/mobile_scanner_state.dart';
 import 'package:mobile_scanner/src/scan_window_calculation.dart';
 
-/// The function signature for the error builder.
-typedef MobileScannerErrorBuilder = Widget Function(
```
navaronbracke, line 12: Now that the function signature is short enough to fit on one line, I opted to remove the typedef

lib/src/mobile_scanner.dart
```diff
@@ -73,7 +66,7 @@ class MobileScanner extends StatefulWidget {
   /// If this is null, a black [ColoredBox] is used as placeholder.
   ///
   /// The placeholder is displayed when the camera preview is being initialized.
-  final Widget Function(BuildContext, Widget?)? placeholderBuilder;
+  final WidgetBuilder? placeholderBuilder;
```
navaronbracke, line 69: Reusing an exiting typedef from the framework here

lib/src/mobile_scanner.dart
```diff
@@ -203,12 +196,11 @@ class _MobileScannerState extends State<MobileScanner>
   Widget build(BuildContext context) {
     return ValueListenableBuilder<MobileScannerState>(
       valueListenable: controller,
-      builder: (BuildContext context, MobileScannerState value, Widget? child) {
+      builder: (BuildContext context, MobileScannerState value, _) {
```
navaronbracke, line 199: Users can create a child widget above the `MobileScanner` widget in their build method, which makes passing in a child widget redundant here. And as reported, the `child` parameter was unused, so it didn't add any value to the function signatures

example/lib/barcode_scanner_controller.dart
```diff
@@ -65,7 +65,7 @@ class _BarcodeScannerWithControllerState
             child: Container(
               alignment: Alignment.bottomCenter,
               height: 100,
-              color: Colors.black.withOpacity(0.4),
```
navaronbracke, line 68: withOpacity is deprecated, so I replaced it everywhere

example/lib/scanner_error_widget.dart
```diff
@@ -19,7 +19,6 @@ class ScannerErrorWidget extends StatelessWidget {
         errorMessage = 'Scanning is unsupported on this device';
       default:
         errorMessage = 'Generic Error';
-        break;
```
navaronbracke, line 22: The linter now reports an obsolete break statement here

Issue #1244, opened 2024-11-21
`analyzeImage` not throw error when picking non-Qr image on iOS


congnguyen-prismtech:
I'm using `analyzeImage` to read a non-QR after picking it on iOS. The app not throw any exception. So that i can't show the toast when picked failure. Please help me solve it.

navaronbracke:
In that case you probably receive `null` as result? This is intentional behavior, though, so that users don't have to handle an extra error. You probably want to chek that the file is an image file, before passing it to analyze image.

congnguyen-prismtech:
Hi @navaronbracke there is a different between Android and iOS. 
in Android, after picking a non-QR code, it continue execute the next line. So that i can show a toast: **"Your QR code is not valid"**.
Meanwhile, in iOS, the function is stopped at `analyzeImage` and not execute the next line whenever i pick a non-QR code. Therefore, it can't show the toast: **"Your QR code is not valid"**.
My expectation is can show that toast whenever i pick a non-QR code in iOS.

navaronbracke:
I see. Yes, this should not have behavioral differences between platforms.

navaronbracke:
On Android it returns the error from MLKit:
https://github.com/juliansteenbakker/mobile_scanner/blob/master/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt#L487 and https://github.com/juliansteenbakker/mobile_scanner/blob/master/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt#L30-L35

But on the old iOS implementation, it does not: https://github.com/juliansteenbakker/mobile_scanner/blob/master/ios/Classes/MobileScanner.swift#L435-L447

On the new iOS/MacOS version (which uses the Vision API), we return nil:
https://github.com/juliansteenbakker/mobile_scanner/blob/ios-vision/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift#L652-L666

I'll probably fix this in the new implementation, as the Vision API generally outperforms the MLKit one.

Thanks for reporting!

lpl-code-github:
Even though I selected an image with a QR code, I didn't analyze it? Always waiting

navaronbracke:
[...]

PR #1273, opened 2025-01-05
fix: emit an error when CIImage cannot be created

navaronbracke:
This PR adds the following changes:
- adjusts the iOS / MacOS implementation to throw an error when the provided file path is not an image.
- adjusts the Android implementation to throw an error when the provided file path is not an image. As reported in the issue, it was hanging previously (due to an unhandled IOException)
- enables Swift Package Manager support in the example app

Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/1244

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -7,12 +7,10 @@ import android.graphics.Matrix
 import android.graphics.Rect
 import android.hardware.display.DisplayManager
 import android.net.Uri
-import android.os.Build
 import android.os.Handler
 import android.os.Looper
 import android.util.Size
 import android.view.Surface
-import android.view.WindowManager
 import androidx.annotation.VisibleForTesting
 import androidx.camera.core.Camera
 import androidx.camera.core.CameraSelector
@@ -32,10 +30,12 @@ import com.google.mlkit.vision.barcode.BarcodeScanning
 import com.google.mlkit.vision.barcode.common.Barcode
 import com.google.mlkit.vision.common.InputImage
 import dev.steenbakker.mobile_scanner.objects.DetectionSpeed
+import dev.steenbakker.mobile_scanner.objects.MobileScannerErrorCodes
 import dev.steenbakker.mobile_scanner.objects.MobileScannerStartParameters
 import dev.steenbakker.mobile_scanner.utils.YuvToRgbConverter
 import io.flutter.view.TextureRegistry
 import java.io.ByteArrayOutputStream
+import java.io.IOException
 import kotlin.math.roundToInt
 
 class MobileScanner(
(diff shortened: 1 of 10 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -7,12 +7,10 @@ import android.graphics.Matrix
 import android.graphics.Rect
 import android.hardware.display.DisplayManager
 import android.net.Uri
-import android.os.Build
```
navaronbracke, line 10: Unused imports

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -432,7 +432,15 @@ class MobileScanner(
         scannerOptions: BarcodeScannerOptions?,
         onSuccess: AnalyzerSuccessCallback,
         onError: AnalyzerErrorCallback) {
-        val inputImage = InputImage.fromFilePath(activity, image)
+        val inputImage: InputImage
+
+        try {
+            inputImage = InputImage.fromFilePath(activity, image)
```
navaronbracke, line 438: Per the docs of `InputImage.fromFilePath` it throws an `IOException` if the path is not an image (with the error message "The image Uri could not be resolved.")

I believe not handling this here is why the method was not returning (we get an unhandled error, instead of calling the callbacks that would resolve the method call)

example/pubspec.yaml
```diff
@@ -7,7 +7,7 @@ version: 0.0.1
 
 environment:
   sdk: ">=3.3.0 <4.0.0"
-  flutter: ">=3.19.0"
+  flutter: ">=3.27.0"
```
navaronbracke, line 10: Since Swift Package Manager support is now available on stable, I updated the constraint here

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -16,6 +16,7 @@
 		97C147011CF9000F007C117D /* LaunchScreen.storyboard in Resources */ = {isa = PBXBuildFile; fileRef = 97C146FF1CF9000F007C117D /* LaunchScreen.storyboard */; };
 		A277671F4E603C31E191FCBF /* Pods_Runner.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = FB50C06655B1959BAA8EAF64 /* Pods_Runner.framework */; };
 		F8C489EFA0FC3A131EBA1838 /* Pods_RunnerTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9F20A7A668B7797BBB5443C1 /* Pods_RunnerTests.framework */; };
+		78A318202AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage in Frameworks */ = {isa = PBXBuildFile; productRef = 78A3181F2AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage */; };
```
navaronbracke, line 19: The changes in this file were generated by the Swift Package manager project migrator, during a `flutter run`.

example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -5,6 +5,24 @@
    <BuildAction
       parallelizeBuildables = "YES"
       buildImplicitDependencies = "YES">
+      <PreActions>
```
navaronbracke, line 8: This change was generated by the Swift Package manager project migrator, during a `flutter run`.

PR #1274, opened 2025-01-05
fix: do not hang on analyze image if there are no results

### LINKED ISSUE #1244: `analyzeImage` not throw error when picking non-Qr image on iOS

navaronbracke:
Fixes #1244

### DIFF
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -665,17 +665,31 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                 }
                     
                 guard let barcodes: [VNBarcodeObservation] = request.results as? [VNBarcodeObservation] else {
+                    DispatchQueue.main.async {
+                        result([
+                            "name": "barcode",
+                            "data": [],
+                        ])
+                    }
                     return
                 }
                     
                 if barcodes.isEmpty {
+                    DispatchQueue.main.async {
+                        result([
+                            "name": "barcode",
+                            "data": [],
+                        ])
+                    }
                     return
                 }
                     
-                result([
-                    "name": "barcode",
-                    "data": barcodes.map({ $0.toMap(imageWidth: Int(ciImage.extent.width), imageHeight: Int(ciImage.extent.height), scanWindow: self.scanWindow) }),
-                ])
+                DispatchQueue.main.async {
+                    result([
+                        "name": "barcode",
+                        "data": barcodes.map({ $0.toMap(imageWidth: Int(ciImage.extent.width), imageHeight: Int(ciImage.extent.height), scanWindow: self.scanWindow) }),
+                    ])
+                }
             })
             
             if !symbologies.isEmpty {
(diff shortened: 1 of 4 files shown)


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -665,17 +665,31 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                 }
                     
                 guard let barcodes: [VNBarcodeObservation] = request.results as? [VNBarcodeObservation] else {
+                    DispatchQueue.main.async {
```
navaronbracke, line 668: We apparently forgot these two :(

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -665,17 +665,31 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+                        ])
+                    }
                     return
                 }
                     
-                result([
-                    "name": "barcode",
-                    "data": barcodes.map({ $0.toMap(imageWidth: Int(ciImage.extent.width), imageHeight: Int(ciImage.extent.height), scanWindow: self.scanWindow) }),
-                ])
+                DispatchQueue.main.async {
```
navaronbracke, line 687: Adding a dispatch queue block, just in case (is the only relevant statement that didn't have it)

example/macos/Runner.xcodeproj/project.pbxproj
```diff
@@ -29,6 +29,7 @@
 		33CC11132044BFA00003C045 /* MainFlutterWindow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 33CC11122044BFA00003C045 /* MainFlutterWindow.swift */; };
 		58A22AC50A792ECA6D027507 /* Pods_RunnerTests.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 419C3BC9593F6DE903D740F0 /* Pods_RunnerTests.framework */; };
 		F209F1436A19CBC32BFFB26A /* Pods_Runner.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = EEB40B96A6FFC92607527710 /* Pods_Runner.framework */; };
+		78A318202AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage in Frameworks */ = {isa = PBXBuildFile; productRef = 78A3181F2AECB46A00862997 /* FlutterGeneratedPluginSwiftPackage */; };
```
navaronbracke, line 32: This is the same change for SPM as in https://github.com/juliansteenbakker/mobile_scanner/pull/1273 but for MacOS, which I forgot to add.

example/macos/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
```diff
@@ -5,6 +5,24 @@
    <BuildAction
       parallelizeBuildables = "YES"
       buildImplicitDependencies = "YES">
+      <PreActions>
```
navaronbracke, line 8: This is the same change for SPM as in https://github.com/juliansteenbakker/mobile_scanner/pull/1273 but for MacOS, which I forgot to add.

navaronbracke:
Working as intended now on macOS, and thus also on iOS, since both share the same code path for the fix.

Issue #1269, opened 2024-12-28
[iOS] Simulator should support analyzeImage


navaronbracke:
As reported in https://github.com/juliansteenbakker/mobile_scanner/issues/1225#issuecomment-2564341580
there is an error when trying to set up the detection model for the iOS Simulator.

We should tweak the model request for the iOS SImulator.

At first glance, we can use the `useCPUOnly` flag, but that flag was deprecated:
https://forums.developer.apple.com/forums/thread/696714
https://developer.apple.com/documentation/vision/vnrequest/usescpuonly

We can just no-op out of it, or figure out a workaround instead.

Perhaps setting `preferBackgroundProcessing` to false might help on the Simulator

navaronbracke:
I did evaluate if `preferBackgroundProcessing` had any effect, but that didn't help. Since `useCPUOnly` is deprecated in iOS 17, I opted into just returning an unsupported operation error instead.

PR #1275, opened 2025-01-06
fix: report analyze image as unsupported on iOS Simulator

navaronbracke:
It seems that the iOS Simulator does not support the neural engine features that are required to run the Vision API,
hence why it fails to create the underlying detection model.

Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/1269

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/MobileScannerErrorCodes.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/MobileScannerErrorCodes.kt
@@ -22,5 +22,6 @@ class MobileScannerErrorCodes {
         const val NO_CAMERA_ERROR_MESSAGE = "No cameras available."
         const val SET_SCALE_WHEN_STOPPED_ERROR = "MOBILE_SCANNER_SET_SCALE_WHEN_STOPPED_ERROR"
         const val SET_SCALE_WHEN_STOPPED_ERROR_MESSAGE = "The zoom scale cannot be changed when the camera is stopped."
+        const val UNSUPPORTED_OPERATION_ERROR = "MOBILE_SCANNER_UNSUPPORTED_OPERATION" // Reserved for future use.
     }
 }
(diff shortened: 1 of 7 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/objects/MobileScannerErrorCodes.kt
```diff
@@ -22,5 +22,6 @@ class MobileScannerErrorCodes {
         const val NO_CAMERA_ERROR_MESSAGE = "No cameras available."
         const val SET_SCALE_WHEN_STOPPED_ERROR = "MOBILE_SCANNER_SET_SCALE_WHEN_STOPPED_ERROR"
         const val SET_SCALE_WHEN_STOPPED_ERROR_MESSAGE = "The zoom scale cannot be changed when the camera is stopped."
+        const val UNSUPPORTED_OPERATION_ERROR = "MOBILE_SCANNER_UNSUPPORTED_OPERATION" // Reserved for future use.
```
navaronbracke, line 25: Added here too, for feature parity with the iOS / macOS implementation, in case we need it in the future.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -630,6 +630,17 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     }
     
     func analyzeImage(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
+        // The iOS Simulator cannot use some of the GPU features that are required for the Vision API.
+        // Thus analyzing images is not supported on the iOS Simulator.
+#if os(iOS) && targetEnvironment(simulator)
```
navaronbracke, line 635: It seems that the neural engine is not supported on the Simulator, see https://forums.developer.apple.com/forums/thread/696714

PR #1277, opened 2025-01-06
feat: Android Impeller support

### LINKED ISSUE #867: Support Impeller on Android

We should do the following:

            android:value="2" />
            <meta-data
            android:value="true" />`
            
navaronbracke:
TODO: test this change + add orientation correction if needed

This PR adds support for SurfaceProducer on Android, which enables support for Impeller on Android.

Fixes #867

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -18,6 +18,7 @@ import androidx.camera.core.ExperimentalGetImage
 import androidx.camera.core.ImageAnalysis
 import androidx.camera.core.ImageProxy
 import androidx.camera.core.Preview
+import androidx.camera.core.SurfaceRequest
 import androidx.camera.core.TorchState
 import androidx.camera.core.resolutionselector.ResolutionSelector
 import androidx.camera.core.resolutionselector.ResolutionStrategy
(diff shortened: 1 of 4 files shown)


navaronbracke:
@juliansteenbakker I will finish up the implementation and then look into getting someone from the Android team to perhaps peer review this change. This is mostly surface producer changes, by taking a look at the prior art from `camera_android_camerax`

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -165,6 +167,70 @@ class MobileScanner(
...
+
+                    when(resultCode) {
+                        SurfaceRequest.Result.RESULT_REQUEST_CANCELLED,
+                        SurfaceRequest.Result.RESULT_WILL_NOT_PROVIDE_SURFACE,
+                        SurfaceRequest.Result.RESULT_SURFACE_ALREADY_PROVIDED,
+                        SurfaceRequest.Result.RESULT_SURFACE_USED_SUCCESSFULLY -> {
+                            // Only need to release, do nothing.
+                        }
+                        SurfaceRequest.Result.RESULT_INVALID_SURFACE -> {
+                            // The surface was invalid, so it is not clear how to recover from this.
```
navaronbracke, line 223: I do not know what to do here, so I left a comment.

Issue #1184, opened 2024-09-20
The screen is still black for a while, even if isInitialized and isRunning are true


annaAzarova:
I want to show the loader until the device camera is opened. I listen to isInitialized and isRunning. When they are true, the screen is still black for about half a second. How can I track when the device camera actually opens?

navaronbracke:
This definitely sounds like a bug in the plugin. I'll have to take a look to see when the first camera frame is available, and adjust the implementation accordingly.

juliansteenbakker:
With the latest version we have 3 parameters on `MobileScannerState` that can be used to check this:

```dart
    required this.isInitialized,
    required this.isStarting,
    required this.isRunning,
```

PR #1283, opened 2025-01-09
feat: Android Impeller support

### LINKED ISSUE #867: Support Impeller on Android

navaronbracke:
This PR adds support for Impeller on Android.

Supercedes #1277 (retargets those changes onto the new develop branch)

TODO:
- [x] replace `SurfaceTexture` with `SurfaceProducer`
- [x] suppress deprecation for the to-be-deprecated `onSurfaceDestroyed` (I need to add the annotation)
- [x] add rotation correction, inspired by the change in camera_android_camerax
- [x] provide the rotation correction with the needed input arguments (also inspired by the change in camera_android_camerax)
   - [x] provide `isPreviewPreTransformed` (either from the surface producer or by checking the SDK level)
   - [x] provide the device's natural orientation (i.e. landscape for tablets)
   - [x] provide the sensor orientation in degrees
   - [x] set up a device orientation EventSink on the Android side, to relay changes in the device orientation
   - [x] listen to changes in the device orientation on the Dart side, so that we can correct the orientation of the preview
 - [ ] investigate additional rotation being applied due to Impeller. See https://github.com/juliansteenbakker/mobile_scanner/pull/1283#discussion_r1927798329

Part of #867

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -46,8 +46,8 @@ class MobileScanner(
     private val textureRegistry: TextureRegistry,
     private val mobileScannerCallback: MobileScannerCallback,
     private val mobileScannerErrorCallback: MobileScannerErrorCallback,
-    private val barcodeScannerFactory: (options: BarcodeScannerOptions?) -> BarcodeScanner = ::defaultBarcodeScannerFactory,
     private val deviceOrientationListener: DeviceOrientationListener,
+    private val barcodeScannerFactory: (options: BarcodeScannerOptions?) -> BarcodeScanner = ::defaultBarcodeScannerFactory,
 ) {
 
     /// Internal variables
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -299,13 +299,16 @@ class MobileScanner(
 //            if (isPaused) {
 //                resumeCamera()
 //                mobileScannerStartedCallback(
-//                    MobileScannerStartParameters(
-//                        if (portrait) width else height,
-//                        if (portrait) height else width,
-//                        currentTorchState,
-//                        surfaceProducer!!.id(),
-//                        numberOfCameras ?: 0
-//                    )
+//                  MobileScannerStartParameters(
+//                    if (portrait) width else height,
+//                    if (portrait) height else width,
+//                    deviceOrientationListener.getUIOrientation().serialize(),
+//                    sensorRotationDegrees,
+//                    surfaceProducer!!.handlesCropAndRotation(),
+//                    currentTorchState,
+//                    surfaceProducer!!.id(),
+//                    numberOfCameras ?: 0
+//                  )
 //                )
 //                return
 //            }
@@ -415,7 +418,8 @@ class MobileScanner(
             val resolution = analysis.resolutionInfo!!.resolution
             val width = resolution.width.toDouble()
             val height = resolution.height.toDouble()
-            val portrait = (camera?.cameraInfo?.sensorRotationDegrees ?: 0) % 180 == 0
+            val sensorRotationDegrees = camera?.cameraInfo?.sensorRotationDegrees ?: 0
+            val portrait = sensorRotationDegrees % 180 == 0
 
             // Start with 'unavailable' torch state.
             var currentTorchState: Int = -1
@@ -434,6 +438,9 @@ class MobileScanner(
                 MobileScannerStartParameters(
                     if (portrait) width else height,
                     if (portrait) height else width,
+                    deviceOrientationListener.getUIOrientation().serialize(),
+                    sensorRotationDegrees,
+                    surfaceProducer!!.handlesCropAndRotation(),
                     currentTorchState,
                     surfaceProducer!!.id(),
                     numberOfCameras ?: 0
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -34,6 +34,7 @@ import dev.steenbakker.mobile_scanner.objects.DetectionSpeed
 import dev.steenbakker.mobile_scanner.objects.MobileScannerErrorCodes
 import dev.steenbakker.mobile_scanner.objects.MobileScannerStartParameters
 import dev.steenbakker.mobile_scanner.utils.YuvToRgbConverter
+import dev.steenbakker.mobile_scanner.utils.serialize
 import io.flutter.view.TextureRegistry
 import java.io.ByteArrayOutputStream
 import java.io.IOException
@@ -46,6 +47,7 @@ class MobileScanner(
     private val mobileScannerCallback: MobileScannerCallback,
     private val mobileScannerErrorCallback: MobileScannerErrorCallback,
     private val barcodeScannerFactory: (options: BarcodeScannerOptions?) -> BarcodeScanner = ::defaultBarcodeScannerFactory,
+    private val deviceOrientationListener: DeviceOrientationListener,
 ) {
 
     /// Internal variables
@@ -426,6 +428,8 @@ class MobileScanner(
                 currentTorchState = it.torchState.value ?: -1
             }
 
+            deviceOrientationListener.start()
+
             mobileScannerStartedCallback(
                 MobileScannerStartParameters(
                     if (portrait) width else height,
@@ -449,6 +453,7 @@ class MobileScanner(
             throw AlreadyStopped()
         }
 
+        deviceOrientationListener.stop()
         pauseCamera()
     }
 
@@ -460,6 +465,7 @@ class MobileScanner(
             throw AlreadyStopped()
         }
 
+        deviceOrientationListener.stop()
         releaseCamera()
     }
 
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -18,6 +18,7 @@ import androidx.camera.core.ExperimentalGetImage
 import androidx.camera.core.ImageAnalysis
 import androidx.camera.core.ImageProxy
 import androidx.camera.core.Preview
+import androidx.camera.core.SurfaceRequest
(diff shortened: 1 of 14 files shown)


pubspec.yaml
```diff
@@ -16,8 +16,8 @@ screenshots:
   path: example/screenshots/overlay.png
 
 environment:
-  sdk: ">=3.4.0 <4.0.0"
-  flutter: ">=3.22.0"
+  sdk: ">=3.6.0 <4.0.0"
```
navaronbracke, line 19: The SurfaceProducer API has seen some changes over a long period of time, due to the lengthy release cadence of the framework itself.

The SurfaceProducer API was available, but non-functional in Flutter 3.22, with the migration doc for plugin authors listing Flutter 3.24 as the minimum required version for the migration. However, in Flutter 3.27 both Impeller on Android has been turned on by default, and there was a name change for one of the two callback in the SurfaceProducer usage. (onSurfaceCreated -> onSurfaceAvailable)

In Flutter 3.28 there will be another name change (onSurfaceDestroyed -> onSurfaceCleanup), but since the current stable is only version 3.27, I added a TODO for that change

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -165,6 +167,70 @@ class MobileScanner(
...
+
+                // The single thread executor is only used to invoke the result callback.
+                // Thus it is safe to use a new executor,
+                // instead of reusing the executor that is passed to the camera process provider.
+                request.provideSurface(surface, Executors.newSingleThreadExecutor()) {
+                    // Handle the result of the request for a surface.
+                    // See: https://developer.android.com/reference/androidx/camera/core/SurfaceRequest.Result
+
+                    // Always attempt a release.
+                    surface.release()
```
navaronbracke, line 211: I think that adding this line fixes any occurrences of A resource failed to call close that we sometimes get.
Previously, we never did anything with the result callback, thus also not releasing the surface.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -165,6 +167,70 @@ class MobileScanner(
...
+
+                    when(resultCode) {
+                        SurfaceRequest.Result.RESULT_REQUEST_CANCELLED,
+                        SurfaceRequest.Result.RESULT_WILL_NOT_PROVIDE_SURFACE,
+                        SurfaceRequest.Result.RESULT_SURFACE_ALREADY_PROVIDED,
+                        SurfaceRequest.Result.RESULT_SURFACE_USED_SUCCESSFULLY -> {
+                            // Only need to release, do nothing.
+                        }
+                        SurfaceRequest.Result.RESULT_INVALID_SURFACE -> {
+                            // The surface was invalid, so it is not clear how to recover from this.
```
navaronbracke, line 223: I think that adding this line fixes any occurrences of A resource failed to call close that we sometimes get.
Previously, we never did anything with the result callback, thus also not releasing the surface.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -246,23 +312,11 @@ class MobileScanner(
             }
 
             cameraProvider?.unbindAll()
-            textureEntry = textureRegistry.createSurfaceTexture()
 
-            // Preview
-            val surfaceProvider = Preview.SurfaceProvider { request ->
-                if (isStopped()) {
```
navaronbracke, line 253: We no longer have this check, I'm not sure if we should put it back or not.

navaronbracke:
@juliansteenbakker I moved over the changes from the old PR into this one.

One thing that I don't know for sure is if we should provide Impeller support on Android in Flutter 3.24 or 3.27.
The current stable is version 3.27, but I think there were fixes for things in this API, between those two stable releases.
See also https://github.com/juliansteenbakker/mobile_scanner/pull/1283#discussion_r1908503362

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -53,6 +53,67 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
 
   int? _textureId;
 
+  /// Rotate the given [child] to account for the device orientation.
+  Widget _buildCameraViewWithDeviceOrientation(Widget child) {
+    final Map<DeviceOrientation, int> degreesForDeviceOrientation =
```
navaronbracke, line 58: This function is still a work in progress, as we need to retrieve some arguments (camera direction, sensor orientation, natural device orientation) and observe the device orientation changes.

navaronbracke:
@juliansteenbakker I see that the pause functionality landed on master. We probably need to rebase develop to pull in those changes. Also, I'll rebase this PR to avoid the merge conflict on the Android implementation side.

Overall the new pause feature looks nice to finally have.

juliansteenbakker:
@navaronbracke i am reviewing all open pr's right now, and once done i'll rebase develop so it has all extra features merged in master

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -263,23 +329,10 @@ class MobileScanner(
             }
 
             cameraProvider?.unbindAll()
-            textureEntry = textureEntry ?: textureRegistry.createSurfaceTexture()
+            surfaceProducer = surfaceProducer ?: textureRegistry.createSurfaceProducer()
```
navaronbracke, line 332: @juliansteenbakker The `?:` expression was introduced with the pause feature. Do you recall why we need it?
I am mostly migrating from `TextureEntry` to `SurfaceProducer`, so I'd rather not change anything else.

navaronbracke:
@juliansteenbakker This has now been rebased onto dev. I will look into the remaining two points tomorrow or over the weekend :)

navaronbracke:
@juliansteenbakker This should now be feature complete, but I still need to test the change.

lib/src/method_channel/android_surface_producer_delegate.dart
```diff
@@ -0,0 +1,165 @@
...
+  ///
+  /// The camera preview will use this orientation as the natural orientation
+  /// to correct its rotation with respect to, if necessary.
+  final DeviceOrientation naturalOrientation;
+
+  /// The sensor orientation of the current camera, in degrees.
+  final int sensorOrientation;
+
+  /// Apply a rotation correction to the given [texture] widget.
+  Widget applyRotationCorrection(Widget texture) {
```
camsim99, line 92: Hi Navaron! I see now you mean you followed what I wrote in https://docs.flutter.dev/release/breaking-changes/android-surface-plugins#note-on-camera-previews. Unfortunately, though this did work for some time, the correctness of this has changed I believe due to changes in Impeller and that's what I'm working through in https://github.com/flutter/flutter/issues/154241. I removed all of this logic in Flutter recently to better debug (and also because I saw the correct behavior without correction in my testing, though it's still broken for many folks).

Because of that, I don't have any advice for you at the moment :/ I'm happy to give your more details and key insights on https://github.com/flutter/flutter/issues/154241 if you're open to helping debug it via Flutter or mobile_scanner, otherwise I can update you as I learn more. Planning to work on it with Impeller folks in the coming week or two.

navaronbracke: Thanks for the update! And yeah, this not working due to changes in Impeller was something that I was thinking of as well. The implementation of this PR is almost identical to the one for `camera_android_camerax`, so I wasn't expecting the behavior I get currently. I'll follow up in https://github.com/flutter/flutter/issues/154241

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
juliansteenbakker: The ?: was added because otherwise a new texture was created instead of the existing one being used after resuming from pause.

lib/src/method_channel/android_surface_producer_delegate.dart
juliansteenbakker: @navaronbracke should we wait until a fix is landed before merging this PR?

navaronbracke: We could also do the SurfaceProducer change and do the orientation correction in a follow-up. But then, as Camille mentioned, there could be bugs relating to the rotation of the preview.

I'll leave it up to you to decide and I'll adjust this PR accordingly.

navaronbracke: Sounds good!

navaronbracke: @juliansteenbakker I turned off the rotation correction (but left it in the code) in https://github.com/juliansteenbakker/mobile_scanner/pull/1283/commits/a7f6b99b623a8ce91c04476a90a9e27ac6fcadd9

Then we at least can provide the current "worst case" scenario in a follow-up beta: The Impeller preview works (no black screens or artifacts), but might be rotated incorrectly on non-portrait orientations. I'll try to look into the orientation independently of Camille, but if anyone finds the culprit, feel free to let me know.

navaronbracke:
@juliansteenbakker I can assume we will revisit the rotation correction later, since this was merged into dev?

juliansteenbakker:
Yeah i think it was safe to merge, because the current beta release also suffers from rotational issues, so we aren't introducing a new bug here.

navaronbracke:
Yeah, I had the same thought about it!

juliansteenbakker:
This beta has _rotationCorrectionEnabled disabled. I will release a new beta with _rotationCorrectionEnabled enabled, which in some cases fixes this issue.

JgomesAT:
OK, I wait for this new beta

JgomesAT:
I have another strange behavior. Randomly, the first time I open the camera, I see a black screen. I had the same problem before supporting the impeller, when the impeller was enabled, but if I close and reopen the camera, it seems to work the second time.

navaronbracke:
@JgomesAT as mentioned above, the rotation is only partially fixed. See also this comment for more info: https://github.com/juliansteenbakker/mobile_scanner/pull/1283#discussion_r1927798329

As for the black screen issue, is this the same as https://github.com/juliansteenbakker/mobile_scanner/issues/1184 or an entirely different issue?

JgomesAT:
I don't know, I try to do more research, but i sees that problem

JgomesAT:
I made some cahnge in my code and now teh black screen doesn't appear

old code:  await cameraController.start();
new code: unawaited(controller.start());

PR #1295, opened 2025-01-16
fix: remove redundant release

navaronbracke:
This PR is a follow up to https://github.com/navaronbracke/mobile_scanner/commit/dead58edf18a6c4e8417fcc46885f690a00b6335

The second release is redundant since we already do it a few lines above.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -478,8 +478,6 @@ class MobileScanner(
             it.zoomState.removeObservers(owner)
             it.cameraState.removeObservers(owner)
         }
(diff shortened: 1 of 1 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -478,8 +478,6 @@ class MobileScanner(
             it.zoomState.removeObservers(owner)
             it.cameraState.removeObservers(owner)
         }
-        textureEntry?.release()
```
navaronbracke, line 481: @juliansteenbakker I decided to remove this one, since then the CameraX code (`camera?.cameraInfo?.let` and `cameraProvider?.unbindAll()`) are grouped together

juliansteenbakker (approved):

PR #1303, opened 2025-01-23
fix: Fix the issue of incorrect QR code position detection when Boxfit is …

ohnobug:
my qrcode scan code:

```
  @override
    // 二维码的位置
    final overlays = <Widget>[
      if (_barcodeCapture != null && _barcodeCapture!.barcodes.isNotEmpty)
        for (final Barcode barcode in _barcodeCapture!.barcodes)
          if (!barcode.size.isEmpty && barcode.corners.isNotEmpty)
            CustomPaint(
              painter: BarcodePainter2(
                barcodeCorners: barcode.corners,
                barcodeSize: barcode.size,
                boxFit: BoxFit.cover,
                cameraPreviewSize: _barcodeCapture!.size,
                color: const Color.fromARGB(183, 0, 255, 34),
                style: PaintingStyle.fill,
              ),
            ),
    ];

    return BlocBuilder<SystemCubit, SystemState>(
        builder: (context, systemState) {
      return Scaffold(
          primary: false,
          appBar: null,
          body: Stack(
            fit: StackFit.expand,
            children: [
              MobileScanner(
                fit: BoxFit.cover,
                controller: controller,
                // onDetect: _handleBarcode,
              ),

              if (_barcodeCapture != null && _barcodeCapture!.image != null)
                Stack(fit: StackFit.expand, children: [
                  // 扫码后暂停结果
                  Image.memory(
                    _barcodeCapture!.image!,
                    fit: BoxFit.cover,
                    filterQuality: FilterQuality.low,
                    frameBuilder: (
                      BuildContext context,
                      Widget child,
                      int? frame,
                      bool? wasSynchronouslyLoaded,
                    ) {
                      if (wasSynchronouslyLoaded == true || frame != null) {
                        return child;
                      }

                      return SizedBox();
                    },
                  ),
                  ...overlays,
                ]),

```

before:
3db30c82481a125934ffc29ecc2eaf2

fixed after:
5f1fa23a7a5791334c0f760070f3b48


lib/src/overlay/barcode_painter.dart
```diff
@@ -40,33 +41,31 @@ class BarcodePainter extends CustomPainter {
       return;
     }
 
-    final adjustedSize = applyBoxFit(boxFit, cameraPreviewSize, size);
+    ScalingRatios ratio = calculateBoxFitRatio(boxFit, cameraPreviewSize, size);
```
navaronbracke, line 44: Can `ratio` be final?

navaronbracke (changes requested):

ohnobug:
fix
https://github.com/user-attachments/assets/3123a5ac-c467-4617-8458-0fd4f997e3f2

example/lib/barcode_scanner_window.dart
```diff
@@ -19,46 +19,165 @@ class _BarcodeScannerWithScanWindowState
...
+                    ),
+                    Positioned(
+                      bottom: 100,
+                      width: containerWidth,
+                      child: Wrap(
+                        spacing: 0.5,
+                        runSpacing: 3.0,
+                        alignment: WrapAlignment.center,
+                        children: <Widget>[
+                          SizedBox(
```
navaronbracke, line 102: Is a SizedBox needed around each button?

example/lib/barcode_scanner_window.dart
```diff
@@ -19,46 +19,165 @@ class _BarcodeScannerWithScanWindowState
...
+                              height: 30,
+                              child: ElevatedButton(
```
navaronbracke, line 104: Perhaps we should have a helper function that creates the buttons? We can also use the enum name for the label?

```
Widget _buildBoxFitButton(BoxFit fit) {
  return ElevatedButton(
    onPressed: () => setState(() {
      boxFit = fit;
    }),
    child: Text(fit.name)
  );
}
```

navaronbracke (changes requested):

example/lib/barcode_scanner_window.dart
```diff
@@ -19,46 +19,165 @@ class _BarcodeScannerWithScanWindowState
...
-              height: 100,
-              color: const Color.fromRGBO(0, 0, 0, 0.4),
-              child: ScannedBarcodeLabel(barcodes: controller.barcodes),
+      body: Container(
+        height: MediaQuery.sizeOf(context).height,
+        width: MediaQuery.sizeOf(context).width,
+        color: Colors.blueGrey,
+        child: Stack(
+          children: [
+            const Text(
```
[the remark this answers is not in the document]
ohnobug: I want to remind users that the camera lens can be placed in a box of any size, not necessarily a box the size of the screen. So I'm creating a background and a foreground text here tips to user.

example/lib/barcode_scanner_window.dart
ohnobug: Yes, otherwise the buttons would be very large.

example/lib/barcode_scanner_window.dart
ohnobug: fixed

ohnobug:
example video：
Uploading 2c8ce1fa535c09094d1a7825f29d72e5.mp4…

ohnobug:
example test video:
https://github.com/user-attachments/assets/a7dbdb13-e212-4f35-a78d-9acb343cc3d6

example/lib/barcode_scanner_window.dart
```diff
@@ -19,46 +19,165 @@ class _BarcodeScannerWithScanWindowState
...
+                        alignment: Alignment.center,
+                        padding: const EdgeInsets.symmetric(
+                            horizontal: 16, vertical: 8),
+                        height: 100,
+                        color: const Color.fromRGBO(0, 0, 0, 0.4),
+                        child:
+                            ScannedBarcodeLabel(barcodes: controller.barcodes),
+                      ),
+                    ),
+                    const Text(
```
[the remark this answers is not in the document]
ohnobug: added！
image

lib/src/overlay/barcode_painter.dart
```diff
@@ -90,3 +77,57 @@ class BarcodePainter extends CustomPainter {
         oldDelegate.style != style;
   }
 }
+
+/// Calculate the scaling ratios for width and height to fit the small box (cameraPreviewSize)
+/// into the large box (size) based on the specified BoxFit mode.
+/// Returns a record containing the width and height scaling ratios.
+({double widthRatio, double heightRatio}) calculateBoxFitRatio(
```
navaronbracke, line 84: Can we add unit tests for this function? Then we can assure that the math is right.

example/lib/barcode_scanner_window.dart
```diff
@@ -19,46 +19,114 @@ class _BarcodeScannerWithScanWindowState
...
+                      ),
+                    ),
+                    Positioned(
+                      bottom: 100,
+                      width: containerWidth,
+                      child: Wrap(
+                        spacing: 0.5,
+                        runSpacing: 3.0,
+                        alignment: WrapAlignment.center,
+                        children: BoxFit.values.map((fit) {
```
navaronbracke, line 107: You can use a tear off here:

```suggestion
                        children: BoxFit.values.map(_buildBoxFitButton).toList(),
```

example/lib/barcode_scanner_window.dart
```diff
@@ -19,46 +19,165 @@ class _BarcodeScannerWithScanWindowState
...
+                    Align(
+                      alignment: Alignment.bottomCenter,
+                      child: Container(
+                        alignment: Alignment.center,
+                        padding: const EdgeInsets.symmetric(
+                            horizontal: 16, vertical: 8),
+                        height: 100,
+                        color: const Color.fromRGBO(0, 0, 0, 0.4),
+                        child:
+                            ScannedBarcodeLabel(barcodes: controller.barcodes),
```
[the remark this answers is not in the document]
navaronbracke: This is not resolved yet, there is no comma after `.barcodes`.

example/lib/barcode_scanner_window.dart
```diff
@@ -19,46 +19,114 @@ class _BarcodeScannerWithScanWindowState
   );
 
   // TODO: Fix BoxFit.fill & BoxFit.fitHeight
```
navaronbracke, line 21: Does this PR fix this TODO? If yes, then you can remove the TODO.

navaronbracke (changes requested):

test/scan_window_test.dart
```diff
@@ -152,6 +154,191 @@ void main() {
       });
     },
   );
+
+  group(
+    'Calculate Box Fit Ratio',
```
navaronbracke, line 159: Can we use the name of the function under test as the topmost group name?

```suggestion
    'calculateBoxFitRatio',
```

test/scan_window_test.dart
```diff
@@ -152,6 +154,191 @@ void main() {
...
+    () {
+      group('Smaller camera preview size in portrait', () {
+        const cameraPreviewSize = Size(480.0, 640.0);
+        const size = Size(432.0, 256.0);
+
+        test('scp p: BoxFit.fill', () {
```
navaronbracke, line 165: What is the meaning of the `scp p:` in the labels?

test/scan_window_test.dart
```diff
@@ -152,6 +154,191 @@ void main() {
...
+          final ratio = calculateBoxFitRatio(
+            BoxFit.fill,
+            cameraPreviewSize,
+            size,
+          );
+          expect(ratio.widthRatio, size.width / cameraPreviewSize.width);
```
navaronbracke, line 171: For each test, I would use the actual numbers of the calculation as the expected value. (the second argument in the `expect()`) That `calculateBoxFitRatio` does `size.width / cameraPreviewSize.width` internally is an implementation detail that should not be exposed in the tests.

navaronbracke (changes requested):

test/scan_window_test.dart
ohnobug: I don't know what the naming convention for test functions is, I just saw that the previous tests used abbreviations.

navaronbracke: I don't know the reasoning behind the existing abbreviations either. I would rather avoid introducing abbreviations altogether. It makes the code harder to reason about.

test/scan_window_test.dart
```diff
@@ -152,6 +154,167 @@ void main() {
...
+        });
+
+        test('scpsip: BoxFit.contain', () {
+          final ratio = calculateBoxFitRatio(
+            BoxFit.contain,
+            cameraPreviewSize,
+            size,
+          );
+
+          expect(ratio.widthRatio, 0.4);
```
navaronbracke, line 182: Can we use several different sizes, to have the numbers be different for each test?

I'd like to avoid introducing a bug because one test group accidentally had `0.9 / 0.9` several times in tests within the group. If we use different sizes, then the outcomes will also be unique among tests in a test group.

When you define the extra sizes, you can reuse them for landscape as well (since that is just flipping the width/height)

test/scan_window_test.dart
```diff
@@ -152,6 +154,167 @@ void main() {
...
   );
+
+  group(
+    'calculateBoxFitRatio',
+    () {
+      group('Smaller camera preview size in portrait', () {
+        const cameraPreviewSize = Size(480.0, 640.0);
+        const size = Size(432.0, 256.0);
+
+        test('scpsip: BoxFit.fill', () {
```
navaronbracke, line 165: Maybe you can change the inner test group name to already mention portrait / landscape? Then that does not need to be repeated in the tests themselves and the abbreviations can be removed?

```
group('calculateBoxFitRatio', () {
  group('smaller portrait camera preview size', () {
    test('fit: BoxFit.fill', () { /* ... */ });
    
    // more tests
  });
  
  group('smaller landscape camera preview size', () {
    test('fit: BoxFit.fill', () { /* ... */ });
    
    // more tests  
  });
})
```

navaronbracke (changes requested):
Just a comment about fixing the test names & a small concern about the test results being duplicated.

test/scan_window_test.dart
juliansteenbakker: I have updated all test names since most of them weren't explanatory.

test/scan_window_test.dart
juliansteenbakker: I have added boxfit check to multiple groups, so i think this is now covered.

juliansteenbakker:
I have updated the existing tests to be more inclusive. The missing coverage that @codecov-commenter is reporting about will be fixed in a new PR, which will also add some mocking tools.

Issue #1042, opened 2024-04-25
MethodChannelMobileScanner.eventsStream never gets destroyed, which makes it really difficult to write tests for the plugin


mbtodorov:
With the latest release of the plugin, the event stream never gets cancelled (even after the controller is stopped) https://github.com/juliansteenbakker/mobile_scanner/blob/master/lib/src/method_channel/mobile_scanner_method_channel.dart#L33

This is not a problem for running the app in production, but if you are writing tests it can be cumbersome, because it means that the `listen` method is called only once per  `_test.dart` file. Let me try to show an example. Assume the following test setup:

```dart
void main() {
  const methodChannel =
      MethodChannel('dev.steenbakker.mobile_scanner/scanner/method');
  const eventChannel =
      EventChannel('dev.steenbakker.mobile_scanner/scanner/event');

  setUp(() {
    TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
      ..setMockMethodCallHandler(
        methodChannel,
        (message) async {
          print('method channel method called: ${message.method}');
          final methodName = message.method;
          if (methodName == 'start') {
            return {
              'size': {
                'width': 100.0,
                'height': 100.0,
              },
              'textureId': 1,
            };
          }
          if (methodName == 'state') {
            return MobileScannerAuthorizationState.authorized.rawValue;
          }
          return null;
        },
      )
      ..setMockMethodCallHandler(
        MethodChannel(eventChannel.name, eventChannel.codec),
        (message) {
          print('event channel method called: ${message.method}');
          return null;
        },
      );
  });

  tearDown(() {
    TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
      ..setMockMethodCallHandler(methodChannel, null)
      ..setMockStreamHandler(eventChannel, null);
  });

  testWidgets('test one', (tester) async {
    await tester.pumpWidget(SomeWidget()); // assume this widget renders MobileScanner and listens to barcodes stream
  });

  testWidgets('test two', (tester) async {
    await tester.pumpWidget(SomeWidget()); // assume this widget renders MobileScanner and listens to barcodes stream
  });
}
```

When running these tests, this is what will get printed:

```
// test one:
event channel method called: listen
method channel method called: state
method channel method called: start

// test two:
method channel method called: state
method channel method called: start
```

[...]

PR #1312, opened 2025-01-31
fix: export `StartOptions` and `MobileScannerViewAttributes` for tests

w3ggy:
I added `StartOptions` and `MobileScannerViewAttributes` to export to let the developers to mock the `MobileScannerPlatform` in tests.

I tried to write golden tests and mock the `MobileScannerPlatform` but I cannot do it because I cannot import classes which are necessary to override `Future<MobileScannerViewAttributes> start(StartOptions startOptions)`


navaronbracke:
@w3ggy Should be mark this as `@visibleForTesting` ? Otherwise, I think this looks good to have.
While there aren't many people that write tests for mobile_scanner usage, there are definitely _some_.

See also https://github.com/juliansteenbakker/mobile_scanner/issues/1042 which is also related to tests and we should look into.

navaronbracke:
@w3ggy I went ahead and verified that we do not need `@visibleForTesting` here, since that allows it to be used only in tests _and the defining library_. The latter is not the case, as the platform interface definition uses it.

I updated your pull request with a changelog entry and fixed the import order, per the https://dart.dev/tools/linter-rules/directives_ordering rule.

Thanks for catching this!

navaronbracke (approved):

PR #1314, opened 2025-02-03
chore: fix analysis on master

navaronbracke:
This PR fixes the current lint warnings/errors on tip-of-tree for `master`:

- `withOpacity` is deprecated, replaced by `fromRGBO`, except for one usage of `Colors.red`, which is replaced by a hexadecimal version of said color, with the opacity applied
- an unreachable `break` statement was reported by the new version of the Dart analyzer (the respective lint now detects this case).
- `onPopWithResult` is not available for the current min sdk version, so that was adjusted from Flutter 3.22 to 3.24

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -36,7 +36,7 @@ jobs:
       - uses: subosito/flutter-action@v2.12.0
         with:
           cache: true
-          flutter-version: '3.22'
+          flutter-version: '3.24'
           channel: 'stable'
       - name: Version
         run: flutter doctor -v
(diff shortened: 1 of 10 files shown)


example/lib/barcode_scanner_window.dart
```diff
@@ -214,7 +214,7 @@ class BarcodeOverlay extends CustomPainter {
     final cutoutPath = Path()..addPolygon(adjustedOffset, true);
 
     final backgroundPaint = Paint()
-      ..color = Colors.red.withOpacity(0.3)
+      ..color = const Color(0x4DF44336)
```
navaronbracke, line 217: `FFF44336` is the original red value from `Colors.red`

.github/workflows/flutter.yml
```diff
@@ -36,7 +36,7 @@ jobs:
       - uses: subosito/flutter-action@v2.12.0
         with:
           cache: true
-          flutter-version: '3.22'
+          flutter-version: '3.24'
```
navaronbracke, line 39: The `onPopWithResult` callback is new in Flutter 3.24

PR #1316, opened 2025-02-04
fix: add payload data as raw bytes in the new iOS/MacOS implementation 

### LINKED ISSUE #1315: Darwin: 7.0.0-beta: Support for QR with byte array payload
Pardon the sloppy ticket. I hope it has enough information regardless.

Version 6.x works fine with byte array payloads. The 7.x beta, however, does not.

Looking at https://github.com/juliansteenbakker/mobile_scanner/blob/develop/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift#L904, I see `rawValue` and `displayValue` both set to `payloadStringValue`.

In the case of a QR code with a byte array payload, the `payloadStringValue` will be an empty string and `payloadData` will contain the actual data I'm expecting to see. In version 6.x, the latter is available as `rawBytes` on the Flutter side of things. In version 7.x beta it is not.

I'd love to see this fixed. I _could_ work on a fix and send it your way, but I'd like to know if this is something you have in the pipeline already.

Thanks.

navaronbracke:
This PR restores support for `Barcode.rawBytes` in the Vision implementation.
This appears to be the only field that wasn't ported over, excluding the various decoded datatypes (which are MLKIt specific and might be added later)

Fixes #1315

### DIFF
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -891,6 +891,13 @@ extension VNBarcodeObservation {
         // Calculate the width and height of the barcode based on adjusted coordinates
         let width = distanceBetween(adjustedTopLeft, adjustedTopRight) * CGFloat(imageWidth)
         let height = distanceBetween(adjustedTopLeft, adjustedBottomLeft) * CGFloat(imageHeight)
+        var rawBytes: FlutterStandardTypedData? = nil
+        
+        if #available(iOS 17.0, macOS 14.0, *) {
+            if let payloadData = payloadData {
+                rawBytes = FlutterStandardTypedData(bytes: payloadData)
+            }
+        }
 
         let data = [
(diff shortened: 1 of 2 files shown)


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -901,6 +908,7 @@ extension VNBarcodeObservation {
                 ["x": bottomLeftX, "y": bottomLeftY],
             ],
             "format": symbology.toInt ?? -1,
+            "rawBytes": rawBytes,
```
juliansteenbakker, line 911: We are providing default values in case of nil on other parameters. Should we do the same here, or do you think nil is no problem in this case?

navaronbracke: We already provide `nil` as the default on line 894, since that also applies for iOS / MacOS versions that cannot use the `payloadData` field.

PR #1325, opened 2025-02-13
fix: turn off camerax info logs

### LINKED ISSUE #1255: How to disable/decrease noisy logs

navaronbracke:
This PR turns off the CameraX logging for the `Log.DEBUG`, `Log.INFO` and `Log.WARN` logging.
Per the docs of CameraX:

```
For apps that want to reduce the logs produced by CameraX, set it to Log. ERROR to avoid all logs except for error.
```

We could technically also allow `Log.WARN`, but I'm not sure if that is going to help much.

Fixes #1255

### DIFF
--- a/.github/workflows/flutter.yml
+++ b/.github/workflows/flutter.yml
@@ -55,5 +55,8 @@ jobs:
       - uses: subosito/flutter-action@v2
         with:
           cache: true
+          # TODO: move the code to the new tall style once Flutter 3.29 is the new minimum.
+          # https://github.com/juliansteenbakker/mobile_scanner/issues/1326
+          flutter-version: '3.27'
       - name: Format
         run: dart format --set-exit-if-changed .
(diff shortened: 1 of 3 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -42,7 +45,6 @@ import dev.steenbakker.mobile_scanner.utils.serialize
 import io.flutter.view.TextureRegistry
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.coroutineScope
```
navaronbracke, line 45: Unused import

.github/workflows/flutter.yml
```diff
@@ -55,5 +55,8 @@ jobs:
       - uses: subosito/flutter-action@v2
         with:
           cache: true
+          # TODO: move the code to the new tall style once Flutter 3.29 is the new minimum.
```
navaronbracke, line 58: The formatting action is using the latest stable Flutter version, 3.29.0.
That opts us into Dart 3.7 which includes new tall-style formatting.

Issue #1319, opened 2025-02-05
7.0.0-beta.6: Incorrect camera preview orientation


disptr:
Using 7.0.0-beta.6 (and Flutter 3.27.0), in my limited testing, the camera preview orientation is incorrect.

The below table lists the camera preview rotation, in degrees, for the different device orientations:
| Device | OS | Portrait Up | Portrait Upside Down | Landscape Left | Landscape Right |
| - | - | - | - | - | - |
| iPad (8th generation) | iPadOS 18.0.1 | +90 | -90 | ±180 | 0 |
| Pixel 6a | Android 15 | -90 | N/A | 0 | ±180 |

The definition of landscape left vs right seems fuzzy. My definitions for the above table are as follows:
- **Landscape Left**: device rotated left from portrait up (e.g. home button on the right)
- **Landscape Right**: device rotated right from portrait up (e.g. home button on the left)

I would expect the rotation to be 0 in all cases, i.e. that the plugin handles this and automatically rotates the camera preview.

I am aware of https://github.com/juliansteenbakker/mobile_scanner/discussions/17.

Will this be resolved with 7.0.0?

navaronbracke:
Thanks for investigating this with a more detailed table.

This is a known issue with the Impeller guidance for camera previews and still needs to be fixed before 7.0.0 will be considered stable.

See https://github.com/juliansteenbakker/mobile_scanner/pull/1283#discussion_r1927798329

@camsim99 This table might be handy to investigate the issue on the Impeller side?

navaronbracke:
There is a potential solution to this problem, mentioned in https://github.com/flutter/flutter/issues/154241#issuecomment-2655098620

I will look into implementing that. Should be easy enough to do that.

navaronbracke:
@juliansteenbakker I have suspicions that the stale facing mode could be related to the orientation bug.
I prepared https://github.com/juliansteenbakker/mobile_scanner/pull/1328 to fix the stale camera direction.

Next, I'm going to look into what Camille said in https://github.com/flutter/flutter/issues/154241#issuecomment-2655098620
We already use `handlesCropAndRotation`, but the stream builder might still be needed.
I'm also going to retest on Flutter 3.29.0 to see if there are changes.

navaronbracke:
Flutter 3.29.0 has the same issue it seems. The preview is still rotated in the wrong direction.

PR #1328, opened 2025-02-13
feat: handle more facing directions

navaronbracke:
This PR adds the remaining facing modes to the `CameraFacing` enum.
It also passes the actual `CameraFacing` value from the camera to the `MobileScannerState` / surface producer delegate, instead of the requested facing direction. That _should_ help prevent a wrong sign for the orientation correction.

Part of https://github.com/juliansteenbakker/mobile_scanner/issues/867
Work towards fixing https://github.com/juliansteenbakker/mobile_scanner/issues/1319

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 Improvements:
 * [Android] Turn off logging for CameraX, except for the `Log.ERROR` logging level.
+* Added `CameraFacing.external` and `CameraFacing.unknown` enum values.
 
 ## 7.0.0-beta.6
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## NEXT
 
+**BREAKING CHANGES:**
+
+* The initial state of the `MobileScannerState` camera facing direction is changed to `CameraFacing.unknown`.
+
 Improvements:
 * [Android] Turn off logging for CameraX, except for the `Log.ERROR` logging level.
 * Added `CameraFacing.external` and `CameraFacing.unknown` enum values.
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -434,10 +434,18 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                 let answer: [String : Any?]
 
                 if let device = self.device {
+                    let cameraDirection: Int? = switch(device.position) {
+                        case .back: 1
+                        case .unspecified: nil
+                        case .front: 0
+                        @unknown default: nil
+                    }
+                    
                     answer = [
                         "textureId": self.textureId,
                         "size": size,
                         "currentTorchState": device.hasTorch ? device.torchMode.rawValue : -1,
+                        "cameraDirection": cameraDirection,
                     ]
                 } else {
                     answer = [
--- a/lib/src/method_channel/mobile_scanner_method_channel.dart
+++ b/lib/src/method_channel/mobile_scanner_method_channel.dart
@@ -4,6 +4,7 @@ import 'package:flutter/foundation.dart';
 import 'package:flutter/services.dart';
 import 'package:flutter/widgets.dart';
 import 'package:mobile_scanner/src/enums/barcode_format.dart';
+import 'package:mobile_scanner/src/enums/camera_facing.dart';
 import 'package:mobile_scanner/src/enums/mobile_scanner_authorization_state.dart';
 import 'package:mobile_scanner/src/enums/mobile_scanner_error_code.dart';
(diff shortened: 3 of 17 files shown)


lib/src/web/mobile_scanner_web.dart
```diff
@@ -126,31 +131,23 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
       return;
     }
 
-    if (videoTrack.getSettings().facingMode == 'user') {
+    final MediaStreamTrack videoTrack =
+        videoStream.getVideoTracks().toDart.first;
+
+    // On MacOS, even though the facing mode is supported, it is not reported.
+    // Use the label for FaceTime cameras to detect the user facing webcam.
+    if (videoTrack.label.contains('FaceTime')) {
```
navaronbracke, line 150: This is an additional workaround for the mirroring on MacOS. For example Google Meet does this for you.
This is also an issue for the native (non-web) MacOS preview. But I might fix that later.

lib/src/web/media_track_extension.dart
```diff
@@ -1,18 +1,19 @@
 import 'dart:js_interop';
 import 'package:web/web.dart';
 
-/// This extension provides nullable properties for [MediaStreamTrack],
-/// for cases where the properties are not supported by all browsers.
-extension NullableMediaStreamTrackCapabilities on MediaStreamTrack {
-  /// The `getCapabilities` function is not supported on Firefox.
-  @JS('getCapabilities')
-  external JSFunction? get getCapabilitiesNullable;
```
navaronbracke, line 9: Now unused, replaced by the nullable getter below

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -434,10 +434,18 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                 let answer: [String : Any?]
 
                 if let device = self.device {
+                    let cameraDirection: Int? = switch(device.position) {
```
navaronbracke, line 437: Fun fact: MacOS reports `.unspecified` for the front facing FaceTime webcam. Probably because there is only 1 camera?

CHANGELOG.md
```diff
@@ -1,7 +1,12 @@
 ## NEXT
 
+**BREAKING CHANGES:**
+
+* The initial state of the `MobileScannerState` camera facing direction is changed to `CameraFacing.unknown`.
```
navaronbracke, line 5: This was done to keep the actual camera state in sync with the value

lib/src/web/mobile_scanner_web.dart
```diff
@@ -126,31 +131,23 @@ class MobileScannerWeb extends MobileScannerPlatform {
     HTMLVideoElement videoElement,
     MediaStream videoStream,
   ) {
-    final List<MediaStreamTrack> tracks = videoStream.getVideoTracks().toDart;
```
navaronbracke, line 129: This was actually duplicated, so I cleaned it up.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -307,13 +308,16 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
+    final CameraFacing cameraDirection =
+        CameraFacing.fromRawValue(startResult['cameraDirection'] as int?);
+
     _textureId = textureId;
 
     if (defaultTargetPlatform == TargetPlatform.android) {
       _surfaceProducerDelegate =
           AndroidSurfaceProducerDelegate.fromConfiguration(
         startResult,
-        startOptions.cameraDirection,
```
navaronbracke, line 316: To stay consistent with the camera state, we should take the camera direction from the start result

lib/src/web/barcode_reader.dart
```diff
@@ -16,16 +16,25 @@ abstract class BarcodeReader {
   /// This constructor is const, for subclasses.
   const BarcodeReader();
 
-  /// Whether the video feed is paused
-  bool? get paused =>
-      throw UnimplementedError('paused has not been implemented.');
+  /// Whether the video feed is paused.
```
navaronbracke, line 19: Formatting, missing period & adding the `videoStream` getter

lib/src/web/media_track_constraints_delegate.dart
```diff
@@ -8,6 +9,31 @@ final class MediaTrackConstraintsDelegate {
   /// Constructs a [MediaTrackConstraintsDelegate] instance.
   const MediaTrackConstraintsDelegate();
 
+  /// Get the camera direction from the given [videoStream].
+  CameraFacing getCameraDirection(MediaStream? videoStream) {
```
navaronbracke, line 13: I moved this conversion code into the delegate, so that we can reuse it and that it stays in one place.

juliansteenbakker (approved):
Thanks for your work! Looks good to me.

PR #1329, opened 2025-02-14
fix: correct android preview orientation

### LINKED ISSUE #1319: 7.0.0-beta.6: Incorrect camera preview orientation

navaronbracke:
This PR adjusts the preview orientation for Android.

Previously the formula was not entirely correct.

Fixes #1319


camsim99:
Thanks for testing! What device are you testing on for `handlesCropAndRotation` is true? And is it an emulator? Either way, I agree; that would be a separate bug that is not camera-specific.

I don't know if naturally landscape oriented devices will work yet but if it does, that'd be a happy surprise, so let me know.

navaronbracke:
I was testing on this device, which is running Android 13: https://www.gsmarena.com/samsung_galaxy_xcover6_pro-11600.php

I did not test emulators, since I know that those might not be working properly yet.
I'll try to see if I can find a tablet to test on as well.

navaronbracke:
Also, the Motorola had its Vulkan support fixed in Flutter 3.29, but I doubt that that is related, as it still reports `handlesCropAndRotation: true`, which indicates it is using the old API under the hood.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -621,7 +638,7 @@ class MobileScanner(
         }
         val paint = Paint().apply { colorFilter = ColorMatrixColorFilter(colorMatrix) }
 
-        val invertedBitmap = Bitmap.createBitmap(bitmap.width, bitmap.height, bitmap.config)
+        val invertedBitmap = Bitmap.createBitmap(bitmap.width, bitmap.height, bitmap.config!!)
```
navaronbracke, line 641: This is a change that wasn't surfaced when using Kotlin 1.7, hence the bump to Kotlin 1.8

example/pubspec.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   flutter:
     sdk: flutter
 
-  image_picker: ^1.0.4
```
navaronbracke, line 22: Bumping image picker (and updating the pubspec.lock locally) fixed a build issue with the removal of the Android v1 embedding.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -191,7 +191,7 @@ class MobileScannerHandler(
                         "textureId" to it.id,
                         "size" to mapOf("width" to it.width, "height" to it.height),
                         "naturalDeviceOrientation" to it.naturalDeviceOrientation,
-                        "isPreviewPreTransformed" to it.isPreviewPreTransformed,
+                        "handlesCropAndRotation" to it.handlesCropAndRotation,
```
navaronbracke, line 194: Renamed to match the actual name from the SurfaceProducer API.

example/android/.gitignore
```diff
@@ -11,3 +11,5 @@ GeneratedPluginRegistrant.java
 key.properties
 **/*.keystore
 **/*.jks
+
+**/.cxx
```
navaronbracke, line 15: I had to add this line, since Flutter 3.29 would add a .cxx folder when compiling on Android. New `flutter create`d templates did include a change like this, too.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -356,7 +365,6 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
 
     _textureId = null;
     _pausing = false;
-    _surfaceProducerDelegate?.dispose();
```
navaronbracke, line 359: The disposal of the "delegate" is no longer needed, since the stream subscription is moved to a widget.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -319,9 +331,6 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
         startResult,
         cameraDirection,
       );
-      _surfaceProducerDelegate?.startListeningToDeviceOrientation(
```
navaronbracke, line 322: This is now handled by a separate widget.

lib/src/method_channel/rotated_preview.dart
```diff
@@ -0,0 +1,131 @@
...
+
+  @override
+  Widget build(BuildContext context) {
+    final double rotation = _computeRotation(
+      deviceOrientation,
+      sensorOrientationDegrees: widget.sensorOrientationDegrees,
+      sign: widget.facingSign,
+    );
+
+    // FIXME: Can this be improved?
```
juliansteenbakker, line 125: I don't think there is a better way to handle this right?

juliansteenbakker (approved):

lib/src/method_channel/rotated_preview.dart
navaronbracke: I agree. I think it is fine to remove the comment here. We can always revisit this specific code if needed.

PR #1332, opened 2025-02-17
ci: migrate to new coverage and quality actions

juliansteenbakker:


.github/workflows/flutter.yml
```diff
@@ -1,62 +0,0 @@
...
-      - uses: actions/setup-java@v4
-        with:
-          java-version: 17
-          distribution: temurin
-      - uses: subosito/flutter-action@v2
-        with:
-          cache: true
-          # TODO: move the code to the new tall style once Flutter 3.29 is the new minimum.
-          # https://github.com/juliansteenbakker/mobile_scanner/issues/1326
-          flutter-version: '3.27'
```
juliansteenbakker, line 60: I think it is safe to remove this in the new actions, because it does not make any changes testing to 3.27 or the current 3.29. What do you think @navaronbracke

navaronbracke: I'm fine with that, we will need to commit a `dart format .` pass with the new style, though.

navaronbracke:
@juliansteenbakker It's good to see code coverage being enabled. However, since we do not have many tests currently, how will this work? It picks the current develop branch as baseline and the "number has to go up" ?

juliansteenbakker:
> @juliansteenbakker It's good to see code coverage being enabled. However, since we do not have many tests currently, how will this work? It picks the current develop branch as baseline and the "number has to go up" ?

Exactly. It will give insight into what is tested and what not, so it will be an incentive for new PRs to also write tests if possible. It does not matter that the number will be low right now.

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,65 @@
...
+            .dart_tool
+          key: ${{ runner.os }}-pub-${{ hashFiles('**/pubspec.yaml') }}
+          restore-keys: |
+            ${{ runner.os }}-pub-
+
+      - name: 🔧 Bootstrap Workspace
+        run: flutter pub get
+
+      - name: ✅ Lint Code
+        run: flutter analyze
```
navaronbracke, line 44: Should we configure fatal info's etc for `flutter analyze` ?

.github/workflows/flutter.yml
juliansteenbakker: `dart format .` will not change until we change the sdk constraint to 3.7.0 or higher, so it's only something to look after when upgrading the constraint.
```pubspec
environment:
  sdk: ">=3.6.0 <4.0.0"
```

.github/workflows/code-quality.yml
```diff
@@ -0,0 +1,65 @@
...
+        uses: subosito/flutter-action@v2
+        with:
+          channel: stable
+          cache: true
+
+      - name: 🔧 Bootstrap Workspace
+        run: flutter pub get
+
+      - name: 🎯 Check Code Formatting
+        run: dart format --set-exit-if-changed .
```
navaronbracke, line 65: Can we explicitly set the page width in the analysis options also?
I have my global line length for Dart set to 120, but I would like to enforce the line length in this repo (to the standard 80?)

image

navaronbracke (changes requested):
Just some minor feedback, but LGTM otherwise.

.github/workflows/flutter.yml
navaronbracke: Right, sorry, I keep forgetting that this is tied to the minimum language version :p

.github/workflows/code-quality.yml
juliansteenbakker: What do you mean exactly? Those are on by default right?

```
    --[no-]fatal-infos        Treat info level issues as fatal.
                              (defaults to on)
    --[no-]fatal-warnings     Treat warning level issues as fatal.
                              (defaults to on)
```

navaronbracke: Ah, they are? Nice

navaronbracke (approved):

juliansteenbakker:
@navaronbracke i saw a comment of you that you've deleted about the dart code width. I would love to make it bigger than the default 80 characters, however, i think its good to wait a little bit and see if there is any new standard size coming up. I can see certain linters like very_good_analysis adapting something like a default "wide view". Until then, i think its best to stick with the default 80 characters for now.

navaronbracke:
The `page_width` can only be set if the minimum language version is 3.7

juliansteenbakker:
Of course that's also a requirement 😃

PR #1335, opened 2025-02-18
refactor: example app update

juliansteenbakker:
This pr updates the example app to be more modern and easier to understand.

It merges all different functions into one "advanced mobile scanner" page. The simple scanner page still exists, as wel as the custom crosshair implementation.

It would be great if we can merge the crosshair implementation into the advanced page.


juliansteenbakker:
@navaronbracke i think this pr is ready for now. The only issue blocking the release is the pause function not working on Android, but i'll fix that in another PR. Also, there is one example not working right now (Picklist example), however this does not have a high priority so i'd rather merge this PR first.

This PR not only reworks the whole example app, it also adds multiple tests, updates the readme.md, fixes the scanwindow not working correctly on apple, and and a few other minor improvements.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -358,6 +363,7 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         }
 
         device.addObserver(self, forKeyPath: #keyPath(AVCaptureDevice.torchMode), options: .new, context: nil)
+        device.addObserver(self, forKeyPath: "videoZoomFactor", options: [.new, .initial], context: nil)
```
navaronbracke, line 366: Is there a constant for `videoZoomFactor ` in AVCaptureDevice? Can we use the keypath for https://developer.apple.com/documentation/avfoundation/avcapturedevice/videozoomfactor ?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+
+class BarcodeScannerAdvanced extends StatefulWidget {
+  const BarcodeScannerAdvanced({super.key});
+
+  @override
+  State<BarcodeScannerAdvanced> createState() => _BarcodeScannerAdvancedState();
+}
+
+class _BarcodeScannerAdvancedState extends State<BarcodeScannerAdvanced> {
+  // Cannot be changed on runtime while the scanner is running
```
navaronbracke, line 35:
```suggestion
  // Cannot be changed while the scanner is running.
```

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -945,7 +970,7 @@ extension VNBarcodeObservation {
                 ["x": bottomLeftX, "y": bottomLeftY],
             ],
             "format": symbology.toInt ?? -1,
-            "rawBytes": rawBytes,
+            "rawBytes": rawBytes ?? "",
```
navaronbracke, line 973: This is not correct. The `rawBytes` is a nillable typed data, not a string

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+  static const useScanWindow = true;
+
+  bool autoZoom = false;
+  bool invertImage = false;
+  bool returnImage = false;
+
+  Size size = const Size(1920, 1080);
+  DetectionSpeed detectionSpeed = DetectionSpeed.unrestricted;
+  int detectionTimeout = 1000; // Default to 1000ms
```
navaronbracke, line 44: Can we use a `Duration` with milliseconds?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+
+class _BarcodeScannerAdvancedState extends State<BarcodeScannerAdvanced> {
+  // Cannot be changed on runtime while the scanner is running
+  static const useScanWindow = true;
+
+  bool autoZoom = false;
+  bool invertImage = false;
+  bool returnImage = false;
+
+  Size size = const Size(1920, 1080);
```
navaronbracke, line 42: Can we rename this to `desiredCameraResolution` ?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+              ),
+            ],
+          ),
+        );
+      },
+    );
+  }
+
+  Future<void> _showResolutionDialog() async {
+    final widthController = TextEditingController();
```
navaronbracke, line 122: We should move these controllers to the widget level & dispose them.

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+
+    await showDialog<void>(
+      context: context,
+      builder: (context) {
+        return AlertDialog(
+          title: const Text('Set Camera Resolution'),
+          content: Column(
+            mainAxisSize: MainAxisSize.min,
+            children: [
+              TextField(
```
navaronbracke, line 133: Do we add some basic validation?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+  }
+
+  Future<void> _showDetectionSpeedDialog() async => showDialog<void>(
+        context: context,
+        builder: (context) {
+          return AlertDialog(
+            title: const Text('Select Detection Speed'),
+            content: Column(
+              mainAxisSize: MainAxisSize.min,
+              children: DetectionSpeed.values.map((speed) {
```
navaronbracke, line 177: Can you use a for literal?

example/lib/widgets/buttons/pause_button.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:flutter/material.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
+
+class PauseButton extends StatelessWidget {
+  const PauseButton({required this.controller, super.key});
+
+  final MobileScannerController controller;
+
+  Future<void> onPressed() async {
```
navaronbracke, line 9: Can we inline this function?

example/lib/widgets/buttons/start_stop_button.dart
```diff
@@ -0,0 +1,40 @@
+import 'package:flutter/material.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
+
+class StartStopButton extends StatelessWidget {
+  const StartStopButton({required this.controller, super.key});
+
+  final MobileScannerController controller;
+
+  Future<void> onPressedStop() async {
```
navaronbracke, line 9: Can we inline start / stop ?

example/lib/utilities/barcode_at_center.dart
```diff
@@ -1,10 +1,10 @@
 import 'package:flutter/services.dart';
 import 'package:mobile_scanner/mobile_scanner.dart';
-import 'package:mobile_scanner_example/picklist/classes/fix_coordinate_space.dart';
+import 'package:mobile_scanner_example/utilities/fix_coordinate_space.dart';
 
 /// This function finds the barcode that touches the center of the
 /// image. If no barcode is found that touches the center, null is returned.
-/// See [_BarcodeScannerPicklistState] and the returnImage option for more info.
+/// See _BarcodeScannerPicklistState and the returnImage option for more info.
```
navaronbracke, line 7:
```suggestion
/// See `_BarcodeScannerPicklistState` and the returnImage option for more info.
```

example/lib/utilities/barcode_at_center.dart
```diff
@@ -57,9 +58,9 @@ bool _isPointInPolygon({
 }) {
   // Initial variables:
   int i; // Loop variable for current vertex
-  int j = polygon.length -
+  var j = polygon.length -
```
navaronbracke, line 61: Personally I do not like using `var`, I would rather have the actual type in code, as it is not obvious during code review.

example/lib/utilities/barcode_at_center.dart
```diff
@@ -69,19 +70,23 @@ bool _isPointInPolygon({
         (point.dx <
             (polygon[j].dx -
                         polygon[i]
-                            .dx) * // Horizontal distance between the vertices of the edge
```
navaronbracke, line 72: Can we move the comments so that the indent is better for the calculation?

example/lib/widgets/buttons/switch_camera_button.dart
```diff
@@ -0,0 +1,50 @@
+import 'package:flutter/material.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
+
+class SwitchCameraButton extends StatelessWidget {
+  const SwitchCameraButton({required this.controller, super.key});
+
+  final MobileScannerController controller;
+
+  Future<void> onPressed() async {
```
navaronbracke, line 9: Can we inline this function?

lib/src/mobile_scanner.dart
```diff
@@ -116,6 +117,11 @@ class MobileScanner extends StatefulWidget {
   /// Defaults to no threshold for scan window updates.
   final double scanWindowUpdateThreshold;
 
+  /// Whether to pause and resume the barcode scanner on lifecycleState change
```
navaronbracke, line 120:
```suggestion
  /// Whether the `MobileScanner` widget should automatically pause and resume when the application lifecycle state changes.
  /// If this is false, the controller should manually be resumed or paused when the application lifecycle changes.
  ///
  /// Defaults to true.
  
```

lib/src/mobile_scanner.dart
```diff
@@ -281,21 +287,23 @@ class _MobileScannerState extends State<MobileScanner>
...
     }
     // When this widget is unmounted, reset the scan window.
-    unawaited(controller.updateScanWindow(null));
+    await controller.updateScanWindow(null);
 
     // Dispose default controller if not provided by user
     if (widget.controller == null) {
-      controller.dispose();
+      await controller.dispose();
       WidgetsBinding.instance.removeObserver(this);
```
navaronbracke, line 298: Since this is now async, we should probably call `removeObserver(this);` right after `super.dispose()`, but before the async gap starts. That is also related to why we do `super.dipose()` before an await.

lib/src/mobile_scanner.dart
```diff
@@ -281,21 +287,23 @@ class _MobileScannerState extends State<MobileScanner>
...
     }
   }
 
   @override
   void didChangeAppLifecycleState(AppLifecycleState state) {
-    if (widget.controller != null || !controller.value.hasCameraPermission) {
+    if (!widget.useAppLifecycleState ||
+        (widget.controller != null && widget.useAppLifecycleState) ||
```
navaronbracke, line 305: Why do we have the middle check? What if the user did not provide a controller but `useAppLifecycleState` is true? Is this a typo?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+        },
+      );
+
+  Future<void> _showDetectionTimeoutDialog() async => showDialog<void>(
+        context: context,
+        builder: (context) {
+          var tempTimeout =
+              detectionTimeout; // Temporary variable to hold the slider value
+
+          return StatefulBuilder(
```
navaronbracke, line 203: Can we use a separate Stateful widget for the dialog?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+
+  @override
+  void initState() {
+    super.initState();
+    unawaited(controller.start());
+  }
+
+  double _zoomFactor = 0;
+
+  Widget _buildZoomScaleSlider() {
```
navaronbracke, line 74: Having all these separate methods feels like an anti pattern? Should we keep these as separate widget classes?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+              }).toList(),
+            ),
+          );
+        },
+      );
+
+  Future<void> _showDetectionTimeoutDialog() async => showDialog<void>(
+        context: context,
+        builder: (context) {
+          var tempTimeout =
```
navaronbracke, line 200: Can we use `Duration` for the temp timeout, instead of var?

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+              detectionTimeout; // Temporary variable to hold the slider value
+
+          return StatefulBuilder(
+            builder: (context, setDialogState) {
+              return AlertDialog(
+                title: const Text('Set Detection Timeout (ms)'),
+                content: SizedBox(
+                  height: 100, // Restrict height
```
navaronbracke, line 208: This comment is a bit redundant :P

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+          return StatefulBuilder(
+            builder: (context, setDialogState) {
+              return AlertDialog(
+                title: const Text('Select BoxFit'),
+                content: Column(
+                  mainAxisSize: MainAxisSize.min,
+                  children: BoxFit.values.map((fit) {
+                    return RadioListTile<BoxFit>(
+                      title:
+                          Text(fit.toString().split('.').last), // Display name
```
navaronbracke, line 261: Use `.name` of enum values!

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+                  autoZoom = !autoZoom;
+                case PopupMenuItems.useBarcodeOverlay:
+                  useBarcodeOverlay = !useBarcodeOverlay;
+              }
+
+              await controller.dispose();
+              controller = initController();
+              await controller.start();
+
+              setState(() {});
```
navaronbracke, line 389: Add a mounted check after the await

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+            },
+            itemBuilder: (context) => [
+              if (Platform.isAndroid)
```
navaronbracke, line 392:
```suggestion
              if (!kIsWeb && Platform.isAndroid)
```

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+              PopupMenuItem(
+                value: PopupMenuItems.boxFit,
+                child: Text(PopupMenuItems.boxFit.name),
+              ),
+              PopupMenuItem(
+                value: PopupMenuItems.formats,
+                child: Text(PopupMenuItems.formats.name),
+              ),
+              const PopupMenuDivider(),
+              if (Platform.isAndroid)
```
navaronbracke, line 415:
```suggestion
              if (!kIsWeb && Platform.isAndroid)
```

example/lib/screens/barcode_scanner_advanced.dart
```diff
@@ -0,0 +1,549 @@
...
+                CheckedPopupMenuItem(
+                  value: PopupMenuItems.autoZoom,
+                  checked: autoZoom,
+                  child: Text(PopupMenuItems.autoZoom.name),
+                ),
+              if (Platform.isAndroid)
```
navaronbracke, line 421:
```suggestion
              if (!kIsWeb && Platform.isAndroid)
```

test/barcode_painter_test.dart
```diff
@@ -0,0 +1,226 @@
...
+        color: Colors.blue,
+        style: PaintingStyle.stroke,
+        barcodeValue: '123456',
+      );
+
+      // Act: Call the paint method
+      barcodePainter.paint(mockCanvas, const Size(200, 200));
+
+      // Assert: Verify drawing operations
+      verify(() => mockCanvas.drawPath(any(), any())).called(1);
```
navaronbracke, line 40: Can we use `paints` matchers, like in the Flutter Framework Widget tests?

test/barcode_painter_test.dart
```diff
@@ -0,0 +1,226 @@
...
+        cameraPreviewSize: Size(200, 200),
+        color: Colors.blue,
+        style: PaintingStyle.stroke,
+        barcodeValue: '123456',
+      );
+
+      barcodePainter.paint(mockCanvas, const Size(200, 200));
+
+      // Ensure text rotation is applied
+      verify(() => mockCanvas.translate(any(), any())).called(2);
```
navaronbracke, line 87: Should we verify paint is called?

test/barcode_painter_test.dart
```diff
@@ -0,0 +1,226 @@
...
+        ],
+        barcodeSize: Size(100, 50),
+        boxFit: BoxFit.contain,
+        cameraPreviewSize: Size(200, 200),
+        color: Colors.red, // Changed color
+        style: PaintingStyle.stroke,
+        barcodeValue: '123456',
+      );
+
+      expect(painter1.shouldRepaint(painter2), isTrue);
```
navaronbracke, line 123: We should test more property changes here

lib/src/overlay/barcode_painter.dart
```diff
@@ -2,16 +2,17 @@ import 'dart:math' as math;
...
-  /// Construct a new [BarcodePainter] instance.
+  /// Constructor for [BarcodePainter]
   const BarcodePainter({
     required this.barcodeCorners,
     required this.barcodeSize,
     required this.boxFit,
     required this.cameraPreviewSize,
     required this.color,
     required this.style,
+    required this.barcodeValue,
```
navaronbracke, line 15: Can we move the barcode value next to the corners & size?

lib/src/overlay/barcode_painter.dart
```diff
@@ -2,16 +2,17 @@ import 'dart:math' as math;
 import 'package:collection/collection.dart';
 import 'package:flutter/material.dart';
 
-/// This class represents a [CustomPainter] that draws the [barcodeCorners] of a single barcode.
+/// CustomPainter to draw an outlined barcode box with rounded corners and a displayed value.
```
navaronbracke, line 5:
```suggestion
/// A [CustomPainter] that draws the barcode as an outlined barcode box with rounded corners and a displayed value.
```

lib/src/overlay/barcode_painter.dart
```diff
@@ -2,16 +2,17 @@ import 'dart:math' as math;
...
 class BarcodePainter extends CustomPainter {
-  /// Construct a new [BarcodePainter] instance.
+  /// Constructor for [BarcodePainter]
```
navaronbracke, line 7: I do prefer the old comment, as it is a full phrase, per Effective Dart for DartDoc.

lib/src/overlay/barcode_painter.dart
```diff
@@ -20,22 +21,24 @@ class BarcodePainter extends CustomPainter {
   /// The size of the barcode.
   final Size barcodeSize;
 
-  /// The [BoxFit] to use when painting the barcode box.
+  /// The BoxFit mode for scaling.
```
navaronbracke, line 24:
```suggestion
  /// The BoxFit mode for scaling the barcode bounding box.
```

lib/src/overlay/barcode_painter.dart
```diff
@@ -56,25 +59,95 @@ class BarcodePainter extends CustomPainter {
...
-      ..color = color
-      ..style = style;
+    // Draw the rotated rectangle
+    final Path path = Path()..addPolygon(adjustedOffset, true);
 
-    canvas.drawPath(cutoutPath, backgroundPaint);
+    final paint = Paint()
+      ..color = color
+      ..style = style
+      ..strokeWidth = 4.0;
```
navaronbracke, line 70: Do we add a parameter for the stroke width?

lib/src/overlay/barcode_painter.dart
```diff
@@ -56,25 +59,95 @@ class BarcodePainter extends CustomPainter {
...
+    final textSpan = TextSpan(
+      text: barcodeValue,
+      style: TextStyle(
+        color: Colors.black, // Ensuring black text
+        fontSize: finalTextSize,
+        fontWeight: FontWeight.bold,
+      ),
+    );
+
+    final textPainter = TextPainter(
```
navaronbracke, line 102: You should dispose the text painter when done with it.

test/mobile_scanner_test.dart
```diff
@@ -1,26 +1,185 @@
...
+
+  group('MobileScanner Widget Tests', () {
+    testWidgets('renders correctly with default parameters', (tester) async {
+      await tester.pumpWidget(
+        const MaterialApp(
+          home: Scaffold(body: MobileScanner()),
+        ),
+      );
+
+      expect(find.byType(MobileScanner), findsOneWidget);
```
navaronbracke, line 56: Should we assert that a preview is drawn?

test/mobile_scanner_test.dart
```diff
@@ -1,26 +1,185 @@
...
+      );
+
+      await tester.pumpWidget(testWidget);
+      await tester.pumpAndSettle();
+      await tester
+          .pumpWidget(Container()); // Remove the widget to trigger disposal
+
+      // Since MobileScanner created its own controller, it should be disposed.
+      // However, we cannot verify it directly because it's internal.
+      // Instead, we ensure there is no exception.
```
navaronbracke, line 138: You can use `expect(tester.takeException(), isNull);` for the latter

test/mobile_scanner_test.dart
```diff
@@ -1,26 +1,185 @@
...
+
+      await tester.pumpAndSettle();
+      await tester.pumpWidget(Container()); // Remove the widget
+
+      // Verify that dispose() was NOT called because it's externally managed
+      verifyNever(() => mockController.dispose());
+    });
+  });
+
+  group('MobileScannerController Tests', () {
```
navaronbracke, line 157: These tests don't actually verify the behavior? We should be using a mock stream handler / method call handler for this I think.

navaronbracke (changes requested):
I left an initial review with a lot of feedback.

Perhaps we can split this up into smaller PR's for easier review of the feedback?

juliansteenbakker:
> I left an initial review with a lot of feedback.
> 
> Perhaps we can split this up into smaller PR's for easier review of the feedback?

Thank you for your in depth review! This PR indeed got a bit out of hand, and solves multiple things that should have been split.

I will first fix the issues you've found, and then look if its feasible to split it into multiple PR's.

example/analysis_options.yaml
```diff
@@ -1 +1,4 @@
-include: package:lint/strict.yaml
+include: package:very_good_analysis/analysis_options.yaml
+linter:
+  rules:
+    public_member_api_docs: false
```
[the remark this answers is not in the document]
juliansteenbakker: Forgot to add docs to be honest. I've added them now!

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -42,6 +42,12 @@ class MobileScannerHandler(
 
     private val analyzeImageSuccessCallback: AnalyzerSuccessCallback = {
         Handler(Looper.getMainLooper()).post {
+            // Also publish on controller result
+            barcodeHandler.publishEvent(mapOf(
```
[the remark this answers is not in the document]
juliansteenbakker: This shouldn't be there, it was a test because the example app didn't react to the result of `analyzeImage`. I've removed it, and we can discuss this in a separate issue/PR. Thanks for the catch!

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: Good catch!

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: I've used the same implementation here as `displayValue`.

The correct fix is to change from `[String : Any]` to `[String : Any?]`, to allow swift to pass nil back to dart.

example/lib/screens/barcode_scanner_advanced.dart
juliansteenbakker: The current implementation is inline with `detectionTimeoutMs` in the controller. If we change this, i think it would be better to introduce a new duration parameter to the controller, and mark the current one as deprecated. What do you think about this? Also, if we decide to do this, it would be better to do in a new PR.

example/lib/utilities/barcode_at_center.dart
juliansteenbakker: I understand, but following `very_good_analysis`, the `omit_local_variable_types` lint rule enforces usage of var here. We can debate if we could change this in another PR, but for this pr i think it's good to follow the current rules.

lib/src/mobile_scanner.dart
juliansteenbakker: The comparison can be refactored indeed because useAppLifecycleState is always true after first check. i've also updated the docs to be more clear about the function.

example/lib/screens/barcode_scanner_advanced.dart
juliansteenbakker: Yes you're absolutely right, i've created a separate widget for this.

example/lib/screens/barcode_scanner_advanced.dart
juliansteenbakker: Lol you're right. I've generated these functions but the comments are not very helpful.. I've removed those.

example/lib/screens/barcode_scanner_advanced.dart
juliansteenbakker: I think it would be better to migrate to duration when we also migrate the other parameters for detection timout to duration.

test/mobile_scanner_test.dart
juliansteenbakker: Indeed. I'm still trying to understand how to do this correctly, so i'll be spending some more time on creating good tests for the package. I think it's better to do this in a separate PR, so ill comment out these tests for now.

lib/src/overlay/barcode_painter.dart
juliansteenbakker: Good one!

juliansteenbakker:
@navaronbracke I'm almost done going trough the review! As said in the comments, i think we can handle the the tests and the duration issues in a different PR.

I'll be doing a final walkthrough tomorrow. Let me know if there is anything else i should change, and thanks again for the review!

Issue #1305, opened 2025-01-27
iPhone requires phone to be rotated to scan QR code (Android works)


chickahoona:
I have this fairly standard code:

```dart
MobileScanner(
  controller: controller,
  onDetect: (BarcodeCapture capture) {
    final List<Barcode> barcodes = capture.barcodes;
    if (barcodes.isNotEmpty) {
      setState(() {
        qrScan = barcodes.first.rawValue;
        redirect = true;
      });
    }
  },
  fit: BoxFit.contain,
  scanWindow: Rect.fromCenter(
    center: Offset(
      MediaQuery.of(context).size.width / 2,
      MediaQuery.of(context).size.height / 2,
    ),
    width: scanArea,
    height: scanArea,
  ),
);
```
Full code of the app here: https://gitlab.com/psono/psono-app/-/blob/master/lib/screens/scan_qr/index.dart?ref_type=heads#L94-114

And when I scan this QR code

![Image](https://github.com/user-attachments/assets/21103eba-62ff-42fd-be2d-a1b26bb3a42e)

Then an iPhone won't accept that QR code unless i rotate the phone by 90°.
On an Android device it works in both rotations...

navaronbracke:
What version of mobile_scanner is this?

chickahoona:
6.0.2 so fairly new

navaronbracke:
Does this reproduce with https://pub.dev/packages/mobile_scanner/versions/7.0.0-beta.5 ?

In version 7 we switched over from MLKit to Apple's Vision API for iOS (MacOS was already using it)

navaronbracke:
@chickahoona I cannot reproduce this with mobile_scanner version 7.0.0-beta.5.

I used your code to modify the scan window in the `MobileScanner with ScanWindow` sample:

<details>

  <summary>Code Sample</summary>

```
import 'package:flutter/material.dart';
import 'package:mobile_scanner/mobile_scanner.dart';
import 'package:mobile_scanner_example/scanned_barcode_label.dart';

import 'package:mobile_scanner_example/scanner_error_widget.dart';

class BarcodeScannerWithScanWindow extends StatefulWidget {
  const BarcodeScannerWithScanWindow({super.key});

  @override
  State<BarcodeScannerWithScanWindow> createState() => _BarcodeScannerWithScanWindowState();
}

class _BarcodeScannerWithScanWindowState extends State<BarcodeScannerWithScanWindow> {
[...]

PR #1339, opened 2025-02-21
fix: improve video orientation handling for iOS 13 and above

SungKyuMM:
When the app's orientation is fixed but the device is held in a different direction, the camera view may sometimes appear rotated incorrectly.

- For iOS 13+: Use UIWindowScene.interfaceOrientation for accurate orientation detection.
- Fallback to UIDevice.current.orientation for earlier iOS versions.

### DIFF
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -275,6 +275,26 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     
     private func getVideoOrientation() -> AVCaptureVideoOrientation {
 #if os(iOS)
+        // Get the orientation from the window scene if available
+        // When the app's orientation is fixed and the app orientation is actually different from the device orientation, it malfunctions.
+        if #available(iOS 13.0, *){
+            if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene{
+                let orientation = windowScene.interfaceOrientation
+                switch orientation {
+                case .portrait:
+                    return .portrait
+                case .portraitUpsideDown:
+                    return .portraitUpsideDown
+                case .landscapeLeft:
+                    return .landscapeLeft
+                case .landscapeRight:
+                    return .landscapeRight
+                default:
+                    break
+                }         
+            }
+        }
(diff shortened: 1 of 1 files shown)


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -275,6 +275,26 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     
     private func getVideoOrientation() -> AVCaptureVideoOrientation {
 #if os(iOS)
+        // Get the orientation from the window scene if available
+        // When the app's orientation is fixed and the app orientation is actually different from the device orientation, it malfunctions.
+        if #available(iOS 13.0, *){
```
navaronbracke, line 280:
```suggestion
        if #available(iOS 13.0, *) {
```

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -275,6 +275,26 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+            if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene{
```
navaronbracke, line 281:
```suggestion
            if let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene {
```

navaronbracke (changes requested):
Minor formatting nit, but otherwise LGTM.

Can you also add a note in the changelog under the `NEXT` entry?

SungKyuMM:
> @SungKyuMM would this fix #1305 ?

While rotation might affect barcodes, it doesn’t seem to affect QR codes with respect to screen rotation. 
As mentioned earlier, I believe it’s likely a focus issue. As noted at https://github.com/juliansteenbakker/mobile_scanner/issues/1305#issuecomment-2618854179, since the code is recognized when the camera is held further away, it doesn’t appear to be directly related to rotation.

It seems that what gets recognized upon rotation is due to the device’s focus changing as the screen rotates—something that wasn’t recognized before becomes recognized. This PR was created because the actual app’s rotation direction differed from the camera view’s rotation, which caused the camera view to appear oddly rotated.

Issue #1381, opened 2025-04-08
java.lang.IllegalArgumentException: Surface was abandoned


AngeloAvv:
After switching from version 6.x to 7.x.x-beta.x, I can't use mobile_scanner on Android anymore.

Every time my QrScannerWidget is rendered, I receive the following error:

```
I/DynamiteModule(29764): Considering local module com.google.mlkit.dynamite.barcode:10000 and remote module com.google.mlkit.dynamite.barcode:0
I/DynamiteModule(29764): Selected local version of com.google.mlkit.dynamite.barcode
I/flutter (29764): [IMPORTANT:flutter/shell/platform/android/platform_view_android.cc(308)] Flutter recommends migrating plugins that create and register surface textures to the new surface producer API. See https://docs.flutter.dev/release/breaking-changes/android-surface-plugins
E/libc    (29764): Access denied finding property "persist.vendor.camera.privapp.list"
W/CameraX-core_ca(29764): type=1400 audit(0.0:9624): avc: denied { read } for name="u:object_r:vendor_persist_camera_prop:s0" dev="tmpfs" ino=29950 scontext=u:r:untrusted_app:s0:c25,c258,c512,c768 tcontext=u:object_r:vendor_persist_camera_prop:s0 tclass=file permissive=0 app=xxx.app
E/Legacy-CameraDevice-JNI(29764): getNativeWindow: Surface had no valid native window.
E/Legacy-CameraDevice-JNI(29764): LegacyCameraDevice_nativeDetectSurfaceDimens: Could not retrieve native window from surface.
W/CaptureSession(29764): Opening session with fail OPENING
W/CaptureSession(29764): java.lang.IllegalArgumentException: Surface was abandoned
W/CaptureSession(29764): 	at android.hardware.camera2.utils.SurfaceUtils.getSurfaceSize(SurfaceUtils.java:91)
W/CaptureSession(29764): 	at android.hardware.camera2.params.OutputConfiguration.<init>(OutputConfiguration.java:261)
W/CaptureSession(29764): 	at android.hardware.camera2.params.OutputConfiguration.<init>(OutputConfiguration.java:205)
W/CaptureSession(29764): 	at androidx.camera.camera2.internal.compat.params.OutputConfigurationCompatApi28Impl.<init>(OutputConfigurationCompatApi28Impl.java:41)
[...]

PR #1382, opened 2025-04-09
fix: change old surface producer API usage

navaronbracke:
This PR addresses a TODO with the SurfaceProducer implementation.

Part of #1381

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -233,9 +233,7 @@ class MobileScanner(
                             // whenever a new Surface is needed.
                         }
 
-                        // TODO: replace with "onSurfaceCleanup" when available in Flutter 3.28 or later
-                        // See https://github.com/flutter/flutter/pull/160937
-                        override fun onSurfaceDestroyed() {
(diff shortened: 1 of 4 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -233,9 +233,7 @@ class MobileScanner(
                             // whenever a new Surface is needed.
                         }
 
-                        // TODO: replace with "onSurfaceCleanup" when available in Flutter 3.28 or later
-                        // See https://github.com/flutter/flutter/pull/160937
-                        override fun onSurfaceDestroyed() {
```
navaronbracke, line 238: Since 3.29.0 has been out for a while, lets fix this bug.

pubspec.yaml
```diff
@@ -16,8 +16,8 @@ screenshots:
   path: example/screenshots/overlay.png
 
 environment:
-  sdk: ">=3.6.0 <4.0.0"
-  flutter: ">=3.27.0"
+  sdk: ">=3.7.0 <4.0.0"
+  flutter: ">=3.29.0"
```
navaronbracke, line 20: This is unfortunate, but the new API is not available otherwise.

PR #1383, opened 2025-04-09
fix: update gradle JVM args

navaronbracke:
When running the example app, it apparently now fails with a Gradle memory error.
Building the example app did work, oddly enough.

There's an open Flutter tool issue for this, so I took the new JVM args from there :)


example/android/gradle.properties
```diff
@@ -1,3 +1,3 @@
-org.gradle.jvmargs=-Xmx1536M
```
navaronbracke, line 1: 1.5 gigs is not even enough to run Emulators there days, so Gradle would probably also be an issue :P

PR #1387, opened 2025-04-11
fix: format according to dart 3.7

juliansteenbakker:
Due to the min sdk being 3.7, we should update our formatting accordingly. However, there is a new option comming that will make it possible to opt-out of this new format. @navaronbracke Do you have any opinion on this?

### DIFF
--- a/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
+++ b/example/ios/Runner.xcodeproj/xcshareddata/xcschemes/Runner.xcscheme
@@ -77,6 +77,7 @@
       ignoresPersistentStateOnLaunch = "NO"
       debugDocumentVersioning = "YES"
       debugServiceExtension = "internal"
+      enableGPUValidationMode = "1"
       allowLocationSimulation = "YES">
       <BuildableProductRunnable
          runnableDebuggingMode = "0">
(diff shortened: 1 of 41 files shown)


navaronbracke:
@juliansteenbakker I'm aware we needed to redo the formatting here (the change for onSurfaceCleanup needed the min Flutter version to go up). Where in dart-lang/sdk was the new opt-out discussed? The new formatter style is nice, but some things still need tweaking on that front I think (I.e. some of the arrow function formatting still needed fixing)

juliansteenbakker:
See https://github.com/dart-lang/dart_style/commit/71ac08514f771363b9b4b105786042d0ab91d569 for the merge request. We can enable it in the future with:

```
formatter:
  trailing_commas: preserve
```

I think the best for now is to wait until this is landed, and apply asap. Otherwise we will have a lot of git diffs for nothing.

navaronbracke:
I noticed it too. They announced it on their socials iirc. I agree we should land the formatting fix as soon as possible.

juliansteenbakker:
It seems that `trailing_commas` will not be available for sdk 3.7, but 3.8 or later. Given that we want to release version 7 with sdk 3.7, i think there is no other option than to format it correctly for this sdk version.

Source: https://github.com/dart-lang/dart_style/blob/main/CHANGELOG.md & https://github.com/dart-lang/dart_style/pull/1707

navaronbracke:
I'm aware that the formatter option was branch cut for 3.8 only. Let's just format as-is now and revisit this later. (there will be fixes for the new formatter style anyway, so we might need to do a second formatting pass once those fixes are in)

PR #1393, opened 2025-04-17
fix: hot reload

### LINKED ISSUE #773: Better support for Hot Reload
Using Hot Reload on either iOS or Android on view with camera causes MobileScanner: Called start() while already started! error.

Repro steps:

build example app to device
open any camera screen
do Hot Reload

This problem still exists in the latest version

### LINKED ISSUE #1037: Hot restart - camera no longer usable
If i do a hot restart while camera is displayed, I no longer have the possibility to use it until app is closed.

```
MobileScannerException: code genericError, message: Called start() while already started
```

mobile_scanner: ^5.0.1

### LINKED ISSUE #1247: Permission error not reset in start/stop method
Hi ! I am developing a usecase in a Flutter APP of qr scanner via your great package. After being denied twice (permanently denied), i use openSettings() function to open settings. 

Incase i configure the camera permission, i need that my app when the lifecycle state -> resumed to start the Camera instead of displaying the ScannerErrorWidget. 
I use Cubit to maintain Permission state across App to enable dialog boxes and rebuild app as and when the state changes.

This is some of the code i have written for it:

```
@override
    if (!controller.value.isInitialized) {
      return;
    }

    switch (state) {
      case AppLifecycleState.detached:
      case AppLifecycleState.hidden:
      case AppLifecycleState.paused:
        return;
      case AppLifecycleState.resumed:
        if (_isReturningFromSettings) {
          _isReturningFromSettings = false;
          context.read<MobileScannerCubit>().checkCameraPermission(); // Recheck permission
        } 
          _initializeCamera();

        break;
      case AppLifecycleState.inactive:
        _isScannerDeactivated = true;
        unawaited(_subscription?.cancel());
        _subscription = null;
        unawaited(controller.stop());
    ```
    
    ```
    void _initializeCamera() {
    if (_isScannerDeactivated) return;
    
    try {
      if (!controller.value.isRunning) {
        _subscription = controller.barcodes.listen(_handleBarcode);
        unawaited(controller.start());
      }
    } catch (e) {
      logger!.d("Error initializing camera: $e");
      _deactivateCamera();
    }
  }
  ```
  
  I maintain two states understand the phase of the flow we are at a particular instance
  ```
    bool _isScannerDeactivated = false;
  bool _isReturningFromSettings = false;
  ```
  
  Cubit main functions :
  
  ```
  Future<void> checkCameraPermission() async {
    if (Platform.isAndroid || Platform.isIOS) {
        var permissionStatus = await Permission.camera.status;
        logger?.d(permissionStatus);
        if (await Permission.camera.isPermanentlyDenied) {
            emit(MobileScannerPermanentlyDeniedState());
        } else if(await Permission.camera.isDenied){
            emit(MobileScannerDeniedState());
        } else if(await Permission.camera.isGranted){
            emit(MobileScannerGrantedState());
        }
    }
  }

  Future<void> openAppPermissionSettings() async{
    openAppSettings();
  }
  
  ```
  
  Bloc Listner Code
  
  ```
  listener: (context, permissionState) {
          if (!_isDialogShowing && !_isScannerDeactivated &&  
          (permissionState is MobileScannerDeniedState || 
              permissionState is MobileScannerPermanentlyDeniedState)) {
            _isDialogShowing = true;  // Set flag before showing dialog
            _showScannerPermissionDialog(context, permissionState);
          } else if (permissionState is MobileScannerGrantedState) {
              _isScannerDeactivated = false;
              _initializeCamera();
          }
        },
        ```
        
        
        ```
  child: MobileScanner(
            controller: controller,
            onDetect: _handleBarcode,
            errorBuilder: (scannerContext, error, child) {
              String errorMessage;
              const Widget defaultError = ColoredBox(
                  color: Colors.black,
                  child: Center(child: Icon(Icons.error, color: Colors.white)),
                );
              if (!_isDialogShowing && !_isScannerDeactivated) {
                if(error.errorCode == MobileScannerErrorCode.permissionDenied){
                  errorMessage = errorCameraPermissionDenied;
                  context.read<MobileScannerCubit>().checkCameraPermission();
                } else {
                  errorMessage = error.toString();
                }
              }
              return defaultError;
            },
 ```
  
  
  I extensively searched across your documentation and examples but couldnt find any examples to hint in this direction.
  
  If you could let me know, any ideas or or direction. 
  
  One of the major issues, i am facing is this even after configuring Camera Permission from settings, the Controller Error Widget goes on an inifinite loop of diaplying the error even when i have verified via `permission_handler` package that the permission status is `granted`

### PR DESCRIPTION
**Pull request by juliansteenbakker**
Fixes hot reload function by adding force parameter to `stop()`, and calling this on every start when `kDebugMode` is true. 
Merge after #1392 

Fixes #773 
Fixes #1037 
possibly fixes #1247

### DIFF
--- a/ios/Classes/MobileScanner.swift
+++ b/ios/Classes/MobileScanner.swift
@@ -303,18 +303,20 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     }
 
     /// Pause scanning for barcodes
-    func pause() throws {
-        if (paused) {
-            throw MobileScannerError.alreadyPaused
-        } else if (stopped) {
-            throw MobileScannerError.alreadyStopped
+    func pause(force: Bool = false) throws {
+        if (!force) {
+            if (paused) {
+                throw MobileScannerError.alreadyPaused
+            } else if (stopped) {
+                throw MobileScannerError.alreadyStopped
+            }
         }
         releaseCamera()
     }
 
     /// Stop scanning for barcodes
-    func stop() throws {
-        if (!paused && stopped) {
+    func stop(force: Bool = false) throws {
+        if (!paused && stopped && !force) {
             throw MobileScannerError.alreadyStopped
         }
         releaseCamera()
@@ -343,7 +345,9 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     }
 
     private func releaseTexture() {
-        registry?.unregisterTexture(textureId)
+        if (textureId != nil) {
+            registry?.unregisterTexture(textureId)
+        }
         textureId = nil
         scanner = nil
     }
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -265,7 +265,7 @@ class _MobileScannerState extends State<MobileScanner>
     // every start is a hot-restart.
     if (kDebugMode) {
       final platformInterface = MobileScannerPlatform.instance;
-      if (platformInterface is MethodChannelMobileScanner) {
+      if (MobileScannerPlatform.instance case final MethodChannelMobileScanner implementation) {
         try {
           await platformInterface.stop(force: true);
         } catch (e) {
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -262,12 +262,13 @@ class _MobileScannerState extends State<MobileScanner>
     // If debug mode is enabled, stop the controller first before starting it.
     // If a hot-restart is initiated, the controller won't be stopped, and because
     // there is no way of knowing if a hot-restart has happened, we must assume
-    // every start is a hot-restart.
+    // every start is a hot-restart. Related issue:
+    // https://github.com/flutter/flutter/issues/10437
     if (kDebugMode) {
-      final platformInterface = MobileScannerPlatform.instance;
-      if (MobileScannerPlatform.instance case final MethodChannelMobileScanner implementation) {
+      if (MobileScannerPlatform.instance
+          case final MethodChannelMobileScanner implementation) {
         try {
-          await platformInterface.stop(force: true);
+          await implementation.stop(force: true);
         } catch (e) {
           // Don't do anything if the controller is already stopped.
           debugPrint('$e');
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: ios/Classes/MobileScanner.swift [hunk 1]
```diff
@@ -343,7 +345,9 @@ public class MobileScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelega
     }
 
     private func releaseTexture() {
-        registry?.unregisterTexture(textureId)
+        if (textureId != nil) {
+            registry?.unregisterTexture(textureId)
```
navaronbracke, line 349: Do we need something similar on MacOS for this check?

lib/src/mobile_scanner.dart
```diff
@@ -257,19 +259,21 @@ class _MobileScannerState extends State<MobileScanner>
   StreamSubscription? _subscription;
 
   Future<void> initMobileScanner() async {
-    // TODO: This will be fixed in another PR
```
navaronbracke, line 260: Can we add a TODO that links to https://github.com/flutter/flutter/issues/10437 when that API is eventually available?

lib/src/mobile_scanner.dart
```diff
@@ -257,19 +259,21 @@ class _MobileScannerState extends State<MobileScanner>
...
     // every start is a hot-restart.
-    // if (kDebugMode) {
-    //   try {
-    //     await controller.stop();
-    //   } catch (e) {
-    //     // Don't do anything if the controller is already stopped.
-    //     debugPrint('$e');
-    //   }
-    // }
+    if (kDebugMode) {
```
navaronbracke, line 266: I guess this is a decent workaround, until we have an actual API to detect this.

lib/src/mobile_scanner.dart
```diff
@@ -257,19 +259,21 @@ class _MobileScannerState extends State<MobileScanner>
...
+      final platformInterface = MobileScannerPlatform.instance;
+      if (platformInterface is MethodChannelMobileScanner) {
```
navaronbracke, line 268: Can we clean this up with an if-case? Just a suggestion though.

```suggestion
      if (MobileScannerPlatform.instance case final MethodChannelMobileScanner implementation) {
```

navaronbracke (approved):
LGTM, although I'd like to see the TODO link added for future reference.

PR #1401, opened 2025-04-21
Fix: remove observer when device is nil

### LINKED ISSUE #1395: mobile_scanner/MobileScanner.swift:341: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value
Hi,

Thanks for maintaining this useful package.

mobile_scanner: ^6.0.8

I have verified the scanner on android and and iOS and i keep on getting below errors:

1. iOS: mobile_scanner/MobileScanner.swift:341: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value
2. Android: MobileScannerException(controllerAlreadyInitialized, The scanner was already started.)

Android error i was able to fix by stopping ad staring scanner but for iOS i cannot even catch this "mobile_scanner/MobileScanner.swift:341: Fatal error: Unexpectedly found nil while implicitly unwrapping an Optional value" as an exception, my emulator just hangs. 

I know scanner is not supported in iOS emulator, but i still want to catch this exception so i can at least handle it. Currently i cannot catch this exception on iOS and therefore it always hangs my emulator app. 

Note, in iOS this exception error occurs ONLY when i run scan second time. First time I get exception (camera not supported) and second time i get fatal error i mentioned above (which does not get caught at startScan method). And I have verified multiple times to ensure resources are disposed before i re-scan but still no luck. I was wondering, if this exception can be captured so at least we can handle it via try catch.

navaronbracke:
Fixes #1395

### DIFF
--- a/macos/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/macos/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -481,6 +481,10 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
             return
         }
 
+        guard let device = device else {
+            return
+        }
+
         captureSession.stopRunning()
         for input in captureSession.inputs {
             captureSession.removeInput(input)
@@ -492,7 +496,7 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
 
         latestBuffer = nil
         self.captureSession = nil
-        device = nil
+        self.device = nil
     }
 
     private func releaseTexture() {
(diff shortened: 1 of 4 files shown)


macos/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -481,6 +481,10 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
             return
         }
 
+        guard let device = device else {
```
navaronbracke, line 484: MacOS also had this bug it seems

navaronbracke:
@juliansteenbakker This will need cherry-picking to dev once merged.

juliansteenbakker:
Thanks for the fix! Will release this at the end of today. I'm also working on a few other small fixes.

PR #1402, opened 2025-04-21
fix: device orientation checked on older and newest sdk

### LINKED ISSUE #18: How to handle device orientation changes
### Discussed in https://github.com/juliansteenbakker/mobile_scanner/discussions/17
Please use the discussion page for discussions and answers.

<div type='discussions-op-text'>

<sup>Originally posted by **juliansteenbakker** February 23, 2022</sup>
Currently rotation is not supported.

We can't check for the right orientation since Flutters own orientation listener doesn't know the difference between landscape left or landscape right. This is why we can only rely on a plugin like [native_device_orientation](https://pub.dev/packages/native_device_orientation).

Another goal of this package is to use as few dependencies as possible. My question right now is if anyone knows a way we can detect for landscape left or landscape right, without the use of native_device_orientation.</div>

### LINKED ISSUE #774: Camera in Landscape mode
is it possible to set mobile_scanner camera in Landscape mode?
I don't see any option to set the Camera in Portrait/Landscape mode, the camera is always in Portrait mode

Thanks for help

### LINKED ISSUE #867: Support Impeller on Android

### LINKED ISSUE #1334: [Orientation] Landscape orientation width not correct
With the latest code i am experiencing issues with the width of the preview in landscape mode. The image seems squashed.

Tested with Android 14 and 10, using samsung s24 ultra and pixel 4a 5G

juliansteenbakker:
This PR fixes all issues regarding device orientation and aspect ratio on all supported sdk levels.

~~TODO: adjust apple implementation~~

Fixes #18
Fixes #774
Fixes #867 
Fixes #1334


lib/src/objects/mobile_scanner_state.dart
```diff
@@ -59,12 +62,18 @@ class MobileScannerState {
   /// The size of the camera output.
   final Size size;
 
+  /// Convenience getter for `previewSize.width / previewSize.height`.
+  double get aspectRatio => size.width / size.height;
```
navaronbracke: Use https://api.flutter.dev/flutter/dart-ui/Size/aspectRatio.html directly, we do not need this getter.

lib/src/mobile_scanner_view_attributes.dart
```diff
@@ -24,4 +26,7 @@ class MobileScannerViewAttributes {
 
   /// The size of the camera output.
   final Size size;
+
+  /// The initial orientation of the device
```
navaronbracke:
```suggestion
  /// The initial orientation of the device.
```

lib/src/mobile_scanner.dart
```diff
@@ -233,18 +234,13 @@ class _MobileScannerState extends State<MobileScanner>
...
               child: SizedBox.fromSize(
                 size: constraints.biggest,
                 child: FittedBox(
                   fit: widget.fit,
-                  child: SizedBox(
-                    width: cameraPreviewSize.width,
-                    height: cameraPreviewSize.height,
-                    child: MobileScannerPlatform.instance.buildCameraView(),
-                  ),
+                  child: CameraPreview(controller,),
```
navaronbracke:
```suggestion
                  child: CameraPreview(controller),
```

lib/src/method_channel/rotated_preview.dart
```diff
@@ -71,60 +71,63 @@ final class _RotatedPreviewState extends State<RotatedPreview> {
...
-      if (!mounted || deviceOrientation == event) {
-        return;
-      }
-
-      setState(() {
-        deviceOrientation = event;
-      });
-    });
+    // Then, subtract the rotation already applied in the CameraPreview widget
+    // (see camera/camera/lib/src/camera_preview.dart).
```
navaronbracke: There is no file "camera/camera/lib/src/camera_preview.dart", so this comment needs updating?

lib/src/method_channel/rotated_preview.dart
```diff
@@ -71,60 +71,63 @@ final class _RotatedPreviewState extends State<RotatedPreview> {
   /// The subscription for the device orientation stream.
   StreamSubscription<Object?>? _deviceOrientationSubscription;
 
+  @override
+  void initState() {
+    deviceOrientation = widget.initialDeviceOrientation;
+    _deviceOrientationSubscription = widget.deviceOrientationStream.listen((
```
navaronbracke: Should we move the method that is passed to `listen` to a private method, so that the indent is better?

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,66 @@
...
+
+  bool _isLandscape() {
+    return <DeviceOrientation>[
+      DeviceOrientation.landscapeLeft,
+      DeviceOrientation.landscapeRight,
+    ].contains(_getApplicableOrientation());
+  }
+
+  int _getQuarterTurns() {
+    final Map<DeviceOrientation, int> turns = <DeviceOrientation, int>{
```
navaronbracke: Can we make this Map a `static const` that is private to the preview?

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,66 @@
...
+            builder: (BuildContext context, MobileScannerState value, Widget? child) {
+              final bool isLand =  _isLandscape();
+              return SizedBox(
+                width: isLand ? value.size.height : value.size.width,
+                height: isLand ? value.size.width : value.size.height,
+                child: _wrapInRotatedBox(child: controller.buildCameraView()),
+              );
+            },
+          )
+        : Container();
```
navaronbracke: Use a `const SizedBox()` instead, per https://dart.dev/tools/linter-rules/avoid_unnecessary_containers

```dart
if (!controller.value.isInitialized) {
  return const SizedBox();
}

// ...
```

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,66 @@
...
+  Widget build(BuildContext context) {
+    return controller.value.isInitialized
+        ? ValueListenableBuilder<MobileScannerState>(
+            valueListenable: controller,
+            builder: (BuildContext context, MobileScannerState value, Widget? child) {
+              final bool isLand =  _isLandscape();
+              return SizedBox(
+                width: isLand ? value.size.height : value.size.width,
+                height: isLand ? value.size.width : value.size.height,
+                child: _wrapInRotatedBox(child: controller.buildCameraView()),
```
navaronbracke: This will rebuild the camera preview even when things like the torch mode changes, which should no affect the preview itself.

navaronbracke (changes requested):

juliansteenbakker:
Aside from the code/doc improvements, can you verify if all rotation issues are fixed by this? @navaronbracke

lib/src/mobile_scanner_preview.dart
juliansteenbakker: I tried changing this to some other implementation, however the `CameraPreview` rebuilds anyway when the `ValueListenableBuilder` updates in the `build` function of `mobile_scanner.dart`

navaronbracke: Right, good point. This is resolved then.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -440,17 +444,17 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
-
-#if os(iOS)
-                // Swap width and height if the image is in portrait mode
-                if orientation == AVCaptureVideoOrientation.portrait || orientation == AVCaptureVideoOrientation.portraitUpsideDown {
-                    let temp = width
-                    width = height
-                    height = temp
-                }
-#endif
+                var width = Double(dimensions.height)
```
navaronbracke: This is flipped in the wrong direction?

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,69 @@
+// Copyright 2013 The Flutter Authors. All rights reserved.
```
navaronbracke: The copyright of this code belongs to you, not the Flutter team! This was probably copy-paste.

navaronbracke (changes requested):

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: Woops yeah, the swift code is not done yet, gonna continue the work tomorrow!

lib/src/mobile_scanner_preview.dart
juliansteenbakker: Yeah i got the base snippet from the camera plugin package, but modified it alot. I will update it tomorrow

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: I added an explanation about why it is flipped in the latest commit.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -277,6 +284,24 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         
         result(nil)
     }
+
+    /// Converts UIDeviceOrientation to correct VideoOrientation
+    func getVideoOrientationForDeviceOrientation(for deviceOrientation: UIDeviceOrientation) -> AVCaptureVideoOrientation {
```
navaronbracke: This code should handle both back & front facing cameras?

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -54,6 +54,9 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     }
 
     var standardZoomFactor: CGFloat = 1
+    
+    var deviceOrientation: UIDeviceOrientation = UIDeviceOrientation.unknown
+    var deviceVideoOrientation: AVCaptureVideoOrientation = AVCaptureVideoOrientation.landscapeRight
```
navaronbracke: Is this a correct default? Isn't it portrait up? (i.e. for MacOS)

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -579,6 +595,32 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         }
 
     }
+    
+    /// Set the device orientation if it differs from previous orientation
+    func setDeviceOrientation(orientation: UIDeviceOrientation) {
+        if (device == nil) {
```
navaronbracke: Can we merge these if statements?

navaronbracke (changes requested):

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: The var isn't actually needed at global level, so i'll remove it.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: I tested it, and works correctly for both facing cameras.

darwin/mobile_scanner/Sources/mobile_scanner/DeviceOrientationStreamHandler.swift
```diff
@@ -0,0 +1,70 @@
+//
+//  DeviceOrientationStreamHandler.swift
+//  Pods
+//
+//  Created by Julian Steenbakker on 21/04/2025.
+//
+
```
navaronbracke, line 7: There's a blank line here

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -1066,3 +1092,62 @@ extension VNBarcodeSymbology {
...
+            return .landscapeLeft
+        case .portraitUpsideDown:
+            return .portraitUpsideDown
+        default:
+            return .portrait
+        }
+    }
+}
+#endif
+
```
navaronbracke: There's some blank lines here

test/mobile_scanner_test.dart
```diff
@@ -1,14 +1,31 @@
...
+      color: Color(0xFF00FF00),
+    );
+  }
+}
+
+class MockMethodChannelMobileScanner extends MethodChannelMobileScanner {
+  @override
+  Future<void> stop({bool force = false}) async {
+    // Do nothing instead of calling platform code
+    debugPrint('Mock stop called with force: $force');
```
navaronbracke: This print is redundant? Shouldn't we use a `verify()` instead?

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,66 @@
...
+  Widget _wrapInRotatedBox({required Widget child}) {
+    if (kIsWeb || defaultTargetPlatform != TargetPlatform.android) {
+      return child;
+    }
+    final turns = _getQuarterTurns();
+    return RotatedBox(quarterTurns: turns, child: child);
+  }
+
+  bool _isLandscape() {
+    return <DeviceOrientation>[
```
navaronbracke: Can we use a `static const Set<DeviceOrientation>` instead?

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,66 @@
...
+  static const Map<DeviceOrientation, int> turns = <DeviceOrientation, int>{
+    DeviceOrientation.portraitUp: 0,
+    DeviceOrientation.landscapeRight: 1,
+    DeviceOrientation.portraitDown: 2,
+    DeviceOrientation.landscapeLeft: 3,
+  };
+
+  @override
+  Widget build(BuildContext context) {
+    return controller.value.isInitialized
```
navaronbracke: Can we use an early return here? Then the `ValueListenableBuilder`'s builder argument will be formatter better, I think.

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,66 @@
...
+        ? ValueListenableBuilder<MobileScannerState>(
+          valueListenable: controller,
+          builder: (
+            BuildContext context,
+            MobileScannerState value,
+            Widget? child,
+          ) {
+            final bool isLandscape = _isLandscape();
+            return SizedBox(
```
navaronbracke: Can we use `SizedBox.fromSize()` together with https://api.flutter.dev/flutter/dart-ui/Size/flipped.html ?

```dart
return SizedBox.fromSize(
   size: isLandscape ? value.size.flipped : value.size,
   child: _wrapInRotatedBox(child: controller.buildCameraView()),
);
```

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,66 @@
...
+          },
+        )
+        : const SizedBox();
+  }
+
+  Widget _wrapInRotatedBox({required Widget child}) {
+    if (kIsWeb || defaultTargetPlatform != TargetPlatform.android) {
+      return child;
+    }
+    final turns = _getQuarterTurns();
```
navaronbracke: Nit: `final int turns = _getQuarterTurns();`

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -326,12 +326,22 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
+
     if (defaultTargetPlatform == TargetPlatform.android) {
       _surfaceProducerDelegate =
           AndroidSurfaceProducerDelegate.fromConfiguration(
             startResult,
             cameraDirection,
           );
+      initialDeviceOrientation =
+          _surfaceProducerDelegate?.initialDeviceOrientation;
+    } else if (defaultTargetPlatform == TargetPlatform.iOS ||
```
navaronbracke: Use a Map pattern with a when clause :)

```suggestion
    } else if (startResult case {'initialDeviceOrientation': final String orientation} when defaultTargetPlatform == TargetPlatform.iOS || defaultTargetPlatform == TargetPlatform.macOS) {
       initialDeviceOrientation = orientation.parseDeviceOrientation();
    }
```

lib/src/mobile_scanner_controller.dart
```diff
@@ -171,6 +177,21 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
 
           value = value.copyWith(zoomScale: zoomScale);
         });
+
+    if (defaultTargetPlatform != TargetPlatform.macOS) {
+      if (MobileScannerPlatform.instance
+          case final MethodChannelMobileScanner implementation) {
```
navaronbracke: Move the outer if to the when clause

```suggestion
          case final MethodChannelMobileScanner implementation when defaultTargetPlatform != TargetPlatform.macOS) {
```

example/lib/screens/mobile_scanner_advanced.dart
```diff
@@ -153,6 +155,8 @@ class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
     }
   }
 
+  bool hide = false;
```
navaronbracke: Can we move this var to be with the others, around line 60?

example/lib/screens/mobile_scanner_advanced.dart
```diff
@@ -189,13 +193,26 @@ class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
...
-              controller = initController();
-              await controller.start();
+              // In order to refresh the controller without leaving the screen
+              // hide the current MobileScanner widget, dispose and remove the
+              // current controller, initiate new controller, and show widget
+              // again while starting the controller.
+              hide = true;
+              setState(() {});
+              // Wait for the UI to settle
+              await Future<void>.delayed(const Duration(milliseconds: 300));
```
navaronbracke: This delay feels like a hack?

example/lib/screens/mobile_scanner_advanced.dart
```diff
@@ -189,13 +193,26 @@ class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
...
 
+              await controller!.dispose();
               if (!mounted) return;
+              controller = null;
 
+              controller = initController();
+              hide = false;
               setState(() {});
+              // Wait for the UI to settle
+              await Future<void>.delayed(const Duration(milliseconds: 300));
```
navaronbracke: Same here

example/lib/screens/mobile_scanner_advanced.dart
```diff
@@ -249,101 +266,111 @@ class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
...
-                crossAxisAlignment: CrossAxisAlignment.start,
+      body:
+          controller == null || hide
+              ? const Placeholder()
+              : Stack(
                 children: [
-                  Expanded(
-                    child: ScannedBarcodeLabel(barcodes: controller.barcodes),
+                  MobileScanner(
+                    // useAppLifecycleState: false, // Only set to false if you want
```
navaronbracke, line 293: Is `useAppLifecycleState` still useful on the widget? Honestly I would move that parameter to the controller anyway. (it feels better to have it there?) Like, if you don't pass a controller, the internal controller defaults to handling the lifecycle, but you can set the flag for the controller you pass in.

navaronbracke (changes requested):

example/lib/screens/mobile_scanner_advanced.dart
juliansteenbakker: It is a hack, a user should not change the controller on the go while `MobileScanner` is active. I added the following docs and moved the hacky code to a separate function:

```dart
  /// This implementation fully disposes and reinitializes the
  /// MobileScannerController every time a setting is changed via the menu.
  ///
  /// This is NOT optimized for production use.
  /// Replacing the controller like this should not happen when MobileScanner is
  /// active. It causes a short visible flicker and can impact user experience.
  ///
  /// The settings should be defined once, or be configurable outside of a
  /// MobileScanner page, not while the MobileScanner is open.
  ///
  /// This is only used here to demonstrate dynamic configuration changes
  /// without restarting the whole app or navigating away from the scanner view.
  Future<void> _reinitializeController() async {
    // Hide the MobileScanner widget temporarily
    setState(() => hide = true);

    // Let the UI settle
    await Future<void>.delayed(const Duration(milliseconds: 300));

    // Dispose of the current controller safely
    await controller?.dispose();
    controller = null;

    // Create a new controller with updated configuration
    controller = initController();

    // Show the scanner again
    setState(() => hide = false);
    await Future<void>.delayed(const Duration(milliseconds: 300));

    // Start scanning again
    await controller?.start();
  }
```

test/mobile_scanner_test.dart
juliansteenbakker: I don't think we should verify that stop is called, since it's only called because of our "hack" to get hot reloading work, and won't be called when kDebugMode is false.

lib/src/mobile_scanner_preview.dart
juliansteenbakker: Maybe we should check out if we want to make that standard behaviour for the lint in #1406

example/lib/screens/mobile_scanner_advanced.dart
juliansteenbakker: This paramater does not do anything when a controller is provided, because `didChangeAppLifecycleState` is only observed when `widget.controller == null` in `initMobileScanner`:

```dart
      WidgetsBinding.instance.addObserver(this);
    }
```

lib/src/mobile_scanner_preview.dart
juliansteenbakker: I fixed it by moving these to an extension, which is cleaner to use imo.

lib/src/mobile_scanner_preview.dart
juliansteenbakker: The analyser has `omit_local_variable_types`, which would remove the type. What's your opinion about this? @navaronbracke

navaronbracke: I'm fine with either as long as the type is obvious from the surrounding code.

test/mobile_scanner_test.dart
navaronbracke: Well, then just removing the debug print is fine.

juliansteenbakker:
@navaronbracke Do you think the PR is ready to merge? I do think so 😄

lib/src/mobile_scanner_preview.dart
```diff
@@ -0,0 +1,61 @@
...
+    if (!controller.value.isInitialized) {
+      return const SizedBox();
+    }
+
+    return ValueListenableBuilder<MobileScannerState>(
+      valueListenable: controller,
+      builder: (BuildContext context, MobileScannerState value, Widget? child) {
+        return SizedBox.fromSize(
+          size:
+              controller.value.deviceOrientation.isLandscape
```
navaronbracke: You can use the value directly here

```suggestion
              value.deviceOrientation.isLandscape
```

navaronbracke (changes requested):
One remark left, but otherwise LGTM.

PR #1406, opened 2025-04-22
chore: migrate from lint to very_good_analysis

juliansteenbakker:
We are using `very_good_analysis` for the example app, but a much more basic linter `lint` in the actual plugin. I propose migrating to `very_good_analysis` since it provides a stricter lint.

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,1 +1,6 @@
 include: package:very_good_analysis/analysis_options.yaml
+
+linter:
+  rules:
+    omit_local_variable_types: false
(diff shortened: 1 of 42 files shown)


juliansteenbakker:
I've opted into [specify_nonobvious_local_variable_types](https://dart.dev/tools/linter-rules/specify_nonobvious_local_variable_types) instead of [omit_local_variable_types](https://dart.dev/tools/linter-rules/omit_local_variable_types), which keeps the code much pretty clear in terms of variable types. It's still experimental, but it looks like it's not going to be removed. https://github.com/dart-lang/sdk/issues/58773

PR #1416, opened 2025-04-28
feat: add error message if controller is used before attached to a mobile scanner widget

### LINKED ISSUE #1415: [mobile_scanner: 7.0.0-rc.2] Camera won't start when using a future before
When adding a future before `_controller = MobileScannerControlle()` the camera will never start.
When i comment the futurue, then it works.

I didn't understand why, i've put the necessary setState...

Tested on Android Emulator and Real devices. Android 15.

```dart
import 'dart:async';

class QrCodeScanner extends StatefulWidget {
  const QrCodeScanner({super.key});

  @override
  State<QrCodeScanner> createState() => _QrCodeScannerState();
}

class _QrCodeScannerState extends State<QrCodeScanner> {
  MobileScannerController? _controller;

  @override
  void initState() {
    start();
    super.initState();
  }

  Future<void> start() async {
    try {
      await Future<void>.delayed(const Duration(seconds: 1)); // comment then it works

      _controller = MobileScannerController(
        autoStart: false,
        detectionSpeed: DetectionSpeed.normal,
        facing: CameraFacing.back,
        cameraResolution: const Size(1920, 2560),
      );

      await _controller?.start();
    } catch (err, stack) {
      debugPrintStack(stackTrace: stack, label: err.toString());
    } finally {
      if (mounted) setState(() {});
    }
  }

  @override
    return ColoredBox(
      color: Colors.black,
      child: Builder(
        builder: (BuildContext context) {
          if (_controller == null) return const SizedBox.expand();

          return ValueListenableBuilder<MobileScannerState>(
            valueListenable: _controller!,
            builder: (BuildContext context, MobileScannerState state, _) {
              return Stack(
                fit: StackFit.expand,
                children: <Widget>[
                  MobileScanner(
                    controller: _controller,
                    errorBuilder: (BuildContext context, _) => const SizedBox(),
                  ),
                ],
              );
            },
          );
        },
      ),
    );
  }
}

```

juliansteenbakker:
Fixes #1415

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -266,6 +266,9 @@ class _MobileScannerState extends State<MobileScanner>
   StreamSubscription<BarcodeCapture>? _subscription;
 
   Future<void> initMobileScanner() async {
+    controller = widget.controller ?? MobileScannerController();
+
+    controller.attach();
     // If debug mode is enabled, stop the controller first before starting it.
     // If a hot-restart is initiated, the controller won't be stopped, and
     // because there is no way of knowing if a hot-restart has happened,
@@ -321,7 +324,6 @@ class _MobileScannerState extends State<MobileScanner>
   @override
   void initState() {
     super.initState();
-    controller = widget.controller ?? MobileScannerController();
     unawaited(initMobileScanner());
   }
 
--- a/lib/src/enums/mobile_scanner_error_code.dart
+++ b/lib/src/enums/mobile_scanner_error_code.dart
@@ -1,5 +1,6 @@
 import 'package:flutter/services.dart';
-import 'package:mobile_scanner/mobile_scanner.dart';
+import 'package:mobile_scanner/src/mobile_scanner.dart' show MobileScanner;
+import 'package:mobile_scanner/src/mobile_scanner_controller.dart';
 
 /// This enum defines the different error codes for the mobile scanner.
 enum MobileScannerErrorCode {
--- a/lib/src/enums/mobile_scanner_error_code.dart
+++ b/lib/src/enums/mobile_scanner_error_code.dart
@@ -1,5 +1,5 @@
 import 'package:flutter/services.dart';
-import 'package:mobile_scanner/src/mobile_scanner_controller.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
 
 /// This enum defines the different error codes for the mobile scanner.
 enum MobileScannerErrorCode {
@@ -28,7 +28,6 @@ enum MobileScannerErrorCode {
   /// Scanning is unsupported on the current device.
   unsupported,
 
-  /// Get the human-readable message for the error code.
   /// The controller is currently initializing.
   ///
   /// This error occurs when [MobileScannerController.start] is called while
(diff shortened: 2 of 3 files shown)


lib/src/mobile_scanner.dart
```diff
@@ -266,6 +266,9 @@ class _MobileScannerState extends State<MobileScanner>
   StreamSubscription<BarcodeCapture>? _subscription;
 
   Future<void> initMobileScanner() async {
+    controller = widget.controller ?? MobileScannerController();
+
+    controller.attach();
```
navaronbracke, line 271: I do not see how this would fix the underlying issue? IIRC the `camera` package doesn't have anything like this.
Is this just some microtask queue weirdness instead?

navaronbracke (changes requested):

navaronbracke:
@juliansteenbakker I think that documenting something like this https://github.com/juliansteenbakker/mobile_scanner/issues/1415#issuecomment-2836611799 is a better thing to do,
having an extra flag just for this kind of thing sounds bad imo.

juliansteenbakker:
@navaronbracke i didn't test it yet, but is this extra check really gotta hurt something? I mean i agree that the situation should not even happen, that a users starts a controller that isn't bound to anything. But if for some reason this happens, it's good to let the user know something is not implemented right. On the other hand, we can let it fail silently, which prevents errors from popping up, but that still doesn't get rid of the bad implementation of the user (calling .start before attached).

navaronbracke:
Maybe we can use an `assert()` ? Although if that doesn't help then I guess this is all we can do.

juliansteenbakker:
I don't think there is any variable we can check to see if the controller is attached, other than to keep track of our own.

lib/src/mobile_scanner.dart
juliansteenbakker: It's the only way to keep track if the controller is attached to the widget or not. The camera package just fails silently, i think it's better to let the user know something is not implemented right.

navaronbracke:
@juliansteenbakker Disposing the controller should have set attached to false? (we're missing a "detach" on dispose) Also, line 339 could have done both a disposed and attached check (we can avoid the second if)

Lastly, what do we do with people that use `autoStart: false` with a manual delay?

juliansteenbakker:
Woops you got me, that should throw a different error than disposed. I'll make a new PR.

I don't see any problems with `autoStart: false`, because the controller will be attached regardeless of that parameter. The user just has to make sure they don't call `.start` before the controller is attached to the widget.

juliansteenbakker:
I fixed it in #1421. If everything is fine for you, i'm going to release the stable 7 version @navaronbracke

navaronbracke:
#1421 is LGTM, but can you also address the following? This is mainly for consistency to be honest.

> Disposing the controller should have set attached to false? (we're missing a "detach" on dispose)

juliansteenbakker:
> #1421 is LGTM, but can you also address the following? This is mainly for consistency to be honest.
> 
> > Disposing the controller should have set attached to false? (we're missing a "detach" on dispose)

Fixed here: https://github.com/juliansteenbakker/mobile_scanner/commit/0a1007c32bc08b63021967b94732ab6f0a300d41

PR #1431, opened 2025-05-12
Update README.md

atnpem:
avoid controllerNotAttached exception


navaronbracke:
I would like to see this added in https://github.com/juliansteenbakker/mobile_scanner/pull/1432 instead, taking into account my remark about the post frame callback.

PR #1432, opened 2025-05-12
fix: update mobile_scanner_advanced.dart

atnpem:
avoid controllerNotAttached exception

### DIFF
--- a/example/lib/screens/mobile_scanner_advanced.dart
+++ b/example/lib/screens/mobile_scanner_advanced.dart
@@ -40,7 +40,7 @@ class MobileScannerAdvanced extends StatefulWidget {
   State<MobileScannerAdvanced> createState() => _MobileScannerAdvancedState();
 }
 
-class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> with WidgetsBindingObserver {
+class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
   MobileScannerController? controller;
 
   // A scan window does work on web, but not the overlay to preview the scan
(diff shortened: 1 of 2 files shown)


example/lib/screens/mobile_scanner_advanced.dart
```diff
@@ -40,7 +40,7 @@ class MobileScannerAdvanced extends StatefulWidget {
   State<MobileScannerAdvanced> createState() => _MobileScannerAdvancedState();
 }
 
-class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
+class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> with WidgetsBindingObserver {
```
navaronbracke: This WidgetsBindingObserver is not needed?

example/lib/screens/mobile_scanner_advanced.dart
```diff
@@ -80,12 +80,18 @@ class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
   @override
   void initState() {
     super.initState();
+    WidgetsBinding.instance.addObserver(this);
     controller = initController();
-    unawaited(controller!.start());
+    WidgetsBinding.instance.addPostFrameCallback((_) async {
```
navaronbracke: Instead of a postframecallback, I would just use a `Future.value()` or `Future.microtask()`, as those will complete in the next event loop frame

navaronbracke (changes requested):

example/lib/screens/mobile_scanner_advanced.dart
atnpem: I replaced by `Future.microtask(() => unawaited(controller?.start()));`

navaronbracke:
This will no longer be needed.

PR #1456, opened 2025-06-04
feat: tap to focus, initial zoom and more lenses

juliansteenbakker:
This pr adds:
- Tap to Focus
- initial zoom selection
- more lens options for ios

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -348,6 +348,7 @@ class MobileScanner(
         detectionTimeout: Long,
         cameraResolutionWanted: Size?,
         invertImage: Boolean,
+        initialZoom: Double,
     ) {
         this.detectionSpeed = detectionSpeed
         this.detectionTimeout = detectionTimeout
@@ -475,6 +476,18 @@ class MobileScanner(
                 // Enable torch if provided
                 if (it.cameraInfo.hasFlashUnit()) {
                     it.cameraControl.enableTorch(torch)
+                }
+
+                try {
+                    if (initialZoom in 0.0..1.0) {
+                        it.cameraControl.setLinearZoom(initialZoom.toFloat())
+                    } else {
+                        it.cameraControl.setZoomRatio(initialZoom.toFloat())
+                    }
+                } catch (e: Exception) {
+                    mobileScannerErrorCallback(ZoomNotInRange())
+
+                    return@addListener
                 }
             }
 
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -10,6 +10,7 @@ import android.graphics.Matrix
 import android.graphics.Paint
 import android.graphics.Rect
 import android.hardware.display.DisplayManager
+import android.icu.util.TimeUnit
 import android.net.Uri
 import android.os.Handler
 import android.os.Looper
@@ -23,9 +24,13 @@ import androidx.camera.core.CameraSelector
 import androidx.camera.core.CameraXConfig
 import androidx.camera.core.ExperimentalGetImage
 import androidx.camera.core.ExperimentalLensFacing
+import androidx.camera.core.FocusMeteringAction
 import androidx.camera.core.ImageAnalysis
 import androidx.camera.core.ImageProxy
+import androidx.camera.core.MeteringPoint
+import androidx.camera.core.MeteringPointFactory
 import androidx.camera.core.Preview
+import androidx.camera.core.SurfaceOrientedMeteringPointFactory
 import androidx.camera.core.SurfaceRequest
 import androidx.camera.core.TorchState
 import androidx.camera.core.resolutionselector.ResolutionSelector
@@ -699,6 +704,24 @@ class MobileScanner(
     fun resetScale() {
         if (camera == null) throw ZoomWhenStopped()
         camera?.cameraControl?.setZoomRatio(1f)
+    }
+
+    fun setFocus(x: Float, y: Float) {
+        val cam = camera ?: throw ZoomWhenStopped()
+
+        // Ensure x,y are normalized (0f..1f)
+        if (x !in 0f..1f || y !in 0f..1f) {
+            throw IllegalArgumentException("Focus coordinates must be between 0.0 and 1.0")
+        }
+
+        val factory: MeteringPointFactory = SurfaceOrientedMeteringPointFactory(1f, 1f)
+        val afPoint: MeteringPoint = factory.createPoint(x, y)
+
+        val action = FocusMeteringAction.Builder(afPoint, FocusMeteringAction.FLAG_AF)
+//            .setAutoCancelDuration(3, TimeUnit.SECONDS) // optional
+            .build()
+
+        cam.cameraControl.startFocusAndMetering(action)
     }
 
     /**
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -10,7 +10,6 @@ import android.graphics.Matrix
 import android.graphics.Paint
 import android.graphics.Rect
 import android.hardware.display.DisplayManager
-import android.icu.util.TimeUnit
 import android.net.Uri
 import android.os.Handler
 import android.os.Looper
@@ -731,7 +730,6 @@ class MobileScanner(
         val afPoint: MeteringPoint = factory.createPoint(x, y)
 
         val action = FocusMeteringAction.Builder(afPoint, FocusMeteringAction.FLAG_AF)
-//            .setAutoCancelDuration(3, TimeUnit.SECONDS) // optional
             .build()
 
         cam.cameraControl.startFocusAndMetering(action)
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerErrorCodes.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerErrorCodes.swift
@@ -24,7 +24,8 @@ struct MobileScannerErrorCodes {
     static let GENERIC_ERROR_MESSAGE = "An unknown error occurred."
     // This message is used with the 'GENERIC_ERROR' error code.
     static let INVALID_ZOOM_SCALE_ERROR_MESSAGE = "The zoom scale should be between 0 and 1 (both inclusive)"
-    static let INVALID_FOCUS_POINT = "The focus needs both X and Y coordinates to be between 0 and 1."
+    static let INVALID_FOCUS_POINT = "MOBILE_SCANNER_INVALID_FOCUS_POINT"
+    static let INVALID_FOCUS_POINT_MESSAGE = "The focus coordinates are not valid."
     static let NO_CAMERA_ERROR = "MOBILE_SCANNER_NO_CAMERA_ERROR"
     static let NO_CAMERA_ERROR_MESSAGE = "No cameras available."
     static let SET_SCALE_WHEN_STOPPED_ERROR = "MOBILE_SCANNER_SET_SCALE_WHEN_STOPPED_ERROR"
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerErrorCodes.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerErrorCodes.swift
@@ -24,7 +24,7 @@ struct MobileScannerErrorCodes {
     static let GENERIC_ERROR_MESSAGE = "An unknown error occurred."
     // This message is used with the 'GENERIC_ERROR' error code.
     static let INVALID_ZOOM_SCALE_ERROR_MESSAGE = "The zoom scale should be between 0 and 1 (both inclusive)"
-    static let INVALID_FOCUS_POINT = "The focus needs both X and Y coordinates between 0 and 1."
+    static let INVALID_FOCUS_POINT = "The focus needs both X and Y coordinates to be between 0 and 1."
     static let NO_CAMERA_ERROR = "MOBILE_SCANNER_NO_CAMERA_ERROR"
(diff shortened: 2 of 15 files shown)


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -360,9 +363,17 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+        }
+#else
         if #available(macOS 10.15, *) {
             device = AVCaptureDevice.DiscoverySession(deviceTypes: [.builtInWideAngleCamera], mediaType: .video, position: position).devices.first
-        } else {
+        }
+#endif
+        
+        if (device == nil) {
             device = AVCaptureDevice.devices(for: .video).filter({$0.position == position}).first
```
SamBergeron, line 377: In our production app, we've noticed that using 
```
device = AVCaptureDevice.default(for: .video)
```
instead of the first in the devices list yields a better result when the camera isn't found in the list returned from the `DiscoverySession` if you want to try that out here

SamBergeron (approved):

AmirAshvins:
@juliansteenbakker may I ask when this feature will be merged with the master branch? I am experiencing issues regarding tap to focus and proper camera selection, specially on pro Iphones (which you have implemented a fix for it in this PR)

Cheers,

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: Yes, it's possible to add this, however the current implementation filters the list based on the selected position. `.default` would not use this.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -701,6 +719,24 @@ class MobileScanner(
...
+        // Ensure x,y are normalized (0f..1f)
+        if (x !in 0f..1f || y !in 0f..1f) {
+            throw IllegalArgumentException("Focus coordinates must be between 0.0 and 1.0")
+        }
+
+        val factory: MeteringPointFactory = SurfaceOrientedMeteringPointFactory(1f, 1f)
+        val afPoint: MeteringPoint = factory.createPoint(x, y)
+
+        val action = FocusMeteringAction.Builder(afPoint, FocusMeteringAction.FLAG_AF)
+//            .setAutoCancelDuration(3, TimeUnit.SECONDS) // optional
```
navaronbracke: Is this still needed? (along with the TimeUnit import above)

lib/src/objects/start_options.dart
```diff
@@ -51,6 +52,13 @@ class StartOptions {
   /// option.
   final bool autoZoom;
 
+  /// If set this sets the zoom scale factor when the camera launches
```
navaronbracke: A suggestion for the entire doc comment:

```suggestion
  /// The initial zoom scale factor for the camera.
  ///
  /// Currently only supported on iOS and Android.
```

lib/src/mobile_scanner_platform_interface.dart
```diff
@@ -83,6 +83,14 @@ abstract class MobileScannerPlatform extends PlatformInterface {
     throw UnimplementedError('setZoomScale() has not been implemented.');
   }
 
+  /// Set the focus position for the camera.
```
navaronbracke, line 86: A suggestion for the doc comment:

```suggestion
  /// Set the focus position for the camera.
  ///
  /// The provided point should be in the range `(0,0) - (1,1)`, where `(0,0)` is the top left
  /// and `(1,1)` is the bottom right.
```

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -586,6 +608,45 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+        
+            result(nil)
+        } catch {
+            result(FlutterError(code: MobileScannerErrorCodes.GENERIC_ERROR,
+                                message: MobileScannerErrorCodes.GENERIC_ERROR_MESSAGE,
+                                details: nil))
+        }
+    }
+
+
```
navaronbracke: There are some extra blank lines here.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerErrorCodes.swift
```diff
@@ -24,6 +24,7 @@ struct MobileScannerErrorCodes {
     static let GENERIC_ERROR_MESSAGE = "An unknown error occurred."
     // This message is used with the 'GENERIC_ERROR' error code.
     static let INVALID_ZOOM_SCALE_ERROR_MESSAGE = "The zoom scale should be between 0 and 1 (both inclusive)"
+    static let INVALID_FOCUS_POINT = "The focus needs both X and Y coordinates to be between 0 and 1."
```
navaronbracke: This is inconsistent with Android? `static let INVALID_FOCUS_POINT = "MOBILE_SCANNER_INVALID_FOCUS_POINT"` is missing and this line should use `INVALID_FOCUS_POINT_ERROR_MESSAGE` ?

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -586,6 +608,45 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
 
     }
     
+    private func setFocus(_ call: FlutterMethodCall, _ result: @escaping FlutterResult) {
+        guard let args = call.arguments as? [String: Any],
+                  let dx = args["dx"] as? CGFloat,
+                  let dy = args["dy"] as? CGFloat else {
+                result(FlutterError(code: MobileScannerErrorCodes.GENERIC_ERROR,
+                                    message: MobileScannerErrorCodes.INVALID_FOCUS_POINT,
+                                    details: "The focus coordinates are not valid."))
```
navaronbracke: `details` can be nil here? The guidance will be in the message?

example/lib/screens/mobile_scanner_advanced.dart
```diff
@@ -302,67 +303,77 @@ class _MobileScannerAdvancedState extends State<MobileScannerAdvanced> {
...
+                    tapToFocus: true,
                     controller: controller,
                     errorBuilder: (context, error) {
                       return ScannerErrorWidget(error: error);
                     },
                     fit: boxFit,
                   ),
                   if (useBarcodeOverlay)
-                    BarcodeOverlay(controller: controller!, boxFit: boxFit),
+                    IgnorePointer( // Needed for tapToFocus
```
navaronbracke: Nit: Can we put the comment _above_ `IgnorePointer(`, so that is does not move on formatting changes? (same for the case below)

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -1,4 +1,5 @@
 import 'dart:async';
+import 'dart:io';
```
navaronbracke: This import is not needed? IIRC `TargetPlatform` is available in the foundation library.

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -275,6 +276,20 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
     await methodChannel.invokeMethod<void>('setScale', zoomScale);
   }
 
+  @override
+  Future<void> setFocusPoint(Offset position) async {
+    final Map<String, dynamic> params = <String, dynamic>{
```
navaronbracke:
```suggestion
    final Map<String, Object?> params = <String, Object?>{
```

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -275,6 +276,20 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
+      'dx': position.dx,
+      'dy': position.dy,
+    };
+    if (defaultTargetPlatform == TargetPlatform.iOS ||
```
navaronbracke: Can we use an early return here?

lib/src/mobile_scanner.dart
```diff
@@ -136,6 +137,18 @@ class MobileScanner extends StatefulWidget {
   /// Defaults to true.
   final bool useAppLifecycleState;
 
+  /// Enables or disables tap-to-focus functionality on the camera preview.
```
navaronbracke, line 140:
```suggestion
  /// Enables or disables tap-to-focus functionality on the camera preview.
  ///
  /// When set to `true`, the camera will adjust focus automatically when the
  /// user taps on a specific point in the preview view.
  /// When set to `false`, tap gestures are ignored, and the camera remains in continuous autofocus mode.
  ///
  /// If this is `true`, the preview (or part of it) should be able to receive gestures,
  /// as other widgets overlaid over the preview will prevent it to receive gestures in those areas.
  ///
  /// Defaults to false and is only supported on iOS and Android.
```

lib/src/mobile_scanner.dart
```diff
@@ -239,12 +252,22 @@ class _MobileScannerState extends State<MobileScanner>
               constraints,
             );
 
-            final Widget scannerWidget = ClipRect(
-              child: SizedBox.fromSize(
-                size: constraints.biggest,
-                child: FittedBox(
-                  fit: widget.fit,
-                  child: CameraPreview(controller),
+            final Widget scannerWidget = GestureDetector(
```
navaronbracke: We should probably not wrap the widget in a GestureDetector if `tapToFocus` is false?

lib/src/mobile_scanner.dart
```diff
@@ -239,12 +252,22 @@ class _MobileScannerState extends State<MobileScanner>
...
+              onTapUp: (details) {
+                final Size size = MediaQuery.of(context).size;
```
navaronbracke: Do we need the _display_ size or the widget layout size here?

If we do need the display size, prefer using `MediaQuery.sizeOf(context)` instead, which only rebuilds when the size changes. (I would wrap the inner widget in a `Builder`, so only the smallest widget rebuilds when that happens)

Otherwise we can use the constraints from the layout builder.

lib/src/mobile_scanner_controller.dart
```diff
@@ -112,6 +113,13 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
   /// Only supported on Android.
   final bool autoZoom;
 
+  /// If set this sets the zoom scale factor when the camera launches
```
navaronbracke: Suggestion for the doc comment:

```suggestion
  /// The initial zoom scale for the camera.
  ///
  /// Defaults to no initial zoom and is only supported on iOS, MacOS and Android.
```

lib/src/mobile_scanner_controller.dart
```diff
@@ -33,6 +33,7 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
     this.torchEnabled = false,
     this.invertImage = false,
     this.autoZoom = false,
+    this.initialZoom = 0,
```
navaronbracke: Isn't this default value wrong? On iOS/MacOS the previous default value was a standard zoom factor of 1, not 0. I'm not sure what Android uses as default (zero?)

lib/src/mobile_scanner_controller.dart
```diff
@@ -318,6 +326,27 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
     await MobileScannerPlatform.instance.setZoomScale(clampedZoomScale);
   }
 
+  /// Set the focus point for the camera.
+  ///
+  /// The [position] must be between an offset with X, Y coordinates.
```
navaronbracke:
```suggestion
  /// The [position] must be a point between `0,0` and `1,1`, both inclusive.
```

lib/src/mobile_scanner_controller.dart
```diff
@@ -318,6 +326,27 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+  ///
+  /// Does nothing if the camera is not running.
+  Future<void> setFocusPoint(Offset position) async {
+    _throwIfNotInitialized();
+
+    if (!value.isRunning) {
+      return;
+    }
+
+    // Clamp coordinates to make sure they are between 1 and 0
```
navaronbracke: This comment is a bit redundant with the doc comment above

navaronbracke (changes requested):

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
juliansteenbakker: Maybe good to implement at a later stage, for now i'll remove it

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerErrorCodes.swift
juliansteenbakker: I've changed it to be consistent with android

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
juliansteenbakker: Yes that seems redundant

lib/src/method_channel/mobile_scanner_method_channel.dart
juliansteenbakker: Yes good catch, leftover from old implementation

lib/src/mobile_scanner_controller.dart
juliansteenbakker: It should be 1 indeed

lib/src/mobile_scanner.dart
juliansteenbakker: I think we need the displaysize due to `globalPosition`. Maybe we can switch to localPosition with layout builder, but for now i changed it to `MediaQuery.sizeOf(context)`

navaronbracke (approved):
LGTM! (I adjusted some documentation nits and fixed an issue with the missing Builder manually)

Issue #1471, opened 2025-07-02
Error when starting app on macos


iamsami101:
I think this issue came after I updated flutter but im not sure but. Whenever I try to run the app on macos, it gives me the following output and prevents me from running the app:

```shell
Users/sami/.pub-cache/hosted/pub.dev/mobile_scanner-7.0.1/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift:470:49: error: expected initial value after '='
                    let cameraDirection: Int? = switch(device.position) {
                                                ^
/Users/sami/.pub-cache/hosted/pub.dev/mobile_scanner-7.0.1/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift:470:48: error: consecutive statements on a line must be separated by ';'
                    let cameraDirection: Int? = switch(device.position) {
                                               ^
                                               ;
/Users/sami/.pub-cache/hosted/pub.dev/mobile_scanner-7.0.1/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift:1007:34: error: cannot find 'payloadData' in scope
            if let payloadData = payloadData {
                                 ^~~~~~~~~~~
```

`flutter doctor -v`:
```shell
[✓] Flutter (Channel stable, 3.32.5, on macOS 12.7.6 21H1320 darwin-x64, locale en-GB) [761ms]
    • Flutter version 3.32.5 on channel stable at /Users/sami/Documents/flutter
    • Upstream repository https://github.com/flutter/flutter.git
    • Framework revision fcf2c11572 (8 days ago), 2025-06-24 11:44:07 -0700
    • Engine revision dd93de6fb1
    • Dart version 3.8.1
    • DevTools version 2.45.1

[✓] Android toolchain - develop for Android devices (Android SDK version 35.0.0) [4.0s]
    • Android SDK at /Users/sami/Library/Android/sdk
    • Platform android-35, build-tools 35.0.0
    • ANDROID_HOME = /Users/sami/Library/Android/sdk
    • Java binary at: /Applications/Android Studio.app/Contents/jbr/Contents/Home/bin/java
      This is the JDK bundled with the latest Android Studio installation on this machine.
      To manually set the JDK path, use: `flutter config --jdk-dir="path/to/jdk"`.
    • Java version OpenJDK Runtime Environment (build 21.0.6+-13391695-b895.109)
    • All Android licenses accepted.

[...]

PR #1472, opened 2025-07-02
fix: Update MobileScannerPlugin.swift

iamsami101:
Fixes issue [1471](https://github.com/juliansteenbakker/mobile_scanner/issues/1471) on macos 12.

this is kind of a workaround and the debug consoles screams at you if you try to run the app but it runs and doesn't show any other signs of errors.

most probable cause of this was that I was using XCode 14 because I was on macos 12 monterey. the API couldnt find some of the functions when compiling the app for macos.

### DIFF
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -464,14 +464,19 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                 let size = ["width": Double(dimensions.width), "height": Double(dimensions.height)]
 #endif
                 // Return the result on the main thread after the session starts.
-                let answer: [String : Any?]
+                var answer: [String : Any?]
 
                 if let device = self.device {
-                    let cameraDirection: Int? = switch(device.position) {
-                        case .back: 1
-                        case .unspecified: nil
-                        case .front: 0
-                        @unknown default: nil
+                    var cameraDirection: Int?
+                    switch(device.position) {
+                        case .back:
+                            cameraDirection = 1
+                        case .front:
+                            cameraDirection = 0
+                        case .unspecified:
+                            cameraDirection = nil
+                        @unknown default:
+                            cameraDirection = nil
                     }
                     
                     answer = [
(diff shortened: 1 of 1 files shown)


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -1003,11 +1008,11 @@ extension VNBarcodeObservation {
         let height = distanceBetween(adjustedTopLeft, adjustedBottomLeft) * CGFloat(imageHeight)
         var rawBytes: FlutterStandardTypedData? = nil
         
-        if #available(iOS 17.0, macOS 14.0, *) {
-            if let payloadData = payloadData {
-                rawBytes = FlutterStandardTypedData(bytes: payloadData)
-            }
-        }
+        // if #available(iOS 17.0, macOS 14.0, *) {
```
navaronbracke, line 1011: Why is this commented out?

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -464,14 +464,19 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
-                let answer: [String : Any?]
+                var answer: [String : Any?]
 
                 if let device = self.device {
-                    let cameraDirection: Int? = switch(device.position) {
-                        case .back: 1
-                        case .unspecified: nil
-                        case .front: 0
-                        @unknown default: nil
+                    var cameraDirection: Int?
```
navaronbracke, line 470: What Swift version are you running with? I'd like to know why this would cause an issue, are switch assignments problematic on older versions of MacOS?

navaronbracke (changes requested):

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
iamsami101: im using swift 5.7, on xcode 14.2

navaronbracke: Does this happen on XCode 14.3 ? This does not happen on XCode 15 at least, so maybe you should just upgrade instead? Apple requires you to build against the latest iOS SDK anyway

juliansteenbakker:
I'm closing this since we do not need to support older versions of xcode.

Issue #1262, opened 2024-12-16
takePicture via MobileScannerController


levmorgot:
Hi, please tell me if I can get a frame from the camera (by pressing a button) in order to analyze it for the information I need?
I want to be able to use your MobileScannerController to get a frame file from the camera.

navaronbracke:
This is currently not supported in `mobile_scanner` and I think it is unlikely to be added to the plugin.

If this is the only thing you need to do, you can probably use https://pub.dev/documentation/camera/latest/camera/CameraController/startImageStream.html

to capture frames, and pass those to `analyzeImage()` instead.

dynamicpace-admin:
I hade some similar thing I wanted to do alongside mobile_scanner. But I manage to do so by using the mobile scanner camera and grabs a screenshot from the screen and uses that.

PR #1479, opened 2025-07-10
feat: support takePicture.

yujune:
Introduced the `takePicture` function in `MobileScannerController` to allow pictures to be captured programmatically.

- Related Issue: https://github.com/juliansteenbakker/mobile_scanner/issues/1262


juliansteenbakker:
@navaronbracke what are your thoughts on this? I always thought that taking pictures for a QR/Barcode scanner was not in the scope, but it's a preaty clean implementation, so i'm in doubt wether to add this or not...

navaronbracke:
Honestly I don't think we should do this. The main use case is a camera preview for live detection. The `analyzeImage()` use case (from a file) is just a nice to have on top of it.

As mentioned in https://github.com/juliansteenbakker/mobile_scanner/issues/1262#issuecomment-2545485646
if people really want this, they should be using the `camera` package, which has an API for this.

juliansteenbakker:
I agree. For now i'm closing this PR as it is out of the scope. However, this may change in the future if more people request this feature.

PR #1481, opened 2025-07-12
feat(android): add photo capture functionality

Mazen-Almortada:
This pull request introduces a new feature that allows developers to capture a photo while the mobile_scanner is active. This functionality is currently implemented for Android only.

A new capturePhoto() method has been added to the MobileScannerController, which can be called to trigger a photo capture. The method returns a Future<String?> containing the path to the saved image file on the device.


juliansteenbakker:
Thank you for your work, however I think #1479 is a more complete version of this implementation

PR #1502, opened 2025-08-18
fix: set best camera quality on darwin

### LINKED ISSUE #1229: Camera FPS
Hi! Thanks for your plugin!

When I use class `MobileScanner`, the camera view seems to run at 30 (or even lower) fps. Is this because my app runs on the debug version, or is it because of the plugin?
I was using an iPhone 13 with a full battery percentage.
How can I make it so it runs at max fps possible?

The code I was using is `example/lib/barcode_scanner_simple.dart`. [()](https://github.com/juliansteenbakker/mobile_scanner/blob/master/example/lib/barcode_scanner_simple.dart)

kevtechi:
I realize that I could let the user decide. 

But:
- this would be only for macos/ios
- 60 FPS default makes the most sense

resolves: https://github.com/juliansteenbakker/mobile_scanner/issues/1229 

Been running this in prod for a few weeks without issues.


navaronbracke (changes requested):
Overall change is LGTM, I assume that `high` maps to any of these, whichever is available? https://developer.apple.com/documentation/avfoundation/avcapturesession/preset#High-Definition

This is going to need a changelog update + minor version bump

kevtechi:
> Overall change is LGTM, I assume that `high` maps to any of these, whichever is available? https://developer.apple.com/documentation/avfoundation/avcapturesession/preset#High-Definition
> 
> 
> 
> This is going to need a changelog update + minor version bump

It was more about the FPS than quality. Did not check for image quality changes.

juliansteenbakker:
> This is going to need a changelog update + minor version bump

I'll add this on develop, lgtm!

PR #1517, opened 2025-09-12
feat(web): auto-select best back camera, apply close-focus, persist deviceId

NicoBallaman:
Web scanner: choose best videoinput (prefer back-facing, continuous AF, shorter min focus distance, higher max resolution); apply focus constraints (focusMode continuous/single-shot/manual with min focusDistance) when supported; remember chosen deviceId in localStorage; fallback to facingMode/default if deviceId fails; remove temporary debug logs.

### DIFF
--- a/lib/src/web/mobile_scanner_web.dart
+++ b/lib/src/web/mobile_scanner_web.dart
@@ -1,5 +1,6 @@
 import 'dart:async';
 import 'dart:js_interop';
+import 'dart:js_util' as js_util;
 import 'dart:ui_web' as ui_web;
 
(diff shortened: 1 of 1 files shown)


navaronbracke:
@juliansteenbakker This seems like its still very much a WIP, there's a lot of random bits in the PR.
I think we should look at improving this ourselves instead. (Also, the "scoring" bit seems very weird and unnecessary)

juliansteenbakker:
I have made an improved version, based on this PR in #1662. I will continue working there, so i'll close this pr. Thank you for the initial improvements!

PR #1528, opened 2025-09-22
feat: Flutter view barcode corners

### LINKED ISSUE #1450: Barcode.corners appear to be in different coordinate system than flutter view
I am getting Barcodes but unfortunately it looks like `corners` are using a different coordinate system, with Y maxxing out around 1400 pixels while the device height in flutter view pixels is 874. I tried dividing by the device pixel ratio and that didn't quite get it.

is there a published example that shows how to draw bounding boxes around a barcode?

navaronbracke:
Fixes #1450

### DIFF
--- a/example/lib/widgets/dialogs/box_fit_dialog.dart
+++ b/example/lib/widgets/dialogs/box_fit_dialog.dart
@@ -46,10 +46,7 @@ class _BoxFitDialogState extends State<BoxFitDialog> {
             child: Column(
               children: [
                 for (final fit in BoxFit.values)
-                  RadioListTile<BoxFit>(
-                    title: Text(fit.name),
-                    value: fit,
-                  ),
+                  RadioListTile<BoxFit>(title: Text(fit.name), value: fit),
               ],
             ),
           ),
--- a/test/barcode_test.dart
+++ b/test/barcode_test.dart
@@ -1,0 +1,71 @@
+import 'dart:ui';
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mobile_scanner/src/objects/barcode.dart';
+
+void main() {
+  group('scaleCorners', () {
+    test('returns empty list if barcode has no size or corners', () {
+      const Barcode barcode = Barcode();
+
+      expect(barcode.scaleCorners(const Size(100, 100)), isEmpty);
+    });
+
+    test('returns empty list if barcode has no corners', () {
+      const Barcode barcode = Barcode(size: Size(200, 200));
+
+      expect(barcode.scaleCorners(const Size(100, 100)), isEmpty);
+    });
+
+    test('returns zeroed corners if barcode has corners but no size', () {
+      const Barcode barcode = Barcode(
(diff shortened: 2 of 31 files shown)


example/lib/widgets/dialogs/box_fit_dialog.dart
```diff
@@ -46,10 +46,7 @@ class _BoxFitDialogState extends State<BoxFitDialog> {
             child: Column(
               children: [
                 for (final fit in BoxFit.values)
-                  RadioListTile<BoxFit>(
-                    title: Text(fit.name),
-                    value: fit,
-                  ),
+                  RadioListTile<BoxFit>(title: Text(fit.name), value: fit),
```
navaronbracke, line 49: This was a small formatting change from a previous PR, but I only fixed the newlines in that PR, in the Github UI.

juliansteenbakker:
Great to see some work done again! I'll pick up where i left on some other PR's later today.

navaronbracke:
@juliansteenbakker This PR includes some doc updates to make the distinction a bit better in code.

navaronbracke:
Yeah, it has been a while :P

juliansteenbakker:
> Yeah, it has been a while :P

My son was born last month, so had some other priorities on my head, but slowly i get more and more time to do other stuff. 

> @juliansteenbakker This PR includes some doc updates to make the distinction a bit better in code.

Looks good to me! LGTM 😄

test/barcode_test.dart
nespjin, line 1: How to know target size for every barcode?

nespjin: Why not use BarcodeCapture.size and previewSize of Camera in Flutter?

nespjin: Can you provide a example for drawing barcode overlay box?

navaronbracke: > Why not use BarcodeCapture.size and previewSize of Camera in Flutter?

The target size (i.e. widget size of the camera preview) is up to the client to provide (we don't have their widget tree).

Do you also want this scale method for the `BarcodeCapture.size` ? (to scale _that_ to widget size as well)

navaronbracke: cc @juliansteenbakker if I should add that also here

navaronbracke: > Can you provide a example for drawing barcode overlay box?

We have an example in the example app, that uses the `BarcodeOverlay` widget (which is provided by this plugin)

juliansteenbakker: @navaronbracke I don't think we need to scale `BarcodeCapture.size` since it's an representation of the capture, so i don't really see the usecase here. I think it's good as is.

juliansteenbakker (approved):
Looks good to me

navaronbracke:
@juliansteenbakker Maybe we can release this along with the focus changes?

navaronbracke:
@juliansteenbakker This can probably also go into the next version (I reverted the version change in this PR, since there's now more than bugfixes)

Issue #1357, opened 2025-03-08
How do I get the bytes of a QR code?


coder0xff:
Using qr_flutter, I made a QR code containing bytes

```dart
QrCode.fromUint8List(
  data: Uint8List.fromList([72, 101, 108, 108, 111, 32, 119, 111, 114, 108, 100]),
  errorCorrectLevel: 0
)
```
These bytes are the UTF-8 encoding of `Hello world`.

![Image](https://github.com/user-attachments/assets/4196c9dd-8d69-40cf-874c-22ee0acdb4f8)

When I scan it with mobile_scanner, I see that rawValue is "Hello world" as expected. However, rawBytes is:

64 180 134 86 198 198 242 7 118 247 38 198 64 236 17 236

As someone who doesn't know much about QR codes, I can't make heads or tails of this. Issue #947 confirms my ignorance.

I found this issue #290, but it was closed before revealing how to get bytes, as rawValue is string valued. I can do `latin1.encode(captured.barcodes[0].rawValue!)` to get bytes, but I'm unsure how reliable that will be if `rawValue` isn't a valid UTF-8 string to begin with.

Is there a way to get the original bytes?

CordMemescape:
Hi there! I looked into the question about the relationship between `rawBytes` and `rawValue` in the mobile_scanner package.

The key thing to understand is that `rawBytes` and `rawValue` serve different purposes:

- `rawValue` gives you the actual decoded content ("Hello world") that was encoded in the QR code - this is what most applications need
- `rawBytes` contains the raw binary data structure of the QR code as returned by CameraX or AVFoundation itself before final decoding - these bytes include QR format information, headers, and error correction codes

These two aren't directly convertible using simple encoding/decoding because rawBytes represents the QR code's internal structure rather than just an encoded version of your text.

**For most use cases, you probably want to use `rawValue` directly.**

If you have a specific need that requires working with the raw QR code data structure, could you share more details about what you're trying to accomplish? There might be a more straightforward approach to your underlying problem.

For example:
- If you need to verify data integrity: Use the `rawValue` with your own hash/checksum
- If you need custom encoding: Apply your encoding to the string before creating the QR code
- If you need binary data: Work with `Uint8List.fromList(utf8.encode(barcode.rawValue))` to get UTF-8 bytes

The internal transformation from content to QR code structure is complex (involves mode indicators, character counts, error correction, etc.) and is typically handled by the QR libraries for you.

[...]

Issue #1317, opened 2025-02-04
[ios][macos] The bytes of the `payloadData` can contain header / padding bytes


navaronbracke:
As mentioned in https://github.com/juliansteenbakker/mobile_scanner/issues/1315#issuecomment-2635255881
the recently added `payloadData` support, introduces a new bug: Previously MLKit on iOS removed the header / padding bytes for us.
But with the Vision API that isn't the case.

We should restore this behavior and also add a caveat in the README for users that use the currently broken implementation.

cc @disptr

juliansteenbakker:
Any idea on how we can tackle this? I want to release v7 after #1402, so it would be a good thing to add a warning in the readme and dart docs about this behaviour before we release the new version.

navaronbracke:
Unfortunately this is dependent on the barcode format I think. So I guess a warning / docs change is the way to go.

f1sh1918:
Hey there :) 
Is there any update on this issue? We use padding bytes and for us the qr code scanning breaks now on ios if we update to v7

navaronbracke:
Unfortunately not, since I do not have information on how the padding bytes changes for the formats, when using the VisionAPI.

f1sh1918:
ok thx for the info. Then we gonna stick to the previous version

navaronbracke:
I had a quick look at what would be needed and apparently the header bytes define the format encoding + character count.
With that we _could_ figure out how big the payload is / where it is in the byte array. Dropping the padding bytes at the end is not the hard part then.

f1sh1918:
hey @navaronbracke, 
we are highly interested in fixing this issue as soon as possible, since we are highly depending on that library and would like to avoid staying on an old v6 version long term.
We would likely to support with some sponsoring/bug bounty for that if there is a rough estimation about the effort

juliansteenbakker:
I think we can get this fixed today. I think i'm on to something on iOS. Still need to take a look at the web implementation though.

juliansteenbakker:
What i can do is convert `payloadStringValue` in swift to a byteArray, which gives the raw bytes without extra padding bytes. The downside of this is that it only works with utf8/latin.1 encoded QR codes.

If this is not the case, only `payloadData` is populated, and nothing else.

@f1sh1918 do you have any example qr's? You can also mail them to juliansteenbakker@gmail.com

f1sh1918:
thx for reaching out!
[...]

PR #1529, opened 2025-09-22
fix: migrate rawBytes to byte array from payloadStringValue.

### LINKED ISSUE #1444: Different rawBytes output when upgrading from v5 to v7
I am encoding arrays of bytes into aztec code using the go package `github.com/boombuler/barcode/aztec`. Since upgrading the `mobile_scanner` from version 5 to 7 I am seeing different outputs for the `rawBytes` value when scanning the aztec code using this package.

Image

I created this test code which encodes the bytes for `hellø` (note the final `ø`) (which are `[104, 101, 108, 108, 195, 184]`). The previous version of this package successfully returned the exact same byte array on iOS and Android. Scanning this code on the new version however returns different values on both web and iOS.

- iOS: `[226, 76, 214, 250, 44, 59, 140, 0]`
- Web: `[226, 76, 214, 252, 88, 119, 24]`

I am not sure where to go from here. It appears that the way the result is encoded has changed since the upgrade and is inconsistent in both the web and mobile versions. Does anyone have any ideas why this would be the case?

The code which captured the byte arrays was:

```dart

  void _onDetect(BarcodeCapture capture, DetectStopFunc stop) async {
    for (final barcode in capture.barcodes) {
      if (barcode.rawBytes != null) {
        debugPrint("Raw bytes: ${barcode.rawBytes}");
      }
    }
  }
```

I tried the same tests when encoding the bytes for the string `hello` (standard `o`) and both the web and iOS returned the same byte array. This could suggest that there is some encoding difference somewhere?

Any help would be greatly appreciated :)

juliansteenbakker:
This PR possibly fixed #1444, #1357 and #1317 by getting the bytes from the string value instead of `payloadData`.


juliansteenbakker:
I have updated the barcode parser on apple, and it now correctly decodes the codes mentioned in #1444, #1357 and #1317.

I have added docs to the readme explaining the difference between the frameworks and limitations on apple. However, i think this PR will fix most issues regarding rawBytes on apple.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -1129,8 +1157,9 @@ extension VNBarcodeObservation {
             ],
             "format": symbology.toInt ?? -1,
             "rawBytes": rawBytes,
+            "rawPayloadData": rawPayloadData,
```
navaronbracke, line 1172: I find this a bit unfortunate. The "raw bytes" is the entire payload data, but we also provide the "raw bytes without header & padding bytes".

Maybe we can use `"rawBytes"` and `"decodedPayloadBytes"` ? Naming is hard. Thoughts?

example/lib/widgets/scanned_barcode_label.dart
```diff
@@ -28,8 +24,36 @@ class ScannedBarcodeLabel extends StatelessWidget {
           );
         }
 
+        final String displayValues = scannedBarcodes
```
navaronbracke: Only do this in the if check below?

example/lib/widgets/scanned_barcode_label.dart
```diff
@@ -28,8 +24,36 @@ class ScannedBarcodeLabel extends StatelessWidget {
...
+
+        if (scannedBarcodes.any((e) => e.displayValue != null)) {
+          return Text(
+            'Display values: $displayValues',
+            overflow: TextOverflow.fade,
+            style: const TextStyle(color: Colors.white),
+          );
+        }
+
+        final String rawValues = scannedBarcodes
```
navaronbracke: Same here

darwin/mobile_scanner/Sources/mobile_scanner/BarcodePayloadParser.swift
```diff
@@ -0,0 +1,99 @@
...
+/// codes), the caller falls back to the ISO-Latin-1 round-trip on
+/// `payloadStringValue`, which correctly recovers ASCII bytes for those modes.
+///
+/// **Limitations, see README for full details:**
+/// - Kanji mode (0x08) is not yet supported and will cause this parser to
+///   return `nil`, falling through to the Latin-1 fallback which cannot
+///   round-trip Japanese characters.
+/// - Aztec, DataMatrix, and PDF417 are handled separately via the ISO-Latin-1
+///   round-trip in the plugin, which has its own constraints.
+enum BarcodePayloadParser {
```
navaronbracke, line 22: Do we need a unit test for this class?

lib/src/objects/barcode.dart
```diff
@@ -128,6 +130,8 @@ class Barcode {
   /// This value may be multiline if line breaks are encoded in the barcode.
   /// This value may include the supplement value.
   ///
+  /// On Apple (iOS & macOS) this value will always be UTF-8.
```
navaronbracke: Per https://api.dart.dev/dart-core/String-class.html strings in Dart are always UTF-16. I would omit this line, since the user can access the code points of the Dart string as-is?

lib/src/objects/barcode.dart
```diff
@@ -151,7 +155,15 @@ class Barcode {
   /// This is null if the raw bytes are not available.
   final Uint8List? rawBytes;
 
-  /// The raw value of `UTF-8` encoded barcodes.
+  /// The raw bytes of the barcode, including header & padding bytes.
+  /// Only available on Apple (iOS & macOS).
+  ///
+  /// If the barcode data is not UTF-8 encoded, this is the only way to access
+  /// the data.
+  final Uint8List? rawPayloadData;
```
navaronbracke: I understand that you wish to add a new field to work around this. However, my personal opinion is that this pollutes the API a little bit.

What if we do the following?

1. Deprecate `final Uint8List? rawBytes;`
2. Add a new field `final BarcodeBytes? rawDecodedBytes;`
3. Create a sealed class `BarcodeBytes`, with two subclasses, of which the special one is used by iOS/macOS.

```
sealed class BarcodeBytes {
  const BarcodeBytes();
}

final class DecodedBarcodeBytes extends BarcodeBytes {
  const DecodedBarcodeBytes(this.bytes);

  /// The decoded raw bytes of the barcode value, without header or padding bytes.
  final Uint8List bytes;
}

final class DecodedVisionBarcodeBytes extends BarcodeBytes {
  const DecodedVisionBarcodeBytes({required this.bytes, required this.rawBytes});

  /// The decoded raw bytes of the barcode value, without header or padding bytes.
  final Uint8List? decodedBytes;

  /// The raw bytes of the barcode from the Vision API.
  /// If the bytes of the original barcode are not encoded in UTF-8, this field can be used to do a manual conversion.
  final Uint8List? rawBytes;
}
```

We can then also move the docs for the limitation in this class, rather than in the README.

navaronbracke (changes requested):

lib/src/objects/barcode.dart
juliansteenbakker: Yeah, this is better, as it doesn't produce empty parameters for certain platforms

darwin/mobile_scanner/Sources/mobile_scanner/BarcodePayloadParser.swift
juliansteenbakker: We currently don't have any apple native tests, maybe we can add that in a separate PR?

navaronbracke: We can do that, yes. I have some experience with writing those.

juliansteenbakker:
I've updated the docs and added `rawDecodedBytes`. @navaronbracke what do you think?

example/lib/widgets/scanned_barcode_label.dart
```diff
@@ -28,8 +24,36 @@ class ScannedBarcodeLabel extends StatelessWidget {
           );
         }
 
+        if (scannedBarcodes.any((e) => e.displayValue != null)) {
+          final String displayValues = scannedBarcodes
```
navaronbracke, line 28: Should we filter on the non-null display values?

lib/src/objects/barcode.dart
```diff
@@ -55,6 +57,21 @@ class Barcode {
     final barcodeWidth = size?['width'] as double?;
     final barcodeHeight = size?['height'] as double?;
 
+    final rawBytesData = data['rawBytes'] as Uint8List?;
+    final rawPayloadData = data['rawPayloadData'] as Uint8List?;
+
+    final BarcodeBytes? rawDecodedBytes;
+    if (rawPayloadData != null) {
```
navaronbracke, line 64:
```suggestion
            if (rawPayloadData != null) {
        final BarcodeBytes? rawDecodedBytes = switch(data) {
          case {'rawPayloadData': final Uint8List rawPayloadData, 'rawBytes': final Uint8List? rawBytesData } => DecodedVisionBarcodeBytes(
                decodedBytes: rawBytesData,
                rawBytes: rawPayloadData,
              ),
           case {'rawBytes': final Uint8List rawBytesData } => DecodedBarcodeBytes(bytes: rawBytesData),
           _ => null,
        };
```

lib/src/objects/barcode.dart
```diff
@@ -149,14 +169,25 @@ class Barcode {
...
+  ///
+  /// This is either a [DecodedBarcodeBytes] (on Android and web)
+  /// or a [DecodedVisionBarcodeBytes] (on Apple platforms),
+  /// which may contain both decoded bytes and raw payload bytes.
+  ///
+  /// This is null if the raw bytes are not available.
+  final BarcodeBytes? rawDecodedBytes;
+
+  /// The raw value of `UTF-8` encoded barcodes on Android.
+  /// On Apple (iOS and macOS), this value can also be Latin-1 (ISO 8859-1)
```
navaronbracke: The line "On Apple (iOS and macOS), this value can also be Latin-1 (ISO 8859-1)" is incorrect? We take a Latin-1 string and convert it back to UTF-8 for Dart?

lib/src/objects/barcode_bytes.dart
```diff
@@ -0,0 +1,35 @@
+import 'dart:typed_data';
+
+/// Represents the raw bytes of a barcode.
+sealed class BarcodeBytes {
+  const BarcodeBytes();
+}
+
+/// Decoded barcode bytes, available on all platforms.
```
navaronbracke: This phrase is not true? Since this is used on Android and web, but not MacOS/iOS

lib/src/objects/barcode_bytes.dart
```diff
@@ -0,0 +1,35 @@
...
+final class DecodedBarcodeBytes extends BarcodeBytes {
+  /// Creates a new [DecodedBarcodeBytes] instance.
+  const DecodedBarcodeBytes({required this.bytes});
+
+  /// The decoded raw bytes of the barcode value.
```
navaronbracke:
```suggestion
  /// The decoded raw bytes of the barcode value.
  ///
  /// This excludes any header or padding bytes.
```

test/objects/barcode_test.dart
```diff
@@ -134,8 +137,29 @@ void main() {
...
+        expect(barcode.rawDecodedBytes, isA<DecodedBarcodeBytes>());
+        final decoded = barcode.rawDecodedBytes! as DecodedBarcodeBytes;
+        expect(decoded.bytes, rawBytes);
       });
 
+      test(
+        'creates DecodedVisionBarcodeBytes with rawBytes and rawPayloadData',
+        () {
+          final rawBytes = Uint8List.fromList([72, 101, 108, 108, 111]);
+          final rawPayloadData = Uint8List.fromList([0, 1, 72, 101, 108]);
```
navaronbracke: Just for correctness of the test, should this be a superset of `rawBytes` ?

test/objects/barcode_test.dart
```diff
@@ -134,8 +137,29 @@ void main() {
...
+          final barcode = Barcode.fromNative(<Object?, Object?>{
+            'rawBytes': rawBytes,
+            'rawPayloadData': rawPayloadData,
+          });
+
+          expect(barcode.rawBytes, rawBytes);
+          expect(barcode.rawDecodedBytes, isA<DecodedVisionBarcodeBytes>());
+          final decoded = barcode.rawDecodedBytes! as DecodedVisionBarcodeBytes;
```
navaronbracke, line 160: I know we are not testing the byte ranges here, should we do that and in the other test?

navaronbracke (changes requested):

example/lib/widgets/scanned_barcode_label.dart
juliansteenbakker: Yes, i'll add `.whereType<String>()`

lib/src/objects/barcode.dart
juliansteenbakker: Do you think this notation is better/cleaner? imho, it's not as easy to read as the if else statement.

navaronbracke: I'm fine with either

lib/src/objects/barcode.dart
juliansteenbakker: The conversion happens for displayValue/rawBytes, not for rawValue.

darwin/mobile_scanner/Sources/mobile_scanner/BarcodePayloadParser.swift
juliansteenbakker: I've created https://github.com/juliansteenbakker/mobile_scanner/issues/1658 for tracking

lib/src/objects/barcode_bytes.dart
juliansteenbakker: Yeah, i was thinking wrong, like raw bytes is available for all platforms, but i've updated the docs.

lib/src/objects/barcode.dart
navaronbracke: True, but Dart strings are still UTF-16

example/lib/widgets/scanned_barcode_label.dart
navaronbracke: `nonNulls` is a bit more concise though?

lib/src/objects/barcode_bytes.dart
juliansteenbakker: Maybe we should change this to match `DecodedVisionBarcodeBytes` `decodedBytes`? As it is the same value, so i would expect the same docs and same parameter name. Or we should change `decodedBytes` to `bytes`...

lib/src/objects/barcode.dart
juliansteenbakker: I'll update the doc, outlining that while the value is latin-1 on the native side, the platform channel will convert this.

lib/src/objects/barcode_bytes.dart
navaronbracke: Maybe do both?

example/lib/widgets/scanned_barcode_label.dart
juliansteenbakker: Ah yeah that's better

lib/src/objects/barcode_bytes.dart
juliansteenbakker: Good idea!

test/objects/barcode_test.dart
juliansteenbakker: Yeah that's more realistic

test/objects/barcode_test.dart
juliansteenbakker: I've added the actual byte comparison for completeness.

lib/src/objects/barcode_bytes.dart
```diff
@@ -0,0 +1,36 @@
...
+/// On Apple platforms, the Vision API may provide both decoded bytes
+/// (without header/padding) and the raw payload bytes
+/// (including header and padding).
+final class DecodedVisionBarcodeBytes extends BarcodeBytes {
+  /// Creates a new [DecodedVisionBarcodeBytes] instance.
+  const DecodedVisionBarcodeBytes({this.bytes, this.rawBytes});
+
+  /// The decoded raw bytes of the barcode value,
+  /// without header or padding bytes.
+  final Uint8List? bytes;
```
navaronbracke, line 29: I still wonder about the nullability here.

1. No valid bytes at all, in which case `rawBytes` and `rawDecodedBytes` is null.
2. DecodedBarcodeBytes for platforms other than iOS/MacOS
3. Valid bytes on iOS/MacOS, so a DecodedVisionBarcodeBytes, where `bytes` is null if decoding of the raw bytes failed.

For 3, can `rawBytes` ever be null actually? Don't we just have `final BarcodeBytes? rawDecodedBytes == null` in that case? (so if we can't get the actual raw bytes, we won't decode them and `BarcodeCapture.BarcodeBytes` is null anyway)

navaronbracke (changes requested):
LGTM, except one thing I'm not sure about.

lib/src/objects/barcode_bytes.dart
juliansteenbakker: We can safely make `rawBytes` non nullable.

lib/src/objects/barcode_bytes.dart
```diff
@@ -0,0 +1,39 @@
...
+///
+/// On Apple platforms, the Vision API may provide both decoded bytes
+/// (without header/padding) and the raw payload bytes
+/// (including header and padding).
+final class DecodedVisionBarcodeBytes extends BarcodeBytes {
+  /// Creates a new [DecodedVisionBarcodeBytes] instance.
+  const DecodedVisionBarcodeBytes({required this.rawBytes, this.bytes});
+
+  /// The decoded raw bytes of the barcode value,
+  /// without header or padding bytes.
```
navaronbracke, line 28:
```suggestion
  /// without header or padding bytes.
  ///
  /// This is null if removing the header or padding bytes from [rawBytes] failed.
```

juliansteenbakker: Fixed directly on develop.

PR #1538, opened 2025-09-23
fix: Assign event channel streams to null on stop()

### LINKED ISSUE #1537: [Android] BarcodeCapture stream is not working after Flutter activity is killed and reopened
I'm using the mobile_scanner plugin on a Flutter module, and it's integrated on a existing Android and iOS app.
We found a bug on Android platform that can be reproduced as follow:

- Open the Android app
- Open a FlutterActivity that has mobile_scanner.
- Scanner is woking normally.
- Close the Flutter screen using `SystemNavigator.pop()`;
- Open the FlutterActivity with mobile_scanner again.
- It is no longer notifying BarcodeCapture events

Obs: iOS is working normally.

### Analysis

The **BarcodeHandler.kt** `onCancel()` method is being invoked when the FlutterActivity is disposed, and it assigns eventSink to null.
When we start another FlutterActivity, the **BarcodeHandler.kt** `onListen()` method is not being called again.
That's because the `MethodChannelMobileScanner._eventsStream` is assigned with null coalescing, only one time.
The proposed solution is to assign `MethodChannelMobileScanner._eventsStream` back to null when the plugin is stopped.

beroso:
Fix #1537

### DIFF
--- a/lib/src/method_channel/mobile_scanner_method_channel.dart
+++ b/lib/src/method_channel/mobile_scanner_method_channel.dart
@@ -400,6 +400,8 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
     _textureId = null;
     _pausing = false;
     _surfaceProducerDelegate = null;
+    _eventsStream = null;
+    _deviceOrientationStream = null;
 
     await methodChannel.invokeMethod<void>('stop', {'force': force});
   }


juliansteenbakker:
Looks good to me!

navaronbracke:
@juliansteenbakker does this fix https://github.com/juliansteenbakker/mobile_scanner/issues/1042 ?

PR #1547, opened 2025-10-02
fix: prevent half-stopped camera state causing false ALREADY_STARTED.

yujune:
### ✨ What's the context?
Fixes a “half‑stopped” state in MobileScannerPlugin where either `device` or `captureSession` could remain non-nil after stop/pause or failed start, causing subsequent start calls to throw `ALREADY_STARTED_ERROR` and potentially leaking resources.

For example, if `captureSession` is instantiated but `device` is not, the early return from `guard let device...` will cause unrelease of `captureSession`.


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -795,29 +795,28 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
-        device.removeObserver(self, forKeyPath: #keyPath(AVCaptureDevice.videoZoomFactor))
+            device.removeObserver(self, forKeyPath: #keyPath(AVCaptureDevice.videoZoomFactor))
 #endif
-
+            self.device = nil
+        }
         latestBuffer = nil
-        self.captureSession = nil
-        self.device = nil
+        self.photoOutput = nil
```
navaronbracke: There is no `photoOutput` or `photoCompletionHandler` in scope? These lines need to be removed, since you already unset device / capture session in the if-let blocks?

navaronbracke (changes requested):

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
yujune: done removing `photoOutput` or `photoCompletionHandler`.

juliansteenbakker (approved):
Seems good to me!

PR #1565, opened 2025-10-24
fix: core style

juliansteenbakker:
We have two special lint rules in place, in order to enforce type annotation on local variables. However, this results in a lower score on pub.dev due to conflicting rules with their own scoring system. (https://pub.dev/packages/mobile_scanner/score)

I propose to remove these special rules and apply default style.

### DIFF
--- a/lib/src/method_channel/rotated_preview.dart
+++ b/lib/src/method_channel/rotated_preview.dart
@@ -27,7 +27,7 @@ final class RotatedPreview extends StatefulWidget {
     required double sensorOrientationDegrees,
     Key? key,
   }) {
-    final int facingSignForDirection = switch (cameraFacingDirection) {
+    final facingSignForDirection = switch (cameraFacingDirection) {
       CameraFacing.front => 1,
       CameraFacing.back => -1,
       CameraFacing.unknown => 1,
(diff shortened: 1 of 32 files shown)


lib/src/method_channel/rotated_preview.dart
```diff
@@ -110,7 +110,7 @@ final class _RotatedPreviewState extends State<RotatedPreview> {
 
     // Rotate the camera preview according to
     // https://developer.android.com/media/camera/camera2/camera-preview#orientation_calculation.
-    double rotationDegrees =
+    var rotationDegrees =
```
navaronbracke, line 113: This should probably be `final` and not `var` ?

navaronbracke (approved):
LGTM, with one remark.

navaronbracke:
@juliansteenbakker It seems the formatting check is unhappy?

lib/src/method_channel/rotated_preview.dart
juliansteenbakker: No, due to the line below substracting deviceOrientationDegrees from it and return this value. 
`return rotationDegrees -= deviceOrientationDegrees;`

navaronbracke: Ah, I missed that this is `-=` :P

PR #1566, opened 2025-10-24
fix: groovy syntax

juliansteenbakker:
This PR fixes deprecated groovy assignments, as well as update kotlin to a supported version.

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -37,7 +37,7 @@ android {
     }
 
     kotlinOptions {
-        jvmTarget = JavaVersion.VERSION_17
+        jvmTarget = JavaVersion.VERSION_17.toString()
     }
 
     sourceSets {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -2,7 +2,7 @@ group = 'dev.steenbakker.mobile_scanner'
 version = '1.0-SNAPSHOT'
 
 buildscript {
-    ext.kotlin_version = '1.8.0'
+    ext.kotlin_version = "2.1.0"
     repositories {
         google()
         mavenCentral()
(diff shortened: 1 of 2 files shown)


android/build.gradle
```diff
@@ -29,15 +29,15 @@ android {
...
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
     }
 
     kotlinOptions {
-        jvmTarget = '17'
+        jvmTarget = JavaVersion.VERSION_17
```
navaronbracke: This should be `JavaVersion.VERSION_17.toString()`, since it requires a string?

In upstream Flutter plugins we also use that.

example/android/app/build.gradle
```diff
@@ -24,15 +24,15 @@ if (flutterVersionName == null) {
...
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_17
-        targetCompatibility JavaVersion.VERSION_17
+        sourceCompatibility = JavaVersion.VERSION_17
+        targetCompatibility = JavaVersion.VERSION_17
     }
 
     kotlinOptions {
-        jvmTarget = '17'
+        jvmTarget = JavaVersion.VERSION_17
```
navaronbracke: This should be JavaVersion.VERSION_17.toString(), since it requires a string?

navaronbracke (changes requested):

android/build.gradle
juliansteenbakker: You are right, fixed!

navaronbracke (approved):

PR #1580, opened 2025-11-06
Claude/fix android privacy compliance

ZQ7:
Design Patterns Applied:

Lazy Initialization Pattern

Object creation deferred until first actual use
Reduces startup overhead and ensures privacy compliance
Factory Method Pattern

Encapsulates object creation logic
Allows flexible instantiation timing
Proxy Pattern

Proxy StreamHandler intercepts subscription events
Controls when real listener gets created


navaronbracke:
Converting this to a draft, until I get more info in https://github.com/juliansteenbakker/mobile_scanner/issues/1581

PR #1587, opened 2025-11-18
feat: Added support for switching between lens types

hopingsteam:
This PR adds support for selecting specific camera lens types (normal, wide, zoom) on devices with multiple cameras. 

The implementation includes a new lensType parameter in MobileScannerController, a getSupportedLenses() method to query available lenses, and platform-specific lens detection using Camera2 API on Android and AVCaptureDevice on iOS. 

I've also updated the example app includes a lens switcher button that cycles through available lenses and a menu item to display supported lens types.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -473,7 +473,7 @@ class MobileScannerHandler(
             .addCameraFilter { cameraInfos ->
                 val cameraManager = activity.getSystemService(Context.CAMERA_SERVICE) as CameraManager
 
-                cameraInfos.filter { cameraInfo ->
+                val filteredCameras = cameraInfos.filter { cameraInfo ->
                     try {
                         // Get the camera ID from CameraInfo
                         val cameraId = androidx.camera.camera2.interop.Camera2CameraInfo.from(cameraInfo).cameraId
@@ -502,6 +502,10 @@ class MobileScannerHandler(
                         true
                     }
                 }
+
+                // If filtering resulted in no cameras, return all cameras with correct facing
+                // to prevent camera binding failures
+                if (filteredCameras.isEmpty()) cameraInfos else filteredCameras
             }
             .build()
     }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
@@ -11,6 +11,7 @@ import android.util.Size
 import androidx.camera.core.CameraSelector
 import androidx.camera.core.ExperimentalGetImage
 import androidx.camera.core.ExperimentalLensFacing
+import androidx.camera.camera2.interop.Camera2CameraInfo
 import com.google.mlkit.vision.barcode.BarcodeScannerOptions
 import dev.steenbakker.mobile_scanner.objects.BarcodeFormats
 import dev.steenbakker.mobile_scanner.objects.DetectionSpeed
@@ -147,6 +148,7 @@ class MobileScannerHandler(
             "pause" -> pause(call, result)
             "stop" -> stop(call, result)
             "toggleTorch" -> toggleTorch(result)
+            "getSupportedLenses" -> getSupportedLenses(result)
             "analyzeImage" -> analyzeImage(call, result)
             "setScale" -> setScale(call, result)
             "resetScale" -> resetScale(result)
@@ -161,6 +163,7 @@ class MobileScannerHandler(
     private fun start(call: MethodCall, result: MethodChannel.Result) {
         val torch: Boolean = call.argument<Boolean>("torch") ?: false
         val facing: Int = call.argument<Int>("facing") ?: 0
+        val lensType: Int = call.argument<Int>("lensType") ?: -1
         val formats: List<Int>? = call.argument<List<Int>>("formats")
         val returnImage: Boolean = call.argument<Boolean>("returnImage") ?: false
         val speed: Int = call.argument<Int>("speed") ?: 1
@@ -177,8 +180,7 @@ class MobileScannerHandler(
 
         val barcodeScannerOptions: BarcodeScannerOptions? = buildBarcodeScannerOptions(formats, autoZoom)
 
-        val position =
-            if (facing == 0) CameraSelector.DEFAULT_FRONT_CAMERA else CameraSelector.DEFAULT_BACK_CAMERA
+        val position = selectCamera(facing, lensType)
 
         val detectionSpeed: DetectionSpeed = when (speed) {
             0 -> DetectionSpeed.NO_DUPLICATES
@@ -288,6 +290,47 @@ class MobileScannerHandler(
     private fun toggleTorch(result: MethodChannel.Result) {
         mobileScanner?.toggleTorch()
         result.success(null)
+    }
+
+    /**
+     * Get the list of supported lens types on this device.
+     *
+     * Analyzes all available cameras and categorizes them by their focal lengths.
+     */
+    private fun getSupportedLenses(result: MethodChannel.Result) {
+        val cameraManager = activity.getSystemService(Context.CAMERA_SERVICE) as CameraManager
+        val supportedLenses = mutableSetOf<Int>()
+
+        try {
+            for (cameraId in cameraManager.cameraIdList) {
+                val characteristics = cameraManager.getCameraCharacteristics(cameraId)
+
+                // Get focal lengths available for this camera
+                val focalLengths = characteristics.get(CameraCharacteristics.LENS_INFO_AVAILABLE_FOCAL_LENGTHS)
+
+                if (focalLengths != null && focalLengths.isNotEmpty()) {
+                    // Use the first focal length as representative
+                    val focalLength = focalLengths[0]
+
+                    // Categorize based on focal length
+                    when {
+                        focalLength < 4.0f -> supportedLenses.add(1)  // Wide
+                        focalLength >= 4.0f && focalLength <= 6.0f -> supportedLenses.add(0)  // Normal
+                        focalLength > 6.0f -> supportedLenses.add(2)  // Zoom
+                    }
+                }
+            }
+
+            // If no lenses were detected, return 'any' (-1)
+            if (supportedLenses.isEmpty()) {
+                result.success(listOf(-1))
+            } else {
+                result.success(supportedLenses.toList())
+            }
+        } catch (e: Exception) {
+            // On error, return 'any' (-1) as a safe default
+            result.success(listOf(-1))
+        }
     }
 
     private fun setScale(call: MethodCall, result: MethodChannel.Result) {
@@ -407,6 +450,60 @@ class MobileScannerHandler(
                 e.localizedMessage
             )
         }
+    }
+
+    /**
+     * Select the appropriate camera based on facing direction and lens type.
+     *
+     * @param facing 0 = front, 1 = back
+     * @param lensType 0 = normal, 1 = wide, 2 = zoom, -1 = any
+     * @return CameraSelector configured for the desired camera
+     */
+    private fun selectCamera(facing: Int, lensType: Int): CameraSelector {
+        val lensFacing = if (facing == 0) CameraSelector.LENS_FACING_FRONT else CameraSelector.LENS_FACING_BACK
+
+        // If no specific lens type is requested, return default camera for facing direction
+        if (lensType == -1) {
+            return if (facing == 0) CameraSelector.DEFAULT_FRONT_CAMERA else CameraSelector.DEFAULT_BACK_CAMERA
+        }
+
+        // Build a camera selector that filters by both facing and lens characteristics
+        return CameraSelector.Builder()
+            .requireLensFacing(lensFacing)
+            .addCameraFilter { cameraInfos ->
+                val cameraManager = activity.getSystemService(Context.CAMERA_SERVICE) as CameraManager
+
+                cameraInfos.filter { cameraInfo ->
+                    try {
+                        // Get the camera ID from CameraInfo
+                        val cameraId = androidx.camera.camera2.interop.Camera2CameraInfo.from(cameraInfo).cameraId
+                        val characteristics = cameraManager.getCameraCharacteristics(cameraId)
+
+                        // Get focal lengths available for this camera
+                        val focalLengths = characteristics.get(CameraCharacteristics.LENS_INFO_AVAILABLE_FOCAL_LENGTHS)
+
+                        if (focalLengths == null || focalLengths.isEmpty()) {
+                            return@filter lensType == -1
+                        }
+
+                        // Use the first focal length as representative
+                        val focalLength = focalLengths[0]
+
+                        // Categorize based on focal length (approximate 35mm equivalent ranges)
+                        // These ranges are heuristic and may need adjustment per device
+                        when (lensType) {
+                            0 -> focalLength >= 4.0f && focalLength <= 6.0f  // Normal (wide angle, ~26-35mm equiv)
+                            1 -> focalLength < 4.0f                            // Wide (ultra-wide, ~13-16mm equiv)
+                            2 -> focalLength > 6.0f                            // Zoom (telephoto, ~50mm+ equiv)
+                            else -> true
+                        }
+                    } catch (e: Exception) {
+                        // If we can't get characteristics, include this camera
+                        true
+                    }
+                }
+            }
+            .build()
     }
 
 }
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -99,6 +99,8 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
             start(call, result)
         case "toggleTorch":
             toggleTorch(result)
+        case "getSupportedLenses":
+            getSupportedLenses(result)
         case "setScale":
             setScale(call, result)
         case "setFocus":
@@ -324,7 +326,7 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                     return .landscapeRight
                 default:
                     break
-                }         
+                }
             }
         }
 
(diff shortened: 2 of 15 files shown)


navaronbracke:
Hi @hopingsteam , I have not yet done a review, but I will get back to you with feedback over the coming days. We would love to finally see lens selection implemented, so let's see if we can make this work!

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -290,6 +292,47 @@ class MobileScannerHandler(
...
+
+                // Get focal lengths available for this camera
+                val focalLengths = characteristics.get(CameraCharacteristics.LENS_INFO_AVAILABLE_FOCAL_LENGTHS)
+
+                if (focalLengths != null && focalLengths.isNotEmpty()) {
+                    // Use the first focal length as representative
+                    val focalLength = focalLengths[0]
+
+                    // Categorize based on focal length
+                    when {
```
navaronbracke, line 316: Is there a specification that classifies lenses based on focal length?
This feels a bit odd, to select lenses based on focal length only.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -290,6 +292,47 @@ class MobileScannerHandler(
...
+                        focalLength < 4.0f -> supportedLenses.add(1)  // Wide
+                        focalLength >= 4.0f && focalLength <= 6.0f -> supportedLenses.add(0)  // Normal
+                        focalLength > 6.0f -> supportedLenses.add(2)  // Zoom
+                    }
+                }
+            }
+
+            // If no lenses were detected, return 'any' (-1)
+            if (supportedLenses.isEmpty()) {
+                result.success(listOf(-1))
```
navaronbracke, line 326: Can we return an empty list instead, for the no lenses case?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -290,6 +292,47 @@ class MobileScannerHandler(
...
+            } else {
+                result.success(supportedLenses.toList())
+            }
+        } catch (e: Exception) {
+            // On error, return 'any' (-1) as a safe default
+            result.success(listOf(-1))
```
navaronbracke, line 332: We should be using `result.error()` with an error here, as there is no real safe default.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -409,4 +452,62 @@ class MobileScannerHandler(
...
+                    try {
+                        // Get the camera ID from CameraInfo
+                        val cameraId = androidx.camera.camera2.interop.Camera2CameraInfo.from(cameraInfo).cameraId
+                        val characteristics = cameraManager.getCameraCharacteristics(cameraId)
+
+                        // Get focal lengths available for this camera
+                        val focalLengths = characteristics.get(CameraCharacteristics.LENS_INFO_AVAILABLE_FOCAL_LENGTHS)
+
+                        if (focalLengths == null || focalLengths.isEmpty()) {
+                            return@filter lensType == -1
```
navaronbracke, line 486: Can we document that this case is "any" lens type?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -409,4 +452,62 @@ class MobileScannerHandler(
...
+                        }
+
+                        // Use the first focal length as representative
+                        val focalLength = focalLengths[0]
+
+                        // Categorize based on focal length (approximate 35mm equivalent ranges)
+                        // These ranges are heuristic and may need adjustment per device
+                        when (lensType) {
```
navaronbracke, line 494: Same remark here for the focal length. Maybe we should move this check into a separate method, so that it is not duplicated?

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -349,6 +351,59 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+                deviceTypes = [.builtInWideAngleCamera]
+            case 1:
+                // Wide lens - prefer ultra-wide camera
+                deviceTypes = [.builtInUltraWideCamera, .builtInWideAngleCamera]
+            case 2:
+                // Zoom lens - prefer telephoto camera
+                deviceTypes = [.builtInTelephotoCamera, .builtInWideAngleCamera]
+            default:
+                // Any lens - use default discovery order
+                deviceTypes = [.builtInTripleCamera, .builtInDualCamera, .builtInWideAngleCamera]
```
navaronbracke, line 379: This does not list _all_ lenses?

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -386,12 +442,10 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
 #endif
-        
-        // Open the camera device
+
+        // Open the camera device based on position and lens type
 #if os(iOS)
-        if #available(iOS 13.0, *) {
-            device = AVCaptureDevice.DiscoverySession(deviceTypes: [.builtInTripleCamera, .builtInDualCamera, .builtInWideAngleCamera], mediaType: .video, position: position).devices.first
-        }
+        device = selectCamera(position: position, lensType: lensType)
```
navaronbracke, line 448: `selectCamera()` should also handle MacOS, so that the code is unified.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
navaronbracke, line 455: Why does `selectCamera()` not incorporate these checks? I would prefer that `selectCamera()` is the only code path that computes the `AVCaptureDevice` to use.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -784,6 +838,72 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+     * Analyzes all available camera devices and returns the supported lens types.
+     */
+    private func getSupportedLenses(_ result: @escaping FlutterResult) {
+#if os(iOS)
+        if #available(iOS 13.0, *) {
+            var supportedLenses = Set<Int>()
+
+            // Check back cameras
+            let backDevices = AVCaptureDevice.DiscoverySession(
+                deviceTypes: [.builtInWideAngleCamera, .builtInUltraWideCamera, .builtInTelephotoCamera],
```
navaronbracke, line 853: Why does this not list all available lens types?

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -784,6 +838,72 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+                case .builtInTelephotoCamera:
+                    supportedLenses.insert(2)  // Zoom
+                default:
+                    break
+                }
+            }
+
+            // Check front cameras
+            let frontDevices = AVCaptureDevice.DiscoverySession(
+                deviceTypes: [.builtInWideAngleCamera, .builtInUltraWideCamera, .builtInTelephotoCamera],
```
navaronbracke, line 873: Same remark here

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -784,6 +838,72 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+                case .builtInTelephotoCamera:
+                    supportedLenses.insert(2)  // Zoom
+                default:
+                    break
+                }
+            }
+
+            // If no lenses were detected, return 'any' (-1)
+            if supportedLenses.isEmpty {
+                result([-1])
```
navaronbracke, line 893: This should return an empty list

example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -278,6 +279,23 @@
 /* End PBXResourcesBuildPhase section */
 
 /* Begin PBXShellScriptBuildPhase section */
+		1AADC473022DB6006ACA0CA0 /* [CP] Embed Pods Frameworks */ = {
```
navaronbracke, line 282: This change should be reverted? What Flutter version did you work on?

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -421,6 +422,24 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
...
 
+  @override
+  Future<List<CameraLensType>> getSupportedLenses() async {
+    final lensTypes = await methodChannel.invokeListMethod<Object?>(
+      'getSupportedLenses',
+    );
+
+    if (lensTypes == null || lensTypes.isEmpty) {
+      // Default to 'any' if no lenses are reported
+      return [CameraLensType.any];
```
navaronbracke, line 433: This should return an empty list?

lib/src/objects/start_options.dart
```diff
@@ -23,6 +25,15 @@ class StartOptions {
   /// The direction for the camera.
   final CameraFacing cameraDirection;
 
+  /// The lens type for the camera.
```
navaronbracke, line 28:
```suggestion
  /// The preferred lens type for the camera.
```

lib/src/mobile_scanner_controller.dart
```diff
@@ -84,6 +86,17 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
 
+  /// The lens type for the camera.
+  ///
+  /// This allows selection between normal, wide, and zoom lenses on devices
+  /// with multiple cameras.
+  ///
+  /// Defaults to [CameraLensType.any], which uses the first available camera
+  /// for the given [facing] direction.
+  ///
+  /// Currently only supported on iOS 13+ and Android.
```
navaronbracke, line 97:
```suggestion
  /// Currently only supported on iOS and Android.
```

IIRC Flutter's minimum iOS version will soon be 13.0

lib/src/mobile_scanner_controller.dart
```diff
@@ -564,6 +578,25 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
     await MobileScannerPlatform.instance.toggleTorch();
   }
 
+  /// Get the list of supported camera lens types for the current device.
```
navaronbracke, line 581: This should be updated for devices that have no cameras (the empty list case)

navaronbracke (changes requested):
Overall I have quite some feedback on the implementation.

I wonder if we should split up this PR into multiple smaller pieces?
1) Adding the lens type enum + tests for the enum in Dart
2) stub implementation of getSupportedLenses in the platform interface + adding the lens type parameter for the controller (but unused until platform implementation is done)
3) Adding the platform implementation for Android
4) Adding the platform implementation for iOS / MacOS

Also, since web is a supported platform, how are we going to eventually port _that_ implementation to use this API?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
hopingsteam: You're right - there's no official specification. The current thresholds are heuristic approximations. A more accurate approach would be calculating the 35mm equivalent focal length. 

However, this adds complexity and may not work on all devices. Would you prefer that I implement the 35mm equivalent calculation, or would documenting this as a best-effort heuristic be acceptable?

hopingsteam:
Hi @navaronbracke 

You made quite some fair points about breaking this into smaller pieces. I can split it as follows:

1. Add CameraLensType enum + Dart tests
2. Stub implementation in platform interface + add lensType parameter to controller (no-op until platforms implement)
3. Android implementation
4. iOS/macOS implementation

Should I close this PR and start fresh with the smaller ones, or would you prefer I keep this open as a reference while creating the split PRs?

Regarding web platform.. it has limited camera selection capabilities through the MediaDevices API. The getUserMedia() constraints allow:
- `facingMode` (user/environment) - already supported
- `deviceId` - to select a specific camera

However, there's no standard way to query lens characteristics (focal length, field of view) in browsers. The best we could do for web:

1. Return `[CameraLensType.any]` from `getSupportedLenses()` (indicating no lens selection support)
2. Ignore the `lensType` parameter in `start()` and fall back to default camera selection

This matches how other unsupported features are handled on web. I can add this stub implementation as part of the platform interface PR, or as a separate web PR. What's your preference?

navaronbracke:
We can keep this PR as a draft, while we work on the smaller pieces.

As for web, we can implement it with an `UnsupportedException()` for now, until we decide on what to do with it, since I doubt that camera selection on web will have high usage.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
navaronbracke: Documenting it is fine then (not sure if there is online documentation around focal lengths that explains it?)
If we can also reuse the less than / greater than checks on the focal length values, that would be nice (so that we only have one place where we need to use it)

hopingsteam:
Implemented the first chunk in PR #1604  - Added lens enum types

hopingsteam:
Implemented the second chunk in PR #1605 - Add getSupportedLenses() stub and lensType controller parameter

navaronbracke:
@hopingsteam I wonder, since the `MediaDevices` web API does not support this, maybe we should "implement" it on the web by  providing a specific error message.

For example like here: https://github.com/juliansteenbakker/mobile_scanner/blob/develop/lib/src/web/mobile_scanner_web.dart#L240-L245

We can do this in a separate PR, after the Android/iOS implementation is done.

navaronbracke:
> @hopingsteam I wonder, since the `MediaDevices` web API does not support this, maybe we should "implement" it on the web by providing a specific error message.
> 
> For example like here: https://github.com/juliansteenbakker/mobile_scanner/blob/develop/lib/src/web/mobile_scanner_web.dart#L240-L245
> 
> We can do this in a separate PR, after the Android/iOS implementation is done.

Sorry, I made a mistake in this suggestion! In the PR that adds the query for lenses, the documentation states: https://github.com/juliansteenbakker/mobile_scanner/blob/develop/lib/src/mobile_scanner_controller.dart#L599-L600

So on web, we can just return an emty list, to match the documentation? Or can we specifically return `any` on web, if the device has available video inputs?

hopingsteam:
@navaronbracke we should distinguish between "no cameras" and "cameras exist but lens type unknown."

On web, we can use `navigator.mediaDevices.enumerateDevices()` to check if video inputs exist.
-   No cameras -> return `empty set`
-   Cameras available-> return `{CameraLensType.any}`

This aligns with the existing tests that expect `CameraLensType.any` as a valid output value. Users can then distinguish between "no camera hardware" vs "camera exists but specific lens type is unknown."

I'll also update the documentation to clarify this behavior to something like:

> Returns an empty set if the device has no cameras. On platforms that do not support querying specific lens types, returns a set containing only `CameraLensType.any` if cameras are available.

hopingsteam:
Implemented the third chunk in PR #1608

navaronbracke:
@hopingsteam I know that updating the example app (for the "advanced" view with all the settings) is something you will still do, once the implementation is done, but I wonder if we should update the implementation of `switchCamera()` ?

https://github.com/juliansteenbakker/mobile_scanner/blob/develop/lib/src/mobile_scanner_controller.dart#L522

Currently that function does not take any arguments, so we can do this as a new feature, instead of a breaking change.

I would think that we can maybe change the signature to support both the CameraFacing and CameraLensType arguments.

What about introducing a new sealed class like this?

```dart
/// The options for a switch camera request.
sealed class SwitchCameraOptions {
  /// The default constructor for subclasses of [SwitchCameraOptions].
  const SwitchCameraOptions();
}

/// An implementation of [SwitchCameraOptions]
/// that specifies that the camera should toggle between [CameraFacing.front] and [CameraFacing.back].
///
/// This option does nothing if the current [CameraFacing] direction is either [CameraFacing.unknown] or [CameraFacing.external].
final class ToggleCameraDirectionOption extends SwitchCameraOptions {
  /// The default constructor for a toggle camera direction request.
  const ToggleCameraDirectionOption();
}

/// An implementation of [SwitchCameraOptions]
/// that specifies that the camera should try to select the given [facingDirection] and [preferredLensType].
final class SelectCameraDirectionAndLensTypeOption extends SwitchCameraOptions {
  /// The default constructor for a select camera request.
  ///
  /// By default the preferred lens type is set to [CameraLensType.any],
  /// to allow for selecting a specific camera on devices that might not have all available lens types.
  const SelectCameraDirectionAndLensTypeOption({required this.facingDirection, this.preferredLensType = CameraLensType.any});
  
  /// The desired facing direction to switch to.
  final CameraFacing facingDirection;
  
  /// The preferred lens type to switch to.
  final CameraLensType preferredLensType;
}
```

so that you can switch to a specific lens (on platforms that support it). Currently we only switch between front / back and ignore the lens type.

The API would then accept an optional positional parameter (since I think that using a single named parameter is a bit strange here), using the current behavior as the default argument.

```dart
Future<void> switchCamera([SwitchCameraOptions options = const ToggleCameraDirectionOption()])
```

hopingsteam:
Implemented fourth chunk in PR #1609

hopingsteam:
Implemented the last chunk in PR #1616 (together with the sealed class suggestion)

navaronbracke:
> @navaronbracke we should distinguish between "no cameras" and "cameras exist but lens type unknown."
> 
> On web, we can use `navigator.mediaDevices.enumerateDevices()` to check if video inputs exist.
> 
> * No cameras -> return `empty set`
> * Cameras available-> return `{CameraLensType.any}`
> 
> This aligns with the existing tests that expect `CameraLensType.any` as a valid output value. Users can then distinguish between "no camera hardware" vs "camera exists but specific lens type is unknown."
> 
> I'll also update the documentation to clarify this behavior to something like:
> 
> > Returns an empty set if the device has no cameras. On platforms that do not support querying specific lens types, returns a set containing only `CameraLensType.any` if cameras are available.

We also should not forget the web implementation? (as small as it is, it is still valuable)

hopingsteam:
I've added the web implementation in #1617

hopingsteam:
Closing this PR, since we've implemented it in chunks

Issue #1225, opened 2024-10-23
Deprecate MLKit for iOS


navaronbracke:
`mobile_scanner` currently uses MLKit on both iOS and Android.

However, for iOS there is a first-party library, namely the Vision API from Apple.
We already use this implementation for MacOS.

We should switch iOS over to use that implementation, which brings a bunch of improvements.
The following checklist lists the points that we should tackle when doing this migration:

- [ ] lower the iOS deployment target back to Flutter's default of 12.0, as Vision is supported on minimum iOS 11
- [ ] add back build support for arm64 iOS Simulators, which are unsupported on MLKit
- [ ] merge both the iOS and MacOS implementation sources, https://github.com/juliansteenbakker/mobile_scanner/issues/782
- [ ] add support for the scan window, using https://developer.apple.com/documentation/vision/vnimagebasedrequest/regionofinterest
- [ ] backfill the barcode capture data, which previously was automatically provided by MLKit

P0labrD:
Hello @juliansteenbakker 
My project would very much benefit from this change, I thank you for it 🙏
If you need any help to fasten the new release, I'd be happy to help 🙂

RutgerVleuten:
Just tested an integration of v7.0.0-beta.3, no issues found so-far ...  looking forward to the public release!

dincerkizildere:
Hello, is there any specific date set for the general release of version 7.0.0?

topdeveloper1219:
Oh....
where are you from @navaronbracke? Can you cooperate with us as soon as possible?

navaronbracke:
@topdeveloper1219 Anyone is free to use the v7 beta. As for the issue with analyze image, I'd have to look into it. But if you do find why it isn't completing after the await, feel free to submit a pull request

chenfisher:
@navaronbracke 
Exception in `analyzeImage`

`mobile_scanner: ^7.0.0-beta.3`

```
[ERROR:flutter/runtime/dart_vm_initializer.cc(40)] Unhandled Exception: MobileScannerBarcodeException(Could not create inference context)
#0      MethodChannelMobileScanner.analyzeImage (package:mobile_scanner/src/method_channel/mobile_scanner_method_channel.dart:191:9)
```

Thanks

navaronbracke:
@chenfisher Thanks for the stacktrace! Do you have the sample file that this occurred with?

navaronbracke:
[...]

PR #1590, opened 2025-11-22
feat: add iOS barcodeType detection

MicTheFox:
### Summary
Implements barcode type detection for iOS/macOS using heuristic pattern matching, matching Android MLKit's valueType behavior. Adds the type field to barcode results on iOS/macOS.

### Background
Android uses MLKit's valueType to detect barcode types (contact info, email, phone, URL, etc.). iOS/macOS Vision framework doesn't provide this, so this PR adds heuristic-based detection to match Android behavior.

Part of #1225

### DIFF
--- a/darwin/mobile_scanner/Sources/mobile_scanner/BarcodeTypeDetector.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/BarcodeTypeDetector.swift
@@ -1,0 +1,129 @@
+import Foundation
+
+/// Extension to detect barcode type from raw string value using heuristics.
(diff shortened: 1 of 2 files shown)


navaronbracke (approved):
LGTM! This is going to need a patch version bump and a changelog entry.

navaronbracke:
Hi, @MicTheFox , I would like to release this with the upcoming version 7.2.0, so I will add this to the changelog in a bit.

navaronbracke:
I will be amending this to the changelog for 7.2.0 tomorrow, now that the lens type feature has also landed, after I finish with work for the day.

Issue #1572, opened 2025-10-30
MobileScannerPlugin triggers SensorManager.getSensorList() causing privacy violation in China app stores


q200892907:
🧾 Issue Title

MobileScannerPlugin triggers SensorManager.getSensorList() causing privacy violation in China app stores

⸻

Description

Hi, thank you for your great work on this plugin!

We recently encountered a privacy compliance rejection from several Chinese Android app stores because the mobile_scanner plugin indirectly calls SensorManager.getSensorList() during initialization.

According to local privacy regulations and automated review systems, this is detected as accessing motion sensors (accelerometer/gyroscope), even if no actual data is collected or uploaded.
This causes the app to be rejected with the reason:

“The app reads specific types of sensor data without user consent.”

⸻

Stack trace

Here’s the relevant stack trace captured during initialization:
at android.hardware.SensorManager.getSensorList()
at android.hardware.SensorManager.getDefaultSensor(SensorManager.java:488)
at android.view.OrientationEventListener.<init>(OrientationEventListener.java:71)
at dev.steenbakker.mobile_scanner.e.<init>(SourceFile:7)
at dev.steenbakker.mobile_scanner.g0.<init>(SourceFile:101)
at dev.steenbakker.mobile_scanner.MobileScannerPlugin.onAttachedToActivity(SourceFile:61)

Deer404:
Encountered the same issue

Deer404:
@juliansteenbakker The Chinese app stores are very sensitive about sensor permissions. Can the plugin provide a configuration option to allow turning off sensor monitoring?

q200892907:
@Deer404 It can be passed using version 7.1.2, and can be temporarily downgraded.

Deer404:
> [@Deer404](https://github.com/Deer404) It can be passed using version 7.1.2, and can be temporarily downgraded.它可以通过版本 7.1.2 传递，并且可以暂时降级。

Thanks, currently only downgrading the version is possible

rean0522:
今天搞了一整天了，排查了很多插件，然后才发现是这个插件的问题，但是7.1.2在iOS中有闪退的bug，所以现在我降到了7.0.1

navaronbracke:
If I understand correctly, it is the creation of the `android.view.OrientationEventListener`, during plugin initialization that is the issue?

In other words, we should probably only set it up when the plugin actually starts the camera?

navaronbracke:
[...]

PR #1592, opened 2025-11-26
fix(android): respect app orientation lock in camera preview

### LINKED ISSUE #1571: Camera orientation does not respect preferred device orientations
`SystemChrome.setPreferredOrientations` is set to only `DeviceOrientation.portraitUp`
When I rotate my device, the camera preview is resized to fit landscape. This distorts the preview.

Version
7.1.3
not reproducible in 7.1.2

Platforms
Android

Device Model
Samsung Galaxy S22 Ultra

Doctor summary (to see all details, run flutter doctor -v):
[✓] Flutter (Channel stable, 3.29.3, on macOS 26.0.1 25A362 darwin-arm64, locale en-US)
[✓] Android toolchain - develop for Android devices (Android SDK version 36.0.0-rc1)
[✓] Xcode - develop for iOS and macOS (Xcode 26.0.1)
[✓] Chrome - develop for the web
[✓] Android Studio (version 2025.1)
[✓] VS Code (version 1.105.1)
[✓] Connected device (6 available)
[✓] Network resources

• No issues found!

wilinz:
Fixes #1571

The camera preview was rotating even when the app orientation was locked to portrait mode using SystemChrome.setPreferredOrientations. This was caused by OrientationEventListener responding to physical device rotation regardless of the app's orientation lock.

Changes:
- Modified onOrientationChanged to use getUIOrientation() instead of raw sensor data
- Added getUIOrientation() method that checks both Configuration.orientation and Display.rotation to respect the app's actual orientation
- Added getDisplay() helper method with Android R compatibility

This ensures the camera preview respects the app's orientation lock and prevents distortion when the device is physically rotated.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
@@ -2,11 +2,15 @@ package dev.steenbakker.mobile_scanner
 
 import android.app.Activity
 import android.content.Context
+import android.content.res.Configuration
 import android.hardware.SensorManager
+import android.os.Build
 import android.os.Handler
 import android.os.Looper
+import android.view.Display
 import android.view.OrientationEventListener
 import android.view.Surface
+import android.view.WindowManager
 import dev.steenbakker.mobile_scanner.utils.serialize
 import io.flutter.embedding.engine.systemchannels.PlatformChannel
 import io.flutter.plugin.common.EventChannel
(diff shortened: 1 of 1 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -50,17 +54,51 @@ class DeviceOrientationListener(
         disable()
```
navaronbracke, line 54: IIRC the camerax plugin sets the last orientation to null here too: https://github.com/flutter/packages/blob/main/packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java#L85

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
navaronbracke, line 44: Not sure if this is needed, but camerax does call stop() before starting it again: https://github.com/flutter/packages/blob/main/packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java#L55

Should we do that here too?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -50,17 +54,51 @@ class DeviceOrientationListener(
...
     }
 
+    @Suppress("deprecation")
+    private fun getDisplay(): Display {
```
navaronbracke, line 58: Display can be null.

See https://github.com/flutter/packages/blob/main/packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java#L211

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -50,17 +54,51 @@ class DeviceOrientationListener(
...
+            (activity.getSystemService(Context.WINDOW_SERVICE) as WindowManager).defaultDisplay
+        }
+    }
+
+    /**
+     * Gets the current user interface orientation based on the display rotation
+     * and configuration orientation. This respects the app's orientation lock.
+     */
+    private fun getUIOrientation(): PlatformChannel.DeviceOrientation {
+        val rotation: Int = getDisplay().rotation
```
navaronbracke, line 71: If the display is null, we should use `Surface.ROTATION_0`, like how camerax does it: https://github.com/flutter/packages/blob/main/packages/camera/camera_android_camerax/android/src/main/java/io/flutter/plugins/camerax/DeviceOrientationManager.java#L195

navaronbracke (changes requested):
Overall this is LGTM, but I have some minor nits / questions.

navaronbracke:
@wilinz Thank you for this change! I will be able to use this fix to also fix the issue of eager creation of the orientation listener.
CameraX does not suffer from that, because it internally creates the listener in start()

That is why we have https://github.com/juliansteenbakker/mobile_scanner/issues/1572 which I will try to fix separately, after this change lands.

navaronbracke:
@juliansteenbakker I will se if I can clean up this PR, so that we can use it to fix the mentioned issue. Then we should be able to land this over the weekend.

Issue #1588, opened 2025-11-19
Camera preview wrongfully rotates even when rotation is locked


frankvollebregt:
## The problem
The Widget where I use the mobile scanner is locked to portrait orientation using [`SystemChrome.setPreferredOrientation`](https://api.flutter.dev/flutter/services/SystemChrome/setPreferredOrientations.html). But when the device rotates more than 45 degrees, the camera is rotated anyways. I expect `mobile_scanner` to respect the device orientations dictated by the native app (as defined in `AndroidManifest.xml` and `Info.plist`), and moreover the orientations as configured within Flutter using the `SystemChrome` API.

### Android
On Android, it doesn't matter if the system screen rotation is set to locked or unlocked: the camera view always rotates with the device (making it misalign with the real world when rotated, if the rest of the UI is locked). This seems to occur since a recent update, my previous app version used Flutter 3.35.3 and mobile_scanner 7.0.1, and the current version (where the bug occurs) is shown below. I also updated some transitive dependencies, so something may also have changed there

### iOS
On iOS, locking the screen rotation from the system tray _does_ prevent the bug from occurring. What's more, the issue was already occuring with the last version (the aforementioned Flutter 3.35.3 and mobile_scanner 7.0.1). It may be a separate issue from Android altogether, though the behavior is exactly the same on both systems.

## Flutter version
Flutter 3.35.7 • channel stable • https://github.com/flutter/flutter.git
Framework • revision adc9010625 (4 weeks ago) • 2025-10-21 14:16:03 -0400
Engine • hash 6b24e1b529bc46df7ff397667502719a2a8b6b72 (revision 035316565a) (28 days ago) • 2025-10-21 14:28:01.000Z
Tools • Dart 3.9.2 • DevTools 2.48.0

## Steps to reproduce
(verified on Android)
1. Clone this repository
2. Start the example app (on a real device)
3. Make sure system rotation is unlocked on the device
4. Open the _Simple Mobile Scanner_
5. Rotate the device and observe the camera preview adapts correctly, changing the rotation and aspect ratio
6. Lock the system rotation
7. If you now rotate > 45 degrees, the camera preview rotates but Flutter does not, misaligning the camera

## Possible resolution
[...]

Issue #1579, opened 2025-11-06
[Android] Unwanted screen rotation still occurs during MobileScanner initialization despite fix


vedjuno236:
The camera screen still experiences a brief, unwanted screen rotation or flicker upon initialization of the `MobileScanner` widget on Android devices, even after upgrading to version 7.1.3.

**Context of Version 7.1.3:**
This version explicitly states in the changelog that it fixed this issue: "[Android] Refactored orientation detection... preventing unwanted screen rotations on MobileScanner initialization." However, the issue persists.

**Platform Configuration:**
I have already set the `screenOrientation` to `portrait` in the `AndroidManifest.xml` to lock the application orientation, but the library still seems to trigger a temporary orientation change internally.

**AndroidManifest.xml Configuration:**
```xml
<activity
    android:name=".MainActivity"
    android:exported="true"
    android:launchMode="singleTop"
    android:theme="@style/LaunchTheme"
    android:configChanges="orientation|keyboardHidden|keyboard|screenSize|smallestScreenSize|locale|layoutDirection|fontScale|screenLayout|density|uiMode"
    android:hardwareAccelerated="true"
    android:windowSoftInputMode="adjustResize"
    android:screenOrientation="portrait"> 
    <!-- ... rest of activity tag -->
</activity>

samih93:
same issue

PR #1595, opened 2025-12-02
feat: add option to disable device orientation listener and fix overlay tap blocking

carlosfiori:
### Summary
This change introduces a way to prevent the controller from subscribing to device orientation updates, useful in scenarios where these updates are unnecessary or interfere with external logic.  
It works on Android and iOS.  
It can also help with the issues below while there is no implementation that fully respects the app orientation settings:

- #1588  
- #1579  
- #1571
- #1596

### What is included
- New `disableDeviceOrientationListener` parameter in `MobileScannerController`, allowing developers to opt out of orientation listening.
- Field documented with purpose and default value.
- Initialization updated to skip subscribing to orientation changes when `disableDeviceOrientationListener` is enabled.

### Additional fix
While working on this improvement, an issue was also fixed where an `IgnorePointer` widget was blocking all taps on the overlay even when `tapToFocus` was false

### Demo
<table>
<tr>
 <th> Listener enabled
 <th> Listener disabled
<tr>
 <td>

[](https://github.com/user-attachments/assets/1e74f697-bb5f-4063-86c0-62e0b5f6d27b)
 <td>

[](https://github.com/user-attachments/assets/342f8b1a-94af-4fea-beaa-dc1c3bd928aa)
</table>

### DIFF
--- a/lib/src/mobile_scanner.dart
+++ b/lib/src/mobile_scanner.dart
@@ -296,7 +296,10 @@ class _MobileScannerState extends State<MobileScanner>
                   tapToFocusScannerWidget
                 else
                   scannerWidget,
-                IgnorePointer(child: overlay),
+                IgnorePointer(
+                  ignoring: widget.tapToFocus,
+                  child: overlay,
+                ),
               ],
             );
           },
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -34,6 +34,7 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
     this.torchEnabled = false,
     this.invertImage = false,
     this.autoZoom = false,
+    this.disableDeviceOrientationListener = false,
     this.initialZoom,
   }) : detectionTimeoutMs =
            detectionSpeed == DetectionSpeed.normal ? detectionTimeoutMs : 0,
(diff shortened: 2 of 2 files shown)


ISkymen:
Unfortunately, the problem is still present on my Galaxy S22 Ultra on this branch (

carlosfiori:
Hi @ISkymen, even after adding `disableDeviceOrientationListener = false` to the `MobileScannerController` constructor?

Could you share how your scanner is configured or some code to reproduce the problem?

ISkymen:
Hi @carlosfiori. Sorry, my fault. I missed this parameter. With "disableDeviceOrientationListener: true" in MobileScannerController everything work as needed. Thank you!

carlosfiori:
Hi @juliansteenbakker, would you take a look a this, I think it will help in many issues

juliansteenbakker:
I wil take a look at it, and other pr's, tomorrow 😄

navaronbracke:
@juliansteenbakker I would not implement the fix like this. The following PR implements th fix correctly, like how the camera plugin does it: https://github.com/juliansteenbakker/mobile_scanner/pull/1592

I would prefer getting that one landed (along with fixing the privacy issue about the eager listener allocation)
I _do_ agree that we should fix the ignore pointer bug, though. I will cherry pick that into a different PR.

carlosfiori:
@navaronbracke the referenced PR (#1592) will be migrate to iOS also? The issue with orientations also occurs on iOS..

navaronbracke:
That PR aims to fix it only for Android. (and hopefully will also address the privacy issue for specific Play Stores on Android)

We will have to look into the iOS specific bug separately. However, adding a flag to "disable" the listener is not something what I would do to work around the issue (hence why I opted to close this PR)

PR #1600, opened 2025-12-09
fix: improve pixel format selection for video output on iPhone 17

### LINKED ISSUE #1578: [iOS] Iphone 17 crashes when starting MobileScanner - Unsupported pixel format
**Issue:** 
iPhone 17 users are experiencing crashes when using MobileScanner

**Platform:**
iOS 26.0.1

**MobileScanner version**
7.1.3

**Cause:** 
Unsupported pixel format type

**Info**
There seems to be an issue with the pixel format type that is being selected in the package on iPhone 17 and above. There are reports on this on other packages too, like the [official flutter camera package ](https://github.com/flutter/flutter/issues/175828)

**Steps to reproduce:** 
With any of the new iPhones (17, 17 Pro or 17 Pro Max) 
Create a MobileScanner view with the package managed MobileScannerController

**Stacktrace** 

```
Fatal Exception: NSInvalidArgumentException
-[AVCaptureVideoDataOutput setVideoSettings:] Unsupported pixel format type - use -availableVideoCVPixelFormatTypes
 ```
```
 Fatal Exception: NSInvalidArgumentException
0  CoreFoundation                 0xc98c8 __exceptionPreprocess
1  libobjc.A.dylib                0x317c4 objc_exception_throw
2  AVFCapture                     0x508c4 -[AVCaptureVideoDataOutput supportedAssetWriterOutputFileTypes]
3  camera_avfoundation            0x6810 (Missing UUID 4661cdbfd8fc3d5a8e12d0371847d35d)
4  camera_avfoundation            0x1024c (Missing UUID 4661cdbfd8fc3d5a8e12d0371847d35d)
5  camera_avfoundation            0x100f8 (Missing UUID 4661cdbfd8fc3d5a8e12d0371847d35d)
6  camera_avfoundation            0xe964 (Missing UUID 4661cdbfd8fc3d5a8e12d0371847d35d)
7  libdispatch.dylib              0x1adc _dispatch_call_block_and_release
8  libdispatch.dylib              0x1b7ec _dispatch_client_callout
9  libdispatch.dylib              0xa468 _dispatch_lane_serial_drain
10 libdispatch.dylib              0xaf44 _dispatch_lane_invoke
11 libdispatch.dylib              0x153ec _dispatch_root_queue_drain_deferred_wlh
12 libdispatch.dylib              0x14ce4 _dispatch_workloop_worker_thread
13 libsystem_pthread.dylib        0x13b8 _pthread_wqthread
14 libsystem_pthread.dylib        0x8c0 start_wqthread```

nngb102:
# Improve Pixel Format Selection for Video Output on iPhone 17

## Problem

iPhone 17 devices (iOS 26.0.1) crash when starting MobileScanner with the error:
```
-[AVCaptureVideoDataOutput setVideoSettings:] Unsupported pixel format type - use -availableVideoCVPixelFormatTypes
```

**Root Cause**: The code was hardcoding a pixel format (`kCVPixelFormatType_32BGRA`) without checking which formats are actually supported by the device. On iPhone 17, this format is not available, causing the crash.

## Solution

Updated the pixel format selection logic in `start()` to:

1. **Check available formats**: Use `availableVideoPixelFormatTypes` from `AVCaptureVideoDataOutput` to query supported pixel formats on the device.

2. **Prioritize formats in order**:
   - `kCVPixelFormatType_32BGRA`
   - `kCVPixelFormatType_420YpCbCr8BiPlanarFullRange`
   - `kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange`

3. **Fallback**: If none of the preferred formats are available, select the first available format from the device's supported formats.

This ensures we only set a pixel format that the device supports, preventing crashes on iPhone 17 and newer devices while maintaining backward compatibility with older iOS devices.

## Code Changes

- **Location**: `darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift` (lines 461-495)
- **Change**: Replaced hardcoded pixel format assignment with dynamic format selection logic that checks device capabilities before setting `videoSettings`.

## Result

- ✅ No more crashes on iPhone 17 when starting the scanner
- ✅ Backward compatible with older iOS devices
- ✅ Automatically selects the best available format based on device capabilities

## References

- Issue: https://github.com/juliansteenbakker/mobile_scanner/issues/1578
- Similar issue reported in Flutter camera package

Fixes #1578

### DIFF
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -457,7 +457,43 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
 
         // Add video output
         let videoOutput = AVCaptureVideoDataOutput()
-        videoOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: kCVPixelFormatType_32BGRA]
+        
+        // Fix for iPhone 17: Check available pixel formats before setting
+        // Issue: https://github.com/juliansteenbakker/mobile_scanner/issues/1578
+        
+        // Define preferred pixel formats in order of preference
+        let preferredFormats: [OSType] = [
+            kCVPixelFormatType_32BGRA,
+            kCVPixelFormatType_420YpCbCr8BiPlanarFullRange,
+            kCVPixelFormatType_420YpCbCr8BiPlanarVideoRange
+        ]
+        
+        // Get available formats and convert from NSNumber to OSType
+        let availableFormats = videoOutput.availableVideoPixelFormatTypes
+        let availablePixelFormats = availableFormats.compactMap { ($0 as NSNumber).uint32Value }
+        
+        // Find the first preferred format that is available
+        var selectedFormat: OSType?
+        for format in preferredFormats {
+            if availablePixelFormats.contains(format) {
+                selectedFormat = format
+                break
+            }
+        }
+        
+        // Fallback: use first available format if no preferred format found
+        if selectedFormat == nil, let firstAvailable = availablePixelFormats.first {
+            selectedFormat = firstAvailable
+        }
+        
+        // Apply the selected format or fallback to the old default
+        if let format = selectedFormat {
+            videoOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: format]
+        } else {
+            // Ultimate fallback: use the original default format
+            videoOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: kCVPixelFormatType_32BGRA]
+        }
+        
         videoOutput.alwaysDiscardsLateVideoFrames = true
 
         videoOutput.setSampleBufferDelegate(self, queue: DispatchQueue.main)
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -455,53 +455,15 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         }
         captureSession!.sessionPreset = AVCaptureSession.Preset.high
 
-        // Add video output
         let videoOutput = AVCaptureVideoDataOutput()
-        
-        // Fix for iPhone 17: Check available pixel formats before setting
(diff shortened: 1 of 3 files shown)


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -457,7 +457,43 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
 
         // Add video output
         let videoOutput = AVCaptureVideoDataOutput()
-        videoOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: kCVPixelFormatType_32BGRA]
+        
+        // Fix for iPhone 17: Check available pixel formats before setting
+        // Issue: https://github.com/juliansteenbakker/mobile_scanner/issues/1578
```
navaronbracke: We do not specifically need a link to the issue (links can get stale). I would omit the link here.

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
```diff
@@ -457,7 +457,43 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
...
+            }
+        }
+        
+        // Fallback: use first available format if no preferred format found
+        if selectedFormat == nil, let firstAvailable = availablePixelFormats.first {
+            selectedFormat = firstAvailable
+        }
+        
+        // Apply the selected format or fallback to the old default
+        if let format = selectedFormat {
```
navaronbracke: Can we put the format selection into a separate method, so that we can do

```
let format = getPreferredVideoFormat(videoOutput: videoOutput)

videoOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: format]
```

navaronbracke (changes requested):
Thanks for this fix! This will also need a patch version bump and a changelog entry, so that we can publish the fix.

navaronbracke:
Edited the OP, so that the linked issue will be closed once this is merged.

nngb102:
@navaronbracke 
Thanks for the review! I've addressed all the feedback:

- Removed the issue link from the code comment
- Extracted the format selection logic into a separate `getPreferredVideoFormat` method
- Bumped version to 7.1.4 and added changelog entry

Ready for review!

navaronbracke (approved):

navaronbracke:
Version 7.1.4 should be available on Pub.dev now.

PR #1604, opened 2025-12-12
feat: Added lens enum types

hopingsteam:


lib/src/enums/camera_lens_type.dart
```diff
@@ -0,0 +1,41 @@
...
+  /// Any available lens type.
+  ///
+  /// When this is specified, the first available camera for the given
+  /// facing direction will be used, regardless of lens type.
+  any(-1);
+
+  const CameraLensType(this.rawValue);
+
+  factory CameraLensType.fromRawValue(int? value) {
+    switch (value) {
```
navaronbracke: This can use a switch expression.

```
return switch (value) {
  0 => normal,
  1 => wide,
  2 => zoom,
  _ => any,
};
```

navaronbracke (changes requested):

lib/src/enums/camera_lens_type.dart
hopingsteam: Solved

navaronbracke (approved):

PR #1605, opened 2025-12-13
feat: Add getSupportedLenses() stub and lensType controller parameter

hopingsteam:


test/mobile_scanner_controller_lens_type_test.dart
```diff
@@ -0,0 +1,133 @@
...
+        lensType: CameraLensType.zoom,
+      );
+
+      expect(controller.lensType, CameraLensType.zoom);
+
+      // Lens type should remain the same even after operations
+      expect(controller.lensType, CameraLensType.zoom);
+
+      await controller.dispose();
+    });
```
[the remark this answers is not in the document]
hopingsteam: Sovled

test/get_supported_lenses_test.dart
```diff
@@ -0,0 +1,310 @@
...
+import 'package:mobile_scanner/src/enums/camera_lens_type.dart';
+import 'package:mobile_scanner/src/method_channel/mobile_scanner_method_channel.dart';
+import 'package:mobile_scanner/src/mobile_scanner_platform_interface.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  group('getSupportedLenses() tests', () {
+    late MethodChannelMobileScanner platform;
+    late List<MethodCall> methodCalls;
```
navaronbracke: I see that not every test uses this variable in an expect. Can we provide this variable only scoped to the tests that actually need it? The tests that do not need it will also not have to call `add()`

test/get_supported_lenses_test.dart
```diff
@@ -0,0 +1,310 @@
...
+      methodCalls = <MethodCall>[];
+
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+          .setMockMethodCallHandler(
+            platform.methodChannel,
+            (MethodCall methodCall) async {
+              methodCalls.add(methodCall);
+
+              switch (methodCall.method) {
+                case 'getSupportedLenses':
```
navaronbracke: I think we should provide the names of the supported methods as constants in the plugin.
Filed https://github.com/juliansteenbakker/mobile_scanner/issues/1606 to track this.

We can follow up with that after work on this feature is done.

test/get_supported_lenses_test.dart
```diff
@@ -0,0 +1,310 @@
...
+    late MethodChannelMobileScanner platform;
+    late List<MethodCall> methodCalls;
+
+    setUp(() {
+      platform = MethodChannelMobileScanner();
+      MobileScannerPlatform.instance = platform;
+      methodCalls = <MethodCall>[];
+
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+          .setMockMethodCallHandler(
```
navaronbracke: Doing `setMockMethodCallHandler()` in `setUp()` feels a bit redundant, since the tests will do their own version of `setMockMethodCallHandler()` anyway, which leads to a bit of duplication.

Can we remove this invocation of setMockMethodCallHandler and work with tear downs in the tests themselves? I think you can use a separate utility method in the tests for the tear down, since `platform` is set during the set up of individual tests:

```dart
void resetMethodCallHandler() {
  TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
          .setMockMethodCallHandler(platform.methodChannel, null);
}
```

test/get_supported_lenses_test.dart
```diff
@@ -0,0 +1,310 @@
...
+
+    test('returns list of supported lenses', () async {
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+          .setMockMethodCallHandler(
+            platform.methodChannel,
+            (MethodCall methodCall) async {
+              methodCalls.add(methodCall);
+
+              if (methodCall.method == 'getSupportedLenses') {
+                return [0, 1, 2]; // normal, wide, zoom
```
navaronbracke: Can we reuse the `rawValue` of the enum?

test/get_supported_lenses_test.dart
```diff
@@ -0,0 +1,310 @@
...
+            },
+          );
+
+      final lenses = await platform.getSupportedLenses();
+
+      expect(lenses, hasLength(1));
+      expect(lenses.first, CameraLensType.any);
+    });
+
+    test('handles unknown integer values (maps to any)', () async {
```
navaronbracke: What with for example `[0, -99, 1]` ?

test/start_options_test.dart
```diff
@@ -0,0 +1,241 @@
...
+        returnImage: false,
+        torchEnabled: false,
+        invertImage: false,
+        autoZoom: false,
+        initialZoom: null,
+      );
+
+      final map = options.toMap();
+
+      expect(map['facing'], CameraFacing.back.rawValue);
```
navaronbracke: Can we build a `Map<String, Object?>` with all the expected values and do one `expect()` call for the entire thing?

test/start_options_test.dart
```diff
@@ -0,0 +1,241 @@
...
+        initialZoom: null,
+      );
+
+      final map = options.toMap();
+
+      expect(map.containsKey('formats'), false);
+    });
+
+    test('toMap correctly maps all lens types', () {
+      const lensTypes = [
```
navaronbracke: Can we use a for loop on `CameraLensType.values` , so that we do not miss any?

test/start_options_test.dart
```diff
@@ -0,0 +1,241 @@
...
+        initialZoom: 2,
+      );
+
+      final map = options.toMap();
+
+      expect(map['initialZoom'], 2);
+    });
+
+    test('toMap handles all camera facing directions', () {
+      const facingDirections = [
```
navaronbracke: Same remark about using `.values`

test/start_options_test.dart
```diff
@@ -0,0 +1,241 @@
...
+          reason:
+              'Camera facing $facing should map to raw value '
+              '${facing.rawValue}',
+        );
+      }
+    });
+
+    test('toMap handles boolean flags correctly', () {
+      const testCases = [
+        (returnImage: true, torch: true, invertImage: true, autoZoom: true),
```
navaronbracke, line 203: We do not test all variants of each flag combination? (not sure if it is important to do so?)

test/mobile_scanner_controller_lens_type_test.dart
```diff
@@ -0,0 +1,119 @@
...
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  group('MobileScannerController lens type tests', () {
+    test('controller initializes with default lens type (any)', () async {
+      final controller = MobileScannerController(autoStart: false);
+
+      expect(controller.lensType, CameraLensType.any);
+
+      await controller.dispose();
```
navaronbracke: I do not think we need to call dispose (since we never call `start()` ?) ? This test only calls the constructor of the controller, to verify instance properties, and does not "start" a (fake) camera.

The same is true for some of the other tests in this file.

test/mobile_scanner_controller_lens_type_test.dart
```diff
@@ -0,0 +1,119 @@
...
+        autoStart: false,
+        lensType: CameraLensType.zoom,
+      );
+
+      expect(controller.lensType, CameraLensType.zoom);
+
+      await controller.dispose();
+    });
+
+    test('different controllers can have different lens types', () async {
```
navaronbracke: While the test itself somewhat makes sense, the plugin currently does not support running multiple controllers side by side. This is because it is unclear which controller should get ownership of the single logical camera driver (regardless of lenses)

I would omit this test.

test/mobile_scanner_controller_lens_type_test.dart
```diff
@@ -0,0 +1,119 @@
...
+      expect(normalController.lensType, CameraLensType.normal);
+      expect(wideController.lensType, CameraLensType.wide);
+      expect(zoomController.lensType, CameraLensType.zoom);
+
+      await normalController.dispose();
+      await wideController.dispose();
+      await zoomController.dispose();
+    });
+
+    test('controller with all lens types can be created', () async {
```
navaronbracke: Isn't this duplicating the individual tests for the single lens type values?

I do prefer keeping this single test, since it covers all lens types.

test/mobile_scanner_controller_lens_type_test.dart
```diff
@@ -0,0 +1,119 @@
...
+          autoStart: false,
+          lensType: lensType,
+        );
+
+        expect(controller.lensType, lensType);
+        await controller.dispose();
+      }
+    });
+
+    test('lens type property is read-only and immutable', () async {
```
navaronbracke: This test makes no sense? You don't actually do `controller.lensType = someOtherType;` in the test.
And because it is a `final` property, you won't be able to anyway.

I would omit this test.

navaronbracke (changes requested):
I went ahead and gave it a review. Overall it is nice to see some extra test coverage!
I do have some feedback, though.

test/get_supported_lenses_test.dart
```diff
@@ -0,0 +1,330 @@
...
+            },
+          );
+
+      final lenses = await platform.getSupportedLenses();
+
+      expect(lenses, hasLength(1));
+      expect(lenses.first, CameraLensType.any);
+    });
+
+    test('filters unknown values while keeping valid ones', () async {
```
navaronbracke: Is this a typo? If I understand correctly, invalid values map to `CameraLensType.any`, so it does not "filter" anything. It maps invalid values to the "any" value. So maybe the test name should be:

`"maps unknown values to 'any' lens type"`

Also, what happens if we have `0, -99, 1, -10`, do we get multiple "any" lens types?
I wonder if the platform interface should return a `Set` and not a `List` then?

navaronbracke (changes requested):
Almost there! One small remark still.

test/get_supported_lenses_test.dart
hopingsteam: The test name was misleading - it's mapping unknown values to any, not filtering them out. I'll rename it.

navaronbracke: Still, would we then get "any" multiple times? It would only make sense if it was included once (if present at all)?

hopingsteam:
I've changed the return type from `List` to `Set` so duplicates are automatically handled. Now if the platform returns something like `[normal, -99, wide, -10, 99]`, we get `{normal, wide, any}` instead of having `any` appear 3 times.

Also merged the `@visibleForTesting` constants PR and updated the tests to use them.

test/get_supported_lenses_test.dart
hopingsteam: I've pushed your suggestion!

test/start_options_test.dart
hopingsteam: The flags are independent - they don't interact with each other in `toMap()`. It just passes each value through, so testing all **16 permutations** would essentially be testing the same simple assignment repeatedly. The three cases cover: all true, all false, and a mixed scenario, which should be enough to catch any mapping issues.

lib/src/mobile_scanner_controller.dart
```diff
@@ -576,6 +590,42 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
     await MobileScannerPlatform.instance.updateScanWindow(window);
   }
 
+  /// Get the set of supported camera lens types for the current device.
+  ///
+  /// Returns a set of [CameraLensType] values that are available on the
+  /// device. This can be used to determine which lens types can be used
+  /// with the scanner. Using a set ensures no duplicate lens types are
```
navaronbracke: The line `Using a set ensures no duplicate lens types are returned.` is redundant, since that is what a Set datatype does anyway. (it does not allow multiple equal objects)

lib/src/mobile_scanner_platform_interface.dart
```diff
@@ -117,6 +118,19 @@ abstract class MobileScannerPlatform extends PlatformInterface {
     throw UnimplementedError('toggleTorch() has not been implemented.');
   }
 
+  /// Get the set of supported camera lens types for the current device.
+  ///
+  /// Returns a set of [CameraLensType] values that are available on the
+  /// device. Using a set ensures no duplicate lens types are returned.
```
navaronbracke: Same here for the line `Using a set ensures no duplicate lens types are returned.`

navaronbracke (changes requested):
One small remark about doc comments, but otherwise this is ready to merge.

hopingsteam:
I added those comments because I was changing from `List` to `Set` and wanted to explain the reasoning behind the change. But as you've said - explaining what a Set does is redundant like explaining that water is wet

navaronbracke (approved):

PR #1608, opened 2025-12-14
feat: Android implementation for querying camera lenses

hopingsteam:
There is no `getLensType()` function in Android. We deduced it using the 35mm Equivalent standard.

$$\text{Equivalent35mm} = \text{ActualFocalLength} \times \frac{43.27}{\sqrt{w^2 + h^2}}$$

* **43.27**: The diagonal size (mm) of a full-frame sensor [(standard reference)](https://en.wikipedia.org/wiki/35_mm_equivalent_focal_length).
* **$$\sqrt{w^2 + h^2}$$**: The diagonal size (mm) of the phone's physical sensor.

**Typical Results are:**

* **~13mm - 16mm**: The "0.5x" or Ultra-Wide lens.
* **~24mm - 26mm**: The "1x" or Standard Main lens.
* **~50mm - 75mm**: The "2x" or "3x" Telephoto lens.
* **~120mm+**: The "5x" or "10x" Periscope lens.


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -33,6 +38,114 @@ class MobileScannerHandler(
     private val addPermissionListener: (RequestPermissionsResultListener) -> Unit,
     textureRegistry: TextureRegistry): MethodChannel.MethodCallHandler {
 
+    companion object {
```
navaronbracke: Just a suggestion, but can we put this new code, that computes the lens type / selects a lens for the camera into a utility class? For example `MobileScannerCameraLensSelector` ? (prefixing it with MobileScanner, because I don't know if CameraX has a class with a similar name)

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -33,6 +38,114 @@ class MobileScannerHandler(
...
+         *
+         * The 35mm equivalent is calculated using the formula:
+         * equivalent = focalLength * (43.27 / sensorDiagonal)
+         *
+         * This approach is more accurate than using raw focal lengths because it
+         * accounts for sensor size variations between devices.
+         *
+         * References:
+         * - 35mm equivalent focal length: https://en.wikipedia.org/wiki/35_mm_equivalent_focal_length
+         * - Android CameraCharacteristics documentation
```
navaronbracke: Can we use the link instead? `https://developer.android.com/reference/android/hardware/camera2/CameraCharacteristics`

android/src/test/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandlerTest.kt
```diff
@@ -0,0 +1,272 @@
...
+    // ==========================================================================
+    // calculate35mmEquivalent tests
+    // ==========================================================================
+
+    @Test
+    fun calculate35mmEquivalent_typicalSmartphoneSensor_returnsExpectedValue() {
+        // Typical smartphone sensor: ~5.6mm x 4.2mm diagonal ≈ 7mm
+        // With 4.3mm focal length: 4.3 * (43.27 / 7) ≈ 26mm (standard main lens)
+        val result = MobileScannerHandler.calculate35mmEquivalent(4.3f, 5.6f, 4.2f)
+        assertEquals(26, result)
```
navaronbracke: Should we use the `absoluteTolerance` parameter in `assertEquals()` to avoid rounding errors? (multiple times in the tests)

navaronbracke (changes requested):

android/src/test/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandlerTest.kt
hopingsteam: in this case we're fine without it. The `calculate35mmEquivalent` function returns `Int` - it truncates the result with `.toInt()` at the end. So the output is always a deterministic whole number, no floating-point values to compare.

The truncation behavior is deterministic:
-  `26.999` -> `26`
-  `26.001` -> `26`

So assertEquals with exact integer matching is appropriate here.

The tolerance parameter would make sense if we were returning `Float` or `Double`, but since we're testing exact integer outputs, standard assertEquals is the right choice here.

navaronbracke (approved):

android/src/test/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandlerTest.kt
navaronbracke: Right, my mistake, sorry!

PR #1616, opened 2025-12-18
feat: Implement example for lens switching

hopingsteam:


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -2,6 +2,7 @@ package dev.steenbakker.mobile_scanner
 
 import android.app.Activity
 import android.content.Context
+import android.hardware.camera2.CameraCharacteristics
```
navaronbracke, line 5: Did we forget this import?

example/lib/widgets/buttons/switch_lens_button.dart
```diff
@@ -0,0 +1,125 @@
...
+  Widget build(BuildContext context) {
+    // Only show button if there are multiple lens options available
+    if (_availableLenses.length < 2) {
+      return const SizedBox.shrink();
+    }
+
+    return Column(
+      mainAxisSize: MainAxisSize.min,
+      children: [
+        IconButton(
```
navaronbracke: I think you should be able to use https://api.flutter.dev/flutter/material/TextButton/TextButton.icon.html here

lib/src/mobile_scanner_controller.dart
```diff
@@ -552,6 +586,84 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
     }
   }
 
+  Future<void> _toggleLensType() async {
+    final supportedLenses = await getSupportedLenses();
```
navaronbracke, line 592: I assume this is not cached for the case where a user attached an external camera?

lib/src/mobile_scanner_controller.dart
```diff
@@ -596,8 +708,9 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
   /// device. This can be used to determine which lens types can be used
   /// with the scanner.
   ///
-  /// Returns an empty set if the device has no cameras, or if the platform
-  /// does not support querying available lens types.
+  /// Returns an empty set if the device has no cameras. On platforms
+  /// that do not support querying specific lens types, returns a set
+  /// containing only `CameraLensType.any` if cameras are available.
```
navaronbracke: This can become a clickable doc link.

```suggestion
  /// containing only [CameraLensType.any] if cameras are available.
```

lib/src/mobile_scanner_platform_interface.dart
```diff
@@ -123,8 +123,9 @@ abstract class MobileScannerPlatform extends PlatformInterface {
   /// Returns a set of [CameraLensType] values that are available on the
   /// device.
   ///
-  /// Returns an empty set if the device has no cameras, or if the platform
-  /// does not support querying available lens types.
+  /// Returns an empty set if the device has no cameras. On platforms
+  /// that do not support querying specific lens types, returns a set
+  /// containing only `CameraLensType.any` if cameras are available.
```
navaronbracke: Here too

```suggestion
  /// containing only [CameraLensType.any] if cameras are available.
```

test/switch_camera_option_test.dart
```diff
@@ -0,0 +1,152 @@
...
+      test('can be created with all CameraLensType values', () {
+        for (final lens in CameraLensType.values) {
+          final option = SelectCamera(lensType: lens);
+
+          expect(option.lensType, lens);
+        }
+      });
+    });
+
+    group('pattern matching', () {
```
navaronbracke, line 91: At first this feels a bit weird to review, but I assume this is to validate the API, so that users of the API can use SwitchCameraOption with pattern matching?

navaronbracke (changes requested):
Some minor remarks, but overall this is great!

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
hopingsteam: Yes, when running the example on my Pixel I've discovered that we've missed it.

lib/src/mobile_scanner_controller.dart
hopingsteam: The supported lenses are fetched fresh each time to handle dynamic camera changes, such as external cameras being attached or detached at runtime. I've added a comment to clarify this.

test/switch_camera_option_test.dart
hopingsteam: Yes, exactly. These tests validate that the sealed class hierarchy works correctly with Dart's pattern matching - ensuring users can match each subtype, extract properties via destructuring, and that the matching is exhaustive.

It serves as both validation and documentation of the intended API usage.

lib/src/mobile_scanner_platform_interface.dart
hopingsteam: Solved

lib/src/mobile_scanner_controller.dart
hopingsteam: Solved

example/lib/widgets/buttons/switch_lens_button.dart
hopingsteam: Solved

navaronbracke (approved):

PR #1617, opened 2025-12-18
feat: Implemented lens selection on web

hopingsteam:


navaronbracke (approved):

navaronbracke:
@hopingsteam Now that we have all the implementation parts done, do you want to do a last PR, for the changelog / version? (so version 7.2.0, since this is a new feature)

I would include that we added support for switching lens types on Android/iOS and constants for testing the method channel methods were also added.

hopingsteam:
@navaronbracke I've opened another PR for that.

PR #1623, opened 2025-12-19
fix: rotation while locked for apple

### LINKED ISSUE #1588: Camera preview wrongfully rotates even when rotation is locked

juliansteenbakker:
Fixes #1588, camera rotating when rotation is enabled, but restricted via:

```dart
    SystemChrome.setPreferredOrientations(<DeviceOrientation>[
      DeviceOrientation.portraitUp,
    ]);
```


darwin/mobile_scanner/Sources/mobile_scanner/DeviceOrientationStreamHandler.swift
```diff
@@ -44,25 +39,55 @@ class DeviceOrientationStreamHandler: NSObject, FlutterStreamHandler {
...
+            sink(orientationString)
+        } else {
+            // Fallback for iOS < 13: use device orientation
+            let orientation = UIDevice.current.orientation
+
+            if orientation == .faceUp || orientation == .faceDown {
+                return
+            }
+
+            let orientationString = orientation.toOrientationString
```
navaronbracke, line 97: Should we set lastSentOrientation here?

navaronbracke, line 97: There is no check to not send an event if the orientation changed, for this code path?

navaronbracke (approved):
LGTM, with minor nits.

navaronbracke:
@juliansteenbakker I probably forgot to ask, but is `setPreferredOrientations()` working correctly on Android with the latest patch? (I didn't verify that yet)

juliansteenbakker:
Not yet i think, this patch is only for apple.

frankvollebregt (approved):
Thank you for looking into this issue :) I've just tested the version from this PR, and it looks like on iOS, this works as expected now for my use case (restricted device orientation using `SystemChrome` correctly prevents the preview from rotating, landscape allowed elsewhere in the app)

Tested on:
- iPhone XR (iOS 18.7.3)
- Flutter 3.38.8 / Dart 3.10.7

navaronbracke:
@juliansteenbakker I have 2 minor questions on this PR (see above) but otherwise this is good to merge.

juliansteenbakker:
Sorry for the delay. Thanks for the sharp eye @navaronbracke. I've added the check for that codepath aswell.

navaronbracke (approved):
New changes LGTM!

PR #1625, opened 2025-12-26
test: Implemented tests for data classes

hopingsteam:


juliansteenbakker:
For some reason the base commit is lost so no codecov message, but coverage increases from 41 to 54%. Thank you for the tests!

Issue #1615, opened 2025-12-16
Android: Camera fails to pause correctly after restarting (isPaused flag not reset)


Politanskyi:
Describe the bug
On Android, if the scanner is paused and then restarted, subsequent calls to pause() fail to actually stop the camera hardware. This happens because the isPaused state flag is not reset to false when the camera starts.

Steps to Reproduce
Start the MobileScanner.

Call controller.pause() -> Camera pauses correctly (isPaused becomes true).

Call controller.start() -> Camera restarts and works.

Call controller.pause() again.

Result: The native pause() method returns immediately (treating it as "AlreadyPaused") without executing pauseCamera(). The camera continues running in the background or fails to update its state correctly.

Root Cause Analysis
I investigated the native Android code in android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt.

When pause() is called, isPaused is set to true.

When start() is called subsequently, the isPaused variable is not reset to false.

Consequently, the next pause() check fails:

`// MobileScanner.kt
fun pause(force: Boolean = false) {
    if (!force) {
        if (isPaused) { // This remains true from the previous cycle
            throw AlreadyPaused()
        }
        // ...
    }
    // ...
    pauseCamera() // This is never reached
}
`

Suggested Fix
Add isPaused = false at the beginning of the start method in MobileScanner.kt.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt

`fun` start(
    barcodeScannerOptions: BarcodeScannerOptions?,
    returnImage: Boolean,
    cameraPosition: CameraSelector,
    torch: Boolean,
    detectionSpeed: DetectionSpeed,
    torchStateCallback: TorchStateCallback,
    zoomScaleStateCallback: ZoomScaleStateCallback,
    mobileScannerStartedCallback: MobileScannerStartedCallback,
    mobileScannerErrorCallback: (exception: Exception) -> Unit,
    detectionTimeout: Long,
    cameraResolutionWanted: Size?,
    invertImage: Boolean,
    initialZoom: Double?,
) {
    this.detectionSpeed = detectionSpeed
    this.detectionTimeout = detectionTimeout
    this.returnImage = returnImage
    this.invertImage = invertImage
        
    // [FIX START] Reset isPaused flag here
    isPaused = false
    // [FIX END]

    if (camera?.cameraInfo != null && preview != null && surfaceProducer != null && !isPaused) {
[...]

PR #1645, opened 2026-02-09
fix: [Android] reset isPaused flag on start() called

vladimir400:
Fix for Android's reset of the flag isPaused, when start is called. After the first pause() it is set to true, but after never reset, so it always stays isPaused.
There is an open issue about this : https://github.com/juliansteenbakker/mobile_scanner/issues/1615


juliansteenbakker:
LGTM!

PR #1647, opened 2026-02-09
fix: latest linter style by very_good_analysis

juliansteenbakker:
This PR fixes current analyzer issues by migrating to the latest very_good_analysis package, as well as pinning it to this version, because updates to linter rules causes the build to fail, so next updates should be verified manually. Dependabot is readded to keep track of updates.

This PR also matches the linter version of the plugin and the example app.

### DIFF
--- a/example/lib/main.dart
+++ b/example/lib/main.dart
@@ -1,3 +1,5 @@
+import 'dart:async';
+
 import 'package:flutter/material.dart';
 import 'package:mobile_scanner/mobile_scanner.dart';
 import 'package:mobile_scanner_example/screens/mobile_scanner_advanced.dart';
@@ -84,10 +86,13 @@ class _ExampleHome extends StatelessWidget {
     IconData icon,
   ) {
     return GestureDetector(
-      onTap:
-          () => Navigator.of(
+      onTap: () {
+        unawaited(
+          Navigator.of(
             context,
           ).push(MaterialPageRoute<void>(builder: (context) => page)),
+        );
+      },
       child: Card(
         elevation: 5,
         shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(15)),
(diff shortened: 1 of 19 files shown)


example/lib/main.dart
```diff
@@ -84,11 +84,10 @@ class _ExampleHome extends StatelessWidget {
     IconData icon,
   ) {
     return GestureDetector(
-      onTap: () {
-        Navigator.of(
-          context,
-        ).push(MaterialPageRoute<void>(builder: (context) => page));
-      },
+      onTap:
```
navaronbracke: I did find the proginal formatting better here?

navaronbracke (approved):
LGTM, with nit.

example/lib/main.dart
juliansteenbakker: The pro is that this doesnt require async/unawaited, the con is indeed different formatting. Do you think wrapping in unawaited is better?

navaronbracke: Why would it not require unawaited? What would be different from

```
() {
  final route = MaterialPageRoute<void>(builder: (context) => page);

  return Navigator.of(context).push(route);
}
```

juliansteenbakker: push is `Future<T?>`, so we can't return it because the function expects a void. The block variant requires an `await` or explicit `unawaited()`. However, this is not needed for the arrow expression.

So its either 
```dart
      onTap: () {
        unawaited(Navigator.of(
          context,
        ).push(MaterialPageRoute<void>(builder: (context) => page)));
      },
```

or

```dart
          () => Navigator.of(
            context,
          ).push(MaterialPageRoute<void>(builder: (context) => page)),
```

navaronbracke: I'd wrap it with unawaited to make it explicit, (since sometimes you want the result of the route) just my opinion.

juliansteenbakker: Fair point, i've changed it 😄

PR #1652, opened 2026-02-10
fix: rotation while locked for android

### LINKED ISSUE #1588: Camera preview wrongfully rotates even when rotation is locked

juliansteenbakker:
Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/1588 for android, camera rotating when rotation is enabled, but restricted via:

    SystemChrome.setPreferredOrientations(<DeviceOrientation>[
      DeviceOrientation.portraitUp,
    ]);
    
    
 I did discover a bug with the overlay, which does not turn when the display rotates and rotation is enabled, but this is also presented in the develop branch, so should be handled by a different PR.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
@@ -2,14 +2,11 @@ package dev.steenbakker.mobile_scanner
 
 import android.app.Activity
 import android.content.Context
-import android.database.ContentObserver
-import android.hardware.SensorManager
+import android.hardware.display.DisplayManager
 import android.os.Build
 import android.os.Handler
 import android.os.Looper
-import android.provider.Settings
 import android.view.Display
-import android.view.OrientationEventListener
 import android.view.Surface
 import android.view.WindowManager
 import dev.steenbakker.mobile_scanner.utils.serialize
@@ -17,13 +14,15 @@ import io.flutter.embedding.engine.systemchannels.PlatformChannel
 import io.flutter.plugin.common.EventChannel
 
 /**
- * This class will listen to device orientation changes.
+ * This class listens to display orientation changes.
  *
- * When a new orientation is received, the registered listener will be invoked.
+ * The orientation is determined from the display rotation rather than raw sensor data.
+ * This ensures that the reported orientation respects both the system rotation lock
+ * andf Flutter [SystemChrome.setPreferredOrientations].
  */
 class DeviceOrientationListener(
     private val activity: Activity,
-) : OrientationEventListener(activity, SensorManager.SENSOR_DELAY_NORMAL), EventChannel.StreamHandler {
+) : EventChannel.StreamHandler {
 
     // The event sink that handles device orientation events.
     private var deviceOrientationEventSink: EventChannel.EventSink? = null
@@ -31,13 +30,12 @@ class DeviceOrientationListener(
     // The last received orientation. This is used to prevent duplicate events.
     private var lastOrientation: PlatformChannel.DeviceOrientation? = null
 
-    // Cached auto-rotate setting to avoid querying ContentResolver on every orientation change.
-    private var autoRotateEnabled: Boolean = false
-
-    // Observer for auto-rotate setting changes.
-    private val autoRotateObserver = object : ContentObserver(Handler(Looper.getMainLooper())) {
-        override fun onChange(selfChange: Boolean) {
-            autoRotateEnabled = queryAutoRotateSetting()
+    // Listener for display configuration changes.
+    private val displayListener = object : DisplayManager.DisplayListener {
+        override fun onDisplayAdded(displayId: Int) {}
+        override fun onDisplayRemoved(displayId: Int) {}
+        override fun onDisplayChanged(displayId: Int) {
+            sendOrientationIfChanged()
         }
     }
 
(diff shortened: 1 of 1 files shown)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -2,42 +2,40 @@ package dev.steenbakker.mobile_scanner
...
 import io.flutter.plugin.common.EventChannel
 
 /**
- * This class will listen to device orientation changes.
+ * This class listens to display orientation changes.
  *
- * When a new orientation is received, the registered listener will be invoked.
+ * The orientation is determined from the display rotation rather than raw sensor data.
+ * This ensures that the reported orientation respects both the system rotation lock
+ * andf Flutter [SystemChrome.setPreferredOrientations].
```
navaronbracke: "and"

navaronbracke: `SystemChrome.setPreferredOrientations` is a Dart symbol, not a Kotlin Symbol, so this should be backticks?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -2,42 +2,40 @@ package dev.steenbakker.mobile_scanner
...
  */
 class DeviceOrientationListener(
     private val activity: Activity,
-) : OrientationEventListener(activity, SensorManager.SENSOR_DELAY_NORMAL), EventChannel.StreamHandler {
```
navaronbracke, line 26: Would this fix https://github.com/juliansteenbakker/mobile_scanner/issues/1572 ?

navaronbracke (approved):
LGTM, with a minor question

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
juliansteenbakker: I guess so, no more need for `SensorManager`, that would be awesome! I've asked if someone can verify this.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
juliansteenbakker: You're right, lets change that

juliansteenbakker: Sharp as always 😄

navaronbracke:
@juliansteenbakker What if we merge this, follow up with that overlay issue in a new PR and release this as a new version? Then more people will probably verify the sensor list bug. If we don't get any new reports by then, then we can probably close that issue.

juliansteenbakker:
Sounds like a good plan, but ill continue on the overlay issue tomorrow.

PR #1654, opened 2026-02-17
fix: replaced dead analysisBuilder with correct target rotation

juliansteenbakker:
This PR fixes the orientation of the overlay on rotation of the device by updating the targetRotation of the analysis object.

It also removes the dead ResolutionSelector code, as the analysisBuilder is never used after building.

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
@@ -30,11 +30,15 @@ class DeviceOrientationListener(
     // The last received orientation. This is used to prevent duplicate events.
     private var lastOrientation: PlatformChannel.DeviceOrientation? = null
 
+    /// Called with the new [Surface.ROTATION_*] value whenever the display rotation changes.
+    var onDisplayRotationChanged: ((Int) -> Unit)? = null
+
     // Listener for display configuration changes.
     private val displayListener = object : DisplayManager.DisplayListener {
         override fun onDisplayAdded(displayId: Int) {}
         override fun onDisplayRemoved(displayId: Int) {}
         override fun onDisplayChanged(displayId: Int) {
+            onDisplayRotationChanged?.invoke(getDisplay().rotation)
             sendOrientationIfChanged()
         }
     }
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
@@ -65,6 +65,7 @@ class DeviceOrientationListener(
     fun stop() {
         val displayManager = activity.getSystemService(Context.DISPLAY_SERVICE) as DisplayManager
         displayManager.unregisterDisplayListener(displayListener)
+        onDisplayRotationChanged = null
     }
 
     @Suppress("deprecation")
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
@@ -30,7 +30,7 @@ class DeviceOrientationListener(
     // The last received orientation. This is used to prevent duplicate events.
     private var lastOrientation: PlatformChannel.DeviceOrientation? = null
 
-    /// Called with the new [Surface.ROTATION_*] value whenever the display rotation changes.
+    // Called with the new [Surface.ROTATION_*] value whenever the display rotation changes.
     var onDisplayRotationChanged: ((Int) -> Unit)? = null
 
     // Listener for display configuration changes.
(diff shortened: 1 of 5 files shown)


juliansteenbakker:
@navaronbracke there is one other bug i found while fixing this issue, that the rotated preview on landscape does not respect the scanWindow. I will try to fix this right now because i want to release asap.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -460,6 +455,7 @@ class MobileScanner(
             }
 
             val analysis = analysisBuilder.build().apply { setAnalyzer(analysisExecutor, captureOutput) }
+            imageAnalysis = analysis
```
navaronbracke, line 443: Should we set this back to null at some point?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -444,13 +445,7 @@ class MobileScanner(
                     override fun onDisplayRemoved(displayId: Int) {}
 
                     override fun onDisplayChanged(displayId: Int) {
-                        val selector = ResolutionSelector.Builder().setResolutionStrategy(
```
navaronbracke: Also remove the unused imports for these?

navaronbracke (changes requested):

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
juliansteenbakker: ResolutionSelector is still used at line 431 for the "normal" start procedure.

navaronbracke: Right, my bad!

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
juliansteenbakker: Yes, i've added this to the dispose function

juliansteenbakker:
I've deduplicated the orientation listener, fixed texture size, cleaned up the code and added changelog entry. Do you think it's ready to merge @navaronbracke ?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -30,11 +30,15 @@ class DeviceOrientationListener(
     // The last received orientation. This is used to prevent duplicate events.
     private var lastOrientation: PlatformChannel.DeviceOrientation? = null
 
+    /// Called with the new [Surface.ROTATION_*] value whenever the display rotation changes.
```
navaronbracke:
```suggestion
    // Called with the new [Surface.ROTATION_*] value whenever the display rotation changes.
```

lib/src/mobile_scanner.dart
```diff
@@ -251,7 +252,13 @@ class _MobileScannerState extends State<MobileScanner>
         ScanWindowUtils.calculateScanWindowRelativeToTextureInPercentage(
           widget.fit,
           widgetScanWindow,
-          textureSize: scannerState.size,
+          textureSize:
+              scannerState.deviceOrientation ==
```
navaronbracke: Use a switch expression on `scannerState.deviceOrientation` ?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -30,11 +30,15 @@ class DeviceOrientationListener(
...
+    var onDisplayRotationChanged: ((Int) -> Unit)? = null
```
navaronbracke, line 34: Should we set this to null? Or is that irrelevant with nulling out the instance of the class?

lib/src/overlay/barcode_overlay.dart
```diff
@@ -55,9 +59,24 @@ class _BarcodeOverlayState extends State<BarcodeOverlay> {
           return const SizedBox();
         }
 
+        // Mark stale when the device orientation changes,
+        // so the StreamBuilder discards its current snapshot.
+        if (_lastOrientation != null &&
+            _lastOrientation != value.deviceOrientation) {
+          _orientationChanged = true;
```
navaronbracke: No side effects in `build()` please. Can we fix this using a listener / stream transformer?

navaronbracke (changes requested):

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
juliansteenbakker: i think its cleaner to set to null at stop()

juliansteenbakker:
All issues are addressed @navaronbracke

navaronbracke (approved):
LGTM! Good idea of using a Key.

PR #1657, opened 2026-02-17
fix: ios overlay orientation

juliansteenbakker:
Merge after #1654

Fixes the position of the overlay on landscape


juliansteenbakker:
@navaronbracke this fixes the overlay position on ios landscape. I think after this fix, all issues with the overlay and rotation are solved. The changed build files are generated by the latest flutter version.

navaronbracke (approved):
LGTM. Thanks for the delegate changes as well.

PR #1662, opened 2026-02-19
feat: web camera improvements

juliansteenbakker:
This PR is a continuation of #1517, with some other improvements.

From the changelog:

> * [Web] The preferred camera device ID is now persisted in localStorage and reused on the next start.
> * [Web] Focus, exposure, and white-balance constraints are now applied automatically when supported by the browser (Image Capture API).
> * [Web] The camera now requests 1920×1080 as the ideal resolution for improved barcode detection.
> * [Web] The barcode overlay is now mirrored when the video preview is mirrored (e.g. front camera).

### DIFF
--- a/lib/src/web/mobile_scanner_web.dart
+++ b/lib/src/web/mobile_scanner_web.dart
@@ -185,7 +185,7 @@ class MobileScannerWeb extends MobileScannerPlatform {
   String? _getStoredDeviceId() {
     try {
       return window.localStorage.getItem(_kPreferredDeviceIdKey);
-    } on Object catch (_) {
+    } on DOMException catch (_) {
       return null;
     }
   }
@@ -194,7 +194,7 @@ class MobileScannerWeb extends MobileScannerPlatform {
   void _storeDeviceId(String deviceId) {
     try {
       window.localStorage.setItem(_kPreferredDeviceIdKey, deviceId);
-    } on Object catch (_) {
+    } on DOMException catch (_) {
       // Ignore — e.g. Safari private browsing mode disables storage.
     }
   }
@@ -208,7 +208,7 @@ class MobileScannerWeb extends MobileScannerPlatform {
       return devices.any(
         (d) => d.kind == 'videoinput' && d.deviceId == deviceId,
       );
-    } on Object catch (_) {
+    } on DOMException catch (_) {
       return false;
     }
   }
--- a/lib/src/web/mobile_scanner_web.dart
+++ b/lib/src/web/mobile_scanner_web.dart
@@ -256,6 +256,10 @@ class MobileScannerWeb extends MobileScannerPlatform {
 
     final capabilities = mediaDevices.getSupportedConstraints();
 
+    // Request high resolution for better barcode detection.
+    final width = ConstrainULongRange(ideal: 1920);
+    final height = ConstrainULongRange(ideal: 1080);
+
     var useStoredDevice = false;
     final MediaStreamConstraints constraints;
 
@@ -268,16 +272,26 @@ class MobileScannerWeb extends MobileScannerPlatform {
 
       constraints = useStoredDevice
           ? MediaStreamConstraints(
-              video: MediaTrackConstraintSet(deviceId: storedDeviceId.toJS),
+              video: MediaTrackConstraintSet(
+                deviceId: storedDeviceId.toJS,
+                width: width,
+                height: height,
+              ),
             )
-          : MediaStreamConstraints(video: true.toJS);
+          : MediaStreamConstraints(
+              video: MediaTrackConstraintSet(width: width, height: height),
+            );
     } else {
       // facingMode is supported (mobile). Always use it so that switching
       // between front and back cameras works correctly.
       final facingMode = _settingsDelegate.getFacingMode(cameraDirection);
 
       constraints = MediaStreamConstraints(
-        video: MediaTrackConstraintSet(facingMode: facingMode.toJS),
+        video: MediaTrackConstraintSet(
+          facingMode: facingMode.toJS,
+          width: width,
+          height: height,
+        ),
       );
     }
 
--- a/lib/src/web/mobile_scanner_web.dart
+++ b/lib/src/web/mobile_scanner_web.dart
@@ -239,7 +239,10 @@ class MobileScannerWeb extends MobileScannerPlatform {
   /// Throws a [MobileScannerException] if the permission was denied,
   /// or if using a video stream, with the given set of constraints, is
   /// unsupported.
-  Future<MediaStream> _prepareVideoStream(CameraFacing cameraDirection) async {
+  Future<MediaStream> _prepareVideoStream(
+    CameraFacing cameraDirection, {
+    Size? cameraResolution,
+  }) async {
     final mediaDevices = window.navigator.mediaDevicesNullable;
 
     if (mediaDevices == null) {
@@ -254,9 +257,12 @@ class MobileScannerWeb extends MobileScannerPlatform {
 
     final capabilities = mediaDevices.getSupportedConstraints();
 
-    // Request high resolution for better barcode detection.
-    final width = ConstrainULongRange(ideal: 1920);
-    final height = ConstrainULongRange(ideal: 1080);
+    final width = ConstrainULongRange(
+      ideal: cameraResolution?.width.toInt() ?? 1920,
+    );
+    final height = ConstrainULongRange(
+      ideal: cameraResolution?.height.toInt() ?? 1080,
+    );
 
     var useStoredDevice = false;
     final MediaStreamConstraints constraints;
@@ -442,7 +448,10 @@ class MobileScannerWeb extends MobileScannerPlatform {
     );
 
     // Request camera permissions and prepare the video stream.
-    final videoStream = await _prepareVideoStream(startOptions.cameraDirection);
+    final videoStream = await _prepareVideoStream(
+      startOptions.cameraDirection,
+      cameraResolution: startOptions.cameraResolution,
+    );
 
     try {
       // Clear the existing barcodes.
--- a/lib/src/web/mobile_scanner_web.dart
+++ b/lib/src/web/mobile_scanner_web.dart
@@ -162,7 +162,9 @@ class MobileScannerWeb extends MobileScannerPlatform {
   /// or if using a video stream, with the given set of constraints, is
   /// unsupported.
   Future<MediaStream> _prepareVideoStream(CameraFacing cameraDirection) async {
-    if (window.navigator.mediaDevices.isUndefinedOrNull) {
+    final mediaDevices = window.navigator.mediaDevicesNullable;
+
+    if (mediaDevices == null) {
       throw const MobileScannerException(
         errorCode: MobileScannerErrorCode.unsupported,
         errorDetails: MobileScannerErrorDetails(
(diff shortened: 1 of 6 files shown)


juliansteenbakker:
@navaronbracke i think this is ready to merge now.

lib/src/web/mobile_scanner_web.dart
```diff
@@ -131,38 +135,117 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+    // On mobile browsers, the facing mode is reliably reported in settings.
+    // On desktop browsers, facingMode is never reported (null), because desktop
+    // cameras have no hardware facing mode. Fall back to the requested
+    // camera direction instead.
+    final facingMode = tracks.first.getSettings().facingModeNullable?.toDart;
 
-    // On MacOS, even though the facing mode is supported, it is not reported.
-    // Use the label for FaceTime cameras to detect the user facing webcam.
-    if (videoTrack.label.contains('FaceTime')) {
+    if (facingMode == 'user' || (facingMode == null)) {
```
navaronbracke: Thanks for cleaning this up!

navaronbracke: The second set of parens feel a bit redundant?

lib/src/web/zxing/zxing_barcode_reader.dart
```diff
@@ -103,6 +105,59 @@ final class ZXingBarcodeReader extends BarcodeReader {
     }
   }
 
+  /// Returns true if the video preview is currently mirrored horizontally,
+  /// meaning barcode corner coordinates must be flipped to match.
+  ///
+  /// Must stay in sync with the logic in `_maybeFlipVideoPreview`.
+  bool _shouldMirrorX() {
```
navaronbracke: Should we move `_maybeFlipVideoPreview()`, `_shouldMirrorX()` and `_mirrorBarcodeX()` into a utility class? That way, if we ever have a different barcode reader in the future, this can be reused.

And that way we can also enforce the contract between all three a bit better.

lib/src/web/mobile_scanner_web.dart
```diff
@@ -131,38 +135,117 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+  /// Silently ignores any errors — these constraints are best-effort.
+  Future<void> _applyFocusConstraints(MediaStreamTrack track) async {
+    try {
+      final caps = track.getCapabilities();
+      var hasConstraints = false;
+
+      final constraints = MediaTrackConstraints();
+
+      final focusModes = caps.focusMode.toDart.map((e) => e.toDart).toList();
+      if (focusModes.contains('continuous')) {
```
navaronbracke: Use constants for these strings? (continuous + single-shot)

lib/src/web/mobile_scanner_web.dart
```diff
@@ -131,38 +135,117 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+  ///
+  /// Silently ignores any errors — these constraints are best-effort.
+  Future<void> _applyFocusConstraints(MediaStreamTrack track) async {
+    try {
+      final caps = track.getCapabilities();
+      var hasConstraints = false;
+
+      final constraints = MediaTrackConstraints();
+
+      final focusModes = caps.focusMode.toDart.map((e) => e.toDart).toList();
```
navaronbracke: Use a Set? Not that it matters much

lib/src/web/mobile_scanner_web.dart
```diff
@@ -131,38 +135,117 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+          caps.whiteBalanceMode.toDart.map((e) => e.toDart).toList();
+      if (wbModes.contains('continuous')) {
+        constraints.whiteBalanceMode = 'continuous'.toJS;
+        hasConstraints = true;
+      }
+
+      if (!hasConstraints) return;
+
+      await track.applyConstraints(constraints).toDart;
+    } on Object catch (_) {
```
navaronbracke, line 178: Why is this on Object? (multiple times)

IIRC for things like the camera permission errors we could catch `DOMException`.

lib/src/web/mobile_scanner_web.dart
```diff
@@ -131,38 +135,117 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+      // Not supported on this browser or device.
+    }
+  }
+
+  /// Return the preferred camera device ID stored in localStorage, or null.
+  String? _getStoredDeviceId() {
```
navaronbracke: Can we move this stuff into a local storage utility class? (purely for organisation purposes)

lib/src/web/mobile_scanner_web.dart
```diff
@@ -172,28 +255,69 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
     try {
       // Retrieving the media devices requests the camera permission.
-      final videoStream =
-          await window.navigator.mediaDevices.getUserMedia(constraints).toDart;
+      final videoStream = await mediaDevices.getUserMedia(constraints).toDart;
+
+      // Apply focus, exposure and white-balance constraints if supported.
+      final videoTrack = videoStream.getVideoTracks().toDart.firstOrNull;
+      if (videoTrack != null) {
+        await _applyFocusConstraints(videoTrack);
```
navaronbracke: "_applyVideoConstraints" ? It is not only focus?

lib/src/web/mobile_scanner_web.dart
```diff
@@ -172,28 +255,69 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+
+        // Persist the device ID so the same camera is preferred next time.
+        final deviceId = videoTrack.getSettings().deviceIdNullable?.toDart;
+        if (deviceId != null) _storeDeviceId(deviceId);
+      }
 
       return videoStream;
     } on DOMException catch (error, stackTrace) {
+      // If the stored device ID failed, clear it so we don't retry it.
+      if (useStoredDevice) _storeDeviceId('');
```
navaronbracke: Can we call a dedicated remove method, instead of using the empty string?

lib/src/web/mobile_scanner_web.dart
```diff
@@ -323,7 +448,10 @@ class MobileScannerWeb extends MobileScannerPlatform {
     );
 
     // Request camera permissions and prepare the video stream.
-    final videoStream = await _prepareVideoStream(startOptions.cameraDirection);
+    final videoStream = await _prepareVideoStream(
+      startOptions.cameraDirection,
```
navaronbracke, line 418: Should we use named params for everything here?

navaronbracke (changes requested):
Good improvements! I have some feedback, though :P

lib/src/web/mobile_scanner_web.dart
juliansteenbakker: I think we can change all except `_applyFocusConstraints`,  browsers that don't implement getCapabilities() at all throw a JSException instead of DOMException.

lib/src/web/mobile_scanner_web.dart
juliansteenbakker: I don't think the current positional + named is bad, as `cameraResolution` is optional.

We could argue that named + required is better, but i think we should also change the following function if that is the case:

```dart
  /// Start the barcode reader and initialize the [videoStream].
  ///
  /// The [options] are used to configure the barcode reader.
  /// The [videoElement] will become the video output element.
  /// The [videoStream] is the input for the barcode reader and video preview
  /// element.
  Future<void> start(
    StartOptions options, {
    required HTMLVideoElement videoElement,
    required MediaStream videoStream,
  }) {
    throw UnimplementedError('start() has not been implemented.');
  }
```

lib/src/web/mobile_scanner_web.dart
```diff
@@ -61,6 +66,10 @@ class MobileScannerWeb extends MobileScannerPlatform {
   /// The video element for the camera view.
   late HTMLVideoElement _videoElement;
 
+  /// Storage for the preferred camera device ID across sessions.
+  final PreferredDeviceStorage _preferredDeviceStorage =
```
navaronbracke:
```suggestion
 static const PreferredDeviceStorage _preferredDeviceStorage = 
    const PreferredDeviceStorage();
```

lib/src/web/mobile_scanner_web.dart
navaronbracke: Since it is an empty catch clause, can we omit the `on Object` ? Or is that a lint issue?

lib/src/web/mobile_scanner_web.dart
```diff
@@ -172,28 +217,69 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+      final videoStream = await mediaDevices.getUserMedia(constraints).toDart;
+
+      // Apply focus, exposure and white-balance constraints if supported.
+      final videoTrack = videoStream.getVideoTracks().toDart.firstOrNull;
+      if (videoTrack != null) {
+        await _applyVideoConstraints(videoTrack);
+
+        // Persist the device ID so the same camera is preferred next time.
+        final deviceId = videoTrack.getSettings().deviceIdNullable?.toDart;
+        if (deviceId != null) _preferredDeviceStorage.write(deviceId);
```
navaronbracke:
```suggestion
        if (deviceId != null) {
          _preferredDeviceStorage.write(deviceId);
        }
```

lib/src/web/mobile_scanner_web.dart
```diff
@@ -172,28 +217,69 @@ class MobileScannerWeb extends MobileScannerPlatform {
...
+      }
 
       return videoStream;
     } on DOMException catch (error, stackTrace) {
+      // If the stored device ID failed, clear it so we don't retry it.
+      if (useStoredDevice) _preferredDeviceStorage.remove();
```
navaronbracke:
```suggestion
      if (useStoredDevice) {
        _preferredDeviceStorage.remove();
      }
```

navaronbracke (changes requested):
Just some nits.

Could we add https://dart.dev/tools/linter-rules/curly_braces_in_flow_control_structures to the lints? That would flag the few uses here too.

lib/src/web/mobile_scanner_web.dart
juliansteenbakker: we still need it, otherwise the scanner would crash, and we can still continue using the scanner without `_applyFocusConstraints`.

we could also do this, but it requires importing dart:js_interop for JSException

```dart
  } on DOMException catch (_) {
  } on JSException catch (_) {
```

navaronbracke: Ah so `} catch (_) {` is insufficient? We _neeed_ the `on Object`?

juliansteenbakker:
> Just some nits.
> 
> Could we add https://dart.dev/tools/linter-rules/curly_braces_in_flow_control_structures to the lints? That would flag the few uses here too.

It's already in all versions of `very_good_analysis` we support https://github.com/VeryGoodOpenSource/very_good_analysis/blob/606b4cba48a685580cf5bbc8e6b604c2fa4f591f/lib/analysis_options.8.0.0.yaml#L70

lib/src/web/mobile_scanner_web.dart
juliansteenbakker: to keep the linter happy: https://dart.dev/tools/linter-rules/avoid_catches_without_on_clauses

PR #1663, opened 2026-02-20
feat: web scan window

juliansteenbakker:
Merge after #1662 

Enables scan window for web.

PR #1664, opened 2026-02-20
feat: native web barcode api & zxing-wasm

juliansteenbakker:
Merge after #1663

This PR improves web implementation by adding support for native web barcode api, as well as the actively maintained https://github.com/Sec-ant/zxing-wasm. Support for the legacy zxing-js is still available, but should be deprecated in a later release.

Changelog:

> * [Web] Added support for three barcode detection backends, selectable via `MobileScannerPlatform.instance.setWebBarcodeReader(WebBarcodeReader reader)`:
>   * `WebBarcodeReader.auto` (default), uses the native `BarcodeDetector` API when available, and falls back to `zxing-wasm` otherwise.
>   * `WebBarcodeReader.barcodeDetector`, uses the W3C Shape Detection API (Chrome 83+, Edge 83+, Safari 17+). No external library is loaded.
>   * `WebBarcodeReader.zxingWasm`, uses zxing-wasm (ZXing C++ compiled to WebAssembly), which works in all modern browsers including Firefox.
>   * `WebBarcodeReader.zxingJs`, the legacy ZXing JavaScript backend, retained for backward compatibility.
> * [Web] Added `MobileScannerPlatform.instance.activeWebReader` to query which backend is currently active.
>


navaronbracke:
Kudos for getting us an update on this! Hopefully this will allow us to add more features down the line. Do let me know when this is ready for a review.

juliansteenbakker:
If you give a go on #1662 we can check #1663, and after that's merged we can tackle this one 😄

navaronbracke:
Wishful thinking but would zxing-cpp open the door to Windows / Linux support? (I don't have a Linux machine, nor a webcam enabled Windows, though)

lib/src/mobile_scanner_platform_interface.dart
```diff
@@ -75,6 +76,16 @@ abstract class MobileScannerPlatform extends PlatformInterface {
   /// This is only supported on the web.
   void setBarcodeLibraryScriptUrl(String scriptUrl) {}
 
+  /// Set the preferred barcode detection backend for the web platform.
+  ///
+  /// This is only supported on the web. On other platforms this is a no-op.
+  /// Must be called before [start].
+  void setWebBarcodeReader(WebBarcodeReader reader) {}
```
navaronbracke, line 83: Should this live under MobileScannerWeb specifically? We have similar options such as the script url setter.

This is a no-op on non-web anyway.

navaronbracke (changes requested):

lib/src/mobile_scanner_platform_interface.dart
juliansteenbakker: The script url setter actually lives on MobileScannerPlatform too, not on MobileScannerWeb, so this follows the existing pattern. The problem with moving it to MobileScannerWeb is that apps would need to cast MobileScannerPlatform.instance as MobileScannerWeb, which requires importing the web implementation. Since that file pulls in dart:ui_web and package:web, it doesn't compile on native targets, so cross-platform apps would be forced into conditional imports (and we'd have to publicly export the web implementation, which we currently don't).

PR #1672, opened 2026-03-04
fix: multi camera instance handling

### LINKED ISSUE #1631: Multi instance support
OS: Android 16 (One UI 8)
Flutter: 3.38.5
mobile_scanner: 7.1.4

I need to use 2 instance of mobile_scanner. 

Given 2 pages using mobile scanner in route stack. When pop the first page and using dispose, it impact the first page using mobile scanner too. 
How can i use 2 separated instance of mobile scanner ?

EArminjon:
Should fix https://github.com/juliansteenbakker/mobile_scanner/issues/1631


lib/src/mobile_scanner_controller.dart
```diff
@@ -466,7 +466,12 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
       _setupListeners();
 
       final viewAttributes = await MobileScannerPlatform.instance.start(
```
clementperineau: I checked this, in release mode, there is an issue when pushing the new page.

We can fix this by stoping the instance before trying to start a new instance
`await MobileScannerPlatform.instance.stop(); // Ensure the camera is stopped before starting a new instance.`

EArminjon:
Didn't work in release mode, WIP.

EArminjon:
> Didn't work in release mode, WIP.

done

juliansteenbakker:
I've decided to go a different route with the fix, for a few reasons:                                                                                                                                        
                                                                                                                                                                                                                          
  - Changing the MobileScannerPlatform.start()/dispose() signatures is a breaking change for platform implementations.                                                                                                                                                                   
  - Auto-stopping/restarting other instances from the method channel embeds navigation policy in the plugin. Whether a covered scanner should resume on pop is an app decision; the plugin's contract should just be that 
  disposing controller B never breaks controller A.                                                   
  - dispose() still called stop()/updateScanWindow(null) on shared state even when the disposed controller wasn't the active one, and the post-frame restart choreography is timing-sensitive (as the release-mode issue  
  you hit showed).
  - The web implementation ignored the new id/callbacks, so the original bug remained there.

#1740 tracks which controller currently owns the camera session and makes dispose() do nothing.

Starting a second controller while another is running will still throw controllerAlreadyInitialized; for stacked routes, stop the covered scanner and restart it in didPopNext.

PR #1673, opened 2026-03-04
fix: cannot start camera after pause on android

### LINKED ISSUE #1666: [7.2.0] After calling pause in Android, start cannot be invoked
It seems that in the commit  (https://github.com/juliansteenbakker/mobile_scanner/pull/1645/changes/64809055f5a4bcd858a6445669d69fd8169b0c16), the timing of assigning the isPause field is incorrect. It should be assigned after
        
```
if (camera?.cameraInfo != null && preview != null && surfaceProducer != null && !isPaused) {

// TODO: resume here for seamless transition
//            if (isPaused) {
//                resumeCamera()
//                val cameraDirection = getCameraLensFacing(camera)
//                mobileScannerStartedCallback(
//                  MobileScannerStartParameters(
//                    if (portrait) width else height,
//                    if (portrait) height else width,
//                    deviceOrientationListener.getOrientation().serialize(),
//                    sensorRotationDegrees,
//                    surfaceProducer!!.handlesCropAndRotation(),
//                    currentTorchState,
//                    surfaceProducer!!.id(),
//                    numberOfCameras ?: 0,
//                    cameraDirection
//                  )
//                )
//                return
//            }
            mobileScannerErrorCallback(AlreadyStarted())

            return
}

// This is the correct position
isPaused = false
```

quinton22:
Fixes #1666

### DIFF
--- a/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
+++ b/android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
@@ -365,7 +365,6 @@ class MobileScanner(
         this.returnImage = returnImage
         this.invertImage = invertImage
 
-        isPaused = false
         
         if (camera?.cameraInfo != null && preview != null && surfaceProducer != null && !isPaused) {
 
@@ -392,6 +391,8 @@ class MobileScanner(
 
             return
         }
+
+        isPaused = false
 
         lastScanned = null
         scanner = barcodeScannerFactory(barcodeScannerOptions)


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -365,7 +365,6 @@ class MobileScanner(
         this.returnImage = returnImage
         this.invertImage = invertImage
 
-        isPaused = false
```
navaronbracke, line 368: Looking at the if check below, this does indeed look like the assignment of isPaused is wrong here?

navaronbracke (approved):

PR #1696, opened 2026-04-17
fix: ensure capture output queue is not the main thread

### LINKED ISSUE #1694: [iOS] AVCaptureVideoDataOutput setSampleBufferDelegate on DispatchQueue.main causes sustained device heating
---                                                                                                                                                                                                                                  
  Environment                                                                                                                                                                                                                          
  - mobile_scanner: 7.2.0                                                                                                                                                                                                           
  - Platform: iOS 
  - Flutter: stable channel                                                                                                                                                                                                            
                                                                                                                                                                                                                                       
  ---                                                                                                                                                                                                                                  
  Description                                                                                                                                                                                                                          
                                                                                                                                                                                                                                       
  We identified sustained device heating on iOS traced directly to a single line in MobileScannerPlugin.swift:                                                                                                                         
                                                                                                                                                                                                                                       
  // MobileScannerPlugin.swift line 445                                                                                                                                                                                                
  videoOutput.setSampleBufferDelegate(self, queue: DispatchQueue.main)                                                                                                                                                                 
                                                                                                                                                                                                                                       
  Setting the sample buffer delegate queue to DispatchQueue.main means every camera frame (30fps+) is delivered as a callback on Flutter's main/UI thread. Inside each callback (captureOutput(_:didOutput:from:)),                    
  registry.textureFrameAvailable(textureId) is called, which schedules a Flutter texture refresh — also on the main thread. This creates continuous main thread pressure at 30fps while the scanner is active.                         
                                                                                                                                                                                                                                       
  Additionally, for each eligible frame, a VNDetectBarcodesRequest is dispatched on DispatchQueue.global(qos: .userInitiated) — a high-priority background thread. The combination of:                                                 
  - 30fps main thread callbacks                                                                                                                                                                                                        
  - High-QoS Vision framework threads running concurrently                                                                                                                                                                             
  - ISP and Neural Engine active for QR detection                                                                                                                                                                                      
                                                                                                                                                                                                                                       
  ...results in sustained CPU/ISP heating that is clearly measurable on physical iOS devices.                                                                                                                                          
                                                                                                                                                                                                                                       
  ---                                                                                                                                                                                                                                  
  How we found it                                                                                                                                                                                                                      
                                                                                                                                                                                                                                       
  We profiled a Flutter app with an always-on QR scanner screen (similar to a UPI payments flow). Steps to reproduce the heat pattern:                                                                                                 
                                                                                                                                                                                                                                       
  1. Open the screen with mobile_scanner active (camera permission granted)                                                                                                                                                            
  2. Leave it idle for ~3 minutes — device heats noticeably                                                                                                                                                                            
  3. Revoke camera permission → reopen same screen → heat drops significantly                                                                                                                                                          
  4. Re-grant camera permission → heat returns immediately                                                                                                                                                                             
                                                                                                                                                                                                                                       
  The delta between camera-on and camera-off heat was the primary signal. Reading MobileScannerPlugin.swift confirmed DispatchQueue.main as the queue for the sample buffer delegate.  

  ---                                                                                                                                                                                                                                  
  Relevant code                                                                                                                                                                                                                     

  // Line 445 — problematic                                                                                                                                                                                                            
  videoOutput.setSampleBufferDelegate(self, queue: DispatchQueue.main)                                                                                                                                                                 
                                                                                                                                                                                                                                       
  // Line 164 — Vision dispatched to background (this part is correct)                                                                                                                                                                 
  DispatchQueue.global(qos: .userInitiated).async { [weak self] in                                                                                                                                                                     
      // VNDetectBarcodesRequest...                                                                                                                                                                                                    
  }                                                                                                                                                                                                                                 
                                                                                                                                                                                                                                       
  Apple's own documentation for setSampleBufferDelegate(_:queue:) states:                                                                                                                                                              
                                                                                                                                                                                                                                       
  ▎ "You should use a serial dispatch queue to ensure that video frames are delivered in order. Use a queue other than the main queue to avoid blocking the main thread."                                                              
                                                                                                                                                                                                                                       
  ---                                                                                                                                                                                                                                  
  Impact                                                                                                                                                                                                                            

  - Sustained device heating whenever the scanner is active
  - Main thread contention between camera frame callbacks and Flutter UI rendering                                                                                                                                                     
  - Affects any Flutter app using mobile_scanner on iOS with an always-visible scanner                                                                                                                                                 
                                                                                                                                                                                                                                       
  ---                                                                                                                                                                                                                                  
  Proposed fix                                                                                                                                                                                                                         
                                                                                                                                                                                                                                    
  Use a dedicated serial dispatch queue for the sample buffer delegate:                                                                                                                                                                
                                                                                                                                                                                                                                       
  // In the scanner setup, declare a session queue:                                                                                                                                                                                    
  private let sessionQueue = DispatchQueue(label: "com.juliansteenbakker.mobile_scanner.session", qos: .userInitiated)                                                                                                                 
                                                                                                                                                                                                                                       
  // Replace line 445 with:                                                                                                                                                                                                            
  videoOutput.setSampleBufferDelegate(self, queue: sessionQueue)                                                                                                                                                                       
                                                                                                                                                                                                                                       
  This moves all frame callbacks off the main thread while preserving ordering (serial queue). The existing Vision dispatch to .global(qos: .userInitiated) is already correct and would remain unchanged.                             
                                                                                                                                                                                                                                       
  Note: registry.textureFrameAvailable(textureId) may need to be dispatched back to the main thread explicitly if Flutter's texture registry requires it — worth verifying in your FlutterTextureRegistry implementation.

navaronbracke:
This pull request fixes a reported overheating issue on device, since `captureOutput` was being called on the main thread.

Tested on MacOS and scanning still works.

Also fixed 2 warnings in code:
- an unused do/catch, because there was no `try` (Star Wars jokes, anyone?)
- an `#if def` was complaining about a return, because we forgot to configure the `#else` after it

Fixes https://github.com/juliansteenbakker/mobile_scanner/issues/1694

cc @juliansteenbakker Since this should fix an overheating issue I'd like to publish 7.2.1 with this fix included.

cc @sawan-scapia Could you confirm the overheating issue is gone, with this patch? I only managed to verify on MacOS at this time.

### DIFF
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -894,8 +894,7 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
             details: nil
         ))
         return
-#endif
-
+#else
         let argReader = MapArgumentReader(call.arguments as? [String: Any])
         let symbologies:[VNBarcodeSymbology] = argReader.toSymbology()
 
@@ -971,6 +970,7 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
                     message: error.localizedDescription, details: nil))
             }
         }
+#endif
     }
 
     // Observer for torch state
--- a/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
+++ b/darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
@@ -37,6 +37,11 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
     /// This is static to avoid accessing `self` in the `VNDetectBarcodesRequest` callback.
     private static var returnImage: Bool = false
 
+    /// The dispatch queue that will process ``captureOutput(_:didOutput:from:)`` calls.
+    ///
+    /// This queue is user initiated because it is tied to an ``AVCaptureSession``.
+    private let sampleBufferQueue = DispatchQueue(label: "com.juliansteenbakker.mobile_scanner.captureOutputQueue", qos: .userInitiated)
+
     var detectionSpeed: DetectionSpeed = DetectionSpeed.noDuplicates
 
     var timeoutSeconds: Double = 0
@@ -443,7 +448,7 @@ public class MobileScannerPlugin: NSObject, FlutterPlugin, FlutterStreamHandler,
         let format = getPreferredVideoFormat(videoOutput: videoOutput)
         videoOutput.videoSettings = [kCVPixelBufferPixelFormatTypeKey as String: format]
         videoOutput.alwaysDiscardsLateVideoFrames = true
-        videoOutput.setSampleBufferDelegate(self, queue: DispatchQueue.main)
+        videoOutput.setSampleBufferDelegate(self, queue: self.sampleBufferQueue)
         captureSession!.addOutput(videoOutput)
         let deviceVideoOrientation = self.getVideoOrientation()
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 **Bug Fixes**
 
 * [Apple] Fixed a race condition in `captureOutput` where `latestBuffer` was read on a background thread after being deallocated on the main thread, causing a crash in `VTCreateCGImageFromCVPixelBuffer`.
+* [Apple] Fixed an issue where `captureOutput` was being processed on the main thread, causing overheating issues on device.
 
 ## 7.2.0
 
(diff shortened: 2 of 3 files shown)


darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerPlugin.swift
navaronbracke, line 891: If this if def is currently active in XCode, I got a warning about everything after the return being unreachable

sawan-scapia (approved):
The camera heating issue has improved, with measurable gains observed using the Flutter thermal package. There is still a slight amount of heat, but this looks good for now. Approving this as a good to have improvement.

navaronbracke:
@sawan-scapia Would you be able to do an analysis to figure out where the remaining heating issues could come from?

sawan-scapia:
> @sawan-scapia Would you be able to do an analysis to figure out where the remaining heating issues could come from?

Yes currently working on it , will update the thread for any future findings

Issue #1536, opened 2025-09-23
Scanning incorrect code intermittently


1205036:
I tried many settings such as higher camera resolution and unrestricted speed, the reader seems to miss read barcodes. In the attached example this is a CODE39 barcode. 
It is easily reproducible just move your camera a bit and you can get an incorrect reading.
It becomes way worse on bad lightning conditions.

![Image](https://github.com/user-attachments/assets/7b9300b1-62c5-40bd-96be-4317a021ef77)

https://github.com/user-attachments/assets/9a48bf9a-bf06-4fba-8954-e690f40ffd73

alexleigh0:
This is happening for me too

azafrak:
I was experiencing the same issue. I had previously been using barcode_scanner2 in my project, but when I came across mobile_scanner while developing a new mobile application, I decided to migrate my existing barcode scanning implementations to this library as well.

After some time, however, I noticed that mobile_scanner was having problems reading certain barcode types. As of today, I had to switch back to barcode_scanner2, and Code 39 barcodes are now being scanned without any issues again.

PR #1704, opened 2026-05-15
feat: add allowedLengths filter to MobileScannerController

rabend:
Allows callers to restrict scanner output to barcodes whose rawValue.length is in a given set. When every barcode in a capture is dropped, the capture itself is suppressed. Also applied to analyzeImage results.

This is primarily useful for formats without a checksum or end-of-barcode indicator (e.g. ITF / Interleaved 2 of 5 or CODE39), where the decoder can return a truncated prefix of the actual code.

This is related to issue #1536

### DIFF
--- a/lib/src/mobile_scanner_controller.dart
+++ b/lib/src/mobile_scanner_controller.dart
@@ -33,6 +33,7 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
     int detectionTimeoutMs = 250,
     this.facing = CameraFacing.back,
     this.formats = const <BarcodeFormat>[],
+    this.allowedLengths = const <int>{},
     this.returnImage = false,
     this.torchEnabled = false,
     this.invertImage = false,
@@ -102,6 +103,25 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
   ///
   /// If this is empty, all supported formats are detected.
   final List<BarcodeFormat> formats;
+
+  /// The set of allowed `rawValue` lengths for scanned barcodes.
+  ///
+  /// When this set is non-empty, any barcode whose `rawValue.length` is not
+  /// contained in it is dropped before being delivered to listeners.
+  /// If every barcode in a capture is dropped, the capture event itself is
+  /// suppressed and no [BarcodeCapture] is emitted for that frame.
+  ///
+  /// This is primarily useful for formats that have no checksum or
+  /// end-of-barcode indicator, such as ITF (Interleaved 2 of 5), where the
+  /// decoder may return a truncated prefix of the actual code. For example,
+  /// a scanner configured for ITF-14 can set `allowedLengths: {14}` to drop
+  /// any partial reads.
+  ///
+  /// Barcodes whose `rawValue` is null are dropped when this set is
+  /// non-empty, because their length cannot be validated.
+  ///
+  /// If this set is empty (the default), no length filtering is applied.
+  final Set<int> allowedLengths;
 
   /// Whether the [BarcodeCapture.image] bytes should be provided.
   ///
@@ -175,7 +195,12 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
               return;
             }
 
-            _barcodesController.add(barcode);
+            final filtered = _applyAllowedLengths(barcode);
+            if (filtered == null) {
+              return;
+            }
+
+            _barcodesController.add(filtered);
           },
           onError: (Object error) {
             if (_barcodesController.isClosed) {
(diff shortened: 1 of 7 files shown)


WladmirJunior (approved):

navaronbracke:
What is the use case here? This is essentially more filtering on the raw value, to avoid bad reads?

lib/src/mobile_scanner_controller.dart
```diff
@@ -103,6 +104,25 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+  /// end-of-barcode indicator, such as ITF (Interleaved 2 of 5), where the
+  /// decoder may return a truncated prefix of the actual code. For example,
+  /// a scanner configured for ITF-14 can set `allowedLengths: {14}` to drop
+  /// any partial reads.
+  ///
+  /// Barcodes whose `rawValue` is null are dropped when this set is
+  /// non-empty, because their length cannot be validated.
+  ///
+  /// If this set is empty (the default), no length filtering is applied.
+  final Set<int> allowedLengths;
```
navaronbracke, line 124: What happens if one forgets to add a specific length in the set? It is dropped?

This feels very error prone to me. Is a "minimum length" (per format) better?

lib/src/mobile_scanner_controller.dart
```diff
@@ -294,8 +319,52 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+  }
+
+  /// Apply [allowedLengths] to [capture].
+  ///
+  /// Returns [capture] unchanged when [allowedLengths] is empty.
+  /// Returns a new [BarcodeCapture] with the filtered barcodes when at least
+  /// one barcode passes the filter.
+  /// Returns null when every barcode in [capture] is dropped, so the caller
+  /// can suppress the capture event entirely.
+  BarcodeCapture? _applyAllowedLengths(BarcodeCapture capture) {
```
navaronbracke: Should this live in BarcodeCapture or a utility? It feels odd to put this here, from a code-maintenance perspective.

lib/src/mobile_scanner_controller.dart
```diff
@@ -294,8 +319,52 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+    if (allowedLengths.isEmpty) {
+      return capture;
+    }
+
+    final filtered = capture.barcodes
```
navaronbracke: Use a for literal?

lib/src/mobile_scanner_controller.dart
```diff
@@ -294,8 +319,52 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+        .where((barcode) {
+          final rawValue = barcode.rawValue;
+          return rawValue != null && allowedLengths.contains(rawValue.length);
```
navaronbracke: Does this accidentally strip out everything when allowedLengths is empty? (which is the default)

lib/src/mobile_scanner_controller.dart
```diff
@@ -294,8 +319,52 @@ class MobileScannerController extends ValueNotifier<MobileScannerState> {
...
+
+    if (filtered.isEmpty) {
+      return null;
+    }
+
+    if (filtered.length == capture.barcodes.length) {
+      return capture;
+    }
+
+    return BarcodeCapture(
```
navaronbracke: Should this live in a copyWith() implementation?

test/mobile_scanner_controller/allowed_lengths_test.dart
```diff
@@ -0,0 +1,246 @@
+import 'dart:async';
+
+import 'package:flutter/material.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mobile_scanner/mobile_scanner.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  Future<(MobileScannerController, _FakeMobileScannerPlatform)> pumpScanner(
```
navaronbracke: This feels a little odd? There is no need for a record as return value

test/mobile_scanner_controller/allowed_lengths_test.dart
```diff
@@ -0,0 +1,246 @@
...
+    WidgetTester tester, {
+    required Set<int> allowedLengths,
+  }) async {
+    final fakePlatform = _FakeMobileScannerPlatform();
```
navaronbracke: This should be in a setUp()

test/mobile_scanner_controller/allowed_lengths_test.dart
```diff
@@ -0,0 +1,246 @@
...
+    MobileScannerPlatform.instance = fakePlatform;
+
+    final controller = MobileScannerController(allowedLengths: allowedLengths);
+
+    await tester.pumpWidget(
+      MaterialApp(home: Scaffold(body: MobileScanner(controller: controller))),
```
navaronbracke: Can we avoid using Material in this test? (The framework is in the proces of decoupling Material, this would give mobile_scanner one less case to fix)

test/mobile_scanner_controller/allowed_lengths_test.dart
```diff
@@ -0,0 +1,246 @@
...
+      await tester.pumpAndSettle();
+
+      expect(received, hasLength(1));
+      expect(received.single.barcodes.map((b) => b.rawValue).toList(), const [
+        '12345678901234',
+      ]);
+    },
+  );
+
+  testWidgets('allowedLengths suppresses captures with no surviving barcodes', (
```
navaronbracke: This test is the same as the one above? I would expect a test case where no barcodes survived, so the list is empty

test/mobile_scanner_controller/allowed_lengths_test.dart
```diff
@@ -0,0 +1,246 @@
...
+      ),
+    );
+    await tester.pumpAndSettle();
+
+    expect(received, hasLength(1));
+    expect(received.single.barcodes, hasLength(1));
+    expect(received.single.barcodes.single.rawValue, '12345678901234');
+  });
+
+  testWidgets('analyzeImage applies allowedLengths to the platform result', (
```
navaronbracke: Should we have a specific test suite that covers analyzeImage() only?

test/mobile_scanner_controller/allowed_lengths_test.dart
```diff
@@ -0,0 +1,246 @@
...
+        barcodes: [Barcode(rawValue: '12345')],
+      );
+
+      final result = await controller.analyzeImage('ignored.png');
+
+      expect(result, isNull);
+    },
+  );
+
+  testWidgets('analyzeImage propagates a null platform result', (tester) async {
```
navaronbracke: The purpose of this test is not specifically related to this PR?

navaronbracke (changes requested):

navaronbracke:
@WladmirJunior Friendly reminder to not randomly approve a pull request. I did not yet have time to finish my review until now. I do understand that you might be interested in this feature, but then I would advise giving a thumbs up on the original issue.

rabend:
> What is the use case here? This is essentially more filtering on the raw value, to avoid bad reads?

Hi @navaronbracke! Yes, @WladmirJunior and I are working on a scanner and had issues with partial reads of barcodes. This can happen specifically with the ITF format, as it doesn't support a completeness checksum or has any indicator for the end of the barcode. So instead of scanning the complete 44 digit code, we often saw 20-something codes being scanned. I already implemented a fix for this on our side, but thought it would be nice to have this capability directly inside the lib. Issue #1536 is directly related, there this happens with the CODE39 format.

I'll work on your comments during the next days, whenever I have some time. Thanks for the review!

lib/src/mobile_scanner_controller.dart
rabend: I feel like the user usually knows what codes he wants to scan, so an explicit length fits better. For example when dealing with ISBNs for books, these are either 13 or 10 digits, but never 14 (like an EAN/GTIN).

PR #1707, opened 2026-05-22
feat: bundled ci street, adding concurrency and improved caching

juliansteenbakker:
This PR updates the CI street to enable concurrency.


lib/src/web/mobile_scanner_web.dart
```diff
@@ -192,7 +192,7 @@ class MobileScannerWeb extends MobileScannerPlatform {
       return devices.any(
         (d) => d.kind == 'videoinput' && d.deviceId == deviceId,
       );
-    } on DOMException catch (_) {
```
navaronbracke, line 195: Any specific reason these were changed?

navaronbracke (approved):

navaronbracke:
Looks like there is an analysis error. Also, fyi: RadioGroup is new in 3.32

lib/src/web/mobile_scanner_web.dart
juliansteenbakker: Catch clause with type 'DOMException' checks whether the caught value is a JS interop type, which might not be platform-consistent.

However, i didn't find any good way to catch the specific exception across wasm and js, so that's were the "on Object" comes from.

PR #1710, opened 2026-06-02
fix: getSupportedLenses only enumerates logical cameras; add optional facing filter

marnberg:
### Problem 1: physical camera enumeration crash

`getSupportedLenses` previously recursed into `physicalCameraIds` on Android 9+ to detect all lenses in a multi-camera group. This caused it to report lens types (e.g. zoom) that CameraX cannot actually select — CameraX operates at the **logical** camera level, not the physical sub-camera level. When `switchCamera` was called with one of these phantom lens types, `selectCamera()` found no matching logical camera and (before this PR) silently fell back to returning all cameras, causing CameraX to bind the wrong camera and crash.

**Fix:** enumerate only entries in `cameraManager.cameraIdList` (logical cameras). Logical cameras expose their own `LENS_INFO_AVAILABLE_FOCAL_LENGTHS` and `SENSOR_INFO_PHYSICAL_SIZE`, which are sufficient for 35mm-equivalent classification. Devices that expose zoom as a standalone logical camera ID (rather than only as a physical sub-camera) are correctly detected.

Also removes the silent fallback in `selectCamera()` that returned all cameras when no lens type matched. It now returns an empty list so CameraX throws a proper error instead of silently binding the wrong camera.

### Problem 2: front-camera lenses polluting back-camera results

Without a facing filter, `getSupportedLenses()` returned lenses from both the front and back cameras combined. Callers checking whether a zoom lens is available for the back camera would get a false positive if only the front camera had a zoom lens (or vice versa).

**Fix:** add an optional `facing` parameter (`null` = all cameras, keeps existing behaviour). The Android handler passes it as a `LENS_FACING` filter; iOS filters by `AVCaptureDevice.Position`; Web ignores it (single logical "camera").

## API change

```dart
// Before (unchanged behaviour):
Future<Set<CameraLensType>> getSupportedLenses()

// After (backwards-compatible):
Future<Set<CameraLensType>> getSupportedLenses({CameraFacing? facing})
```

Passing no argument continues to return lenses for all cameras, so existing callers are unaffected.

## Test plan

- [ ] `fvm flutter test test/method_channel/get_supported_lenses_test.dart` — existing tests + 5 new facing-filter tests
- [ ] `fvm flutter test test/mobile_scanner_controller/get_supported_lenses_test.dart` — existing tests + facing-forwarding test
- [ ] `fvm flutter test test/mobile_scanner_controller/switch_camera_test.dart` — no regressions
- [ ] Manual: on a multi-camera Android device, confirm `getSupportedLenses(facing: CameraFacing.back)` no longer returns zoom when zoom is only available on the front camera (or as a non-selectable physical sub-camera)
- [ ] Manual: confirm switching to a lens type returned by `getSupportedLenses(facing: CameraFacing.back)` does not crash


test/method_channel/get_supported_lenses_test.dart
```diff
@@ -360,4 +361,135 @@ void main() {
...
+            return null;
+          });
+
+      await platform.getSupportedLenses();
+
+      expect(capturedCall, isNotNull);
+      expect(capturedCall!.arguments, isNull);
+    });
+
+    test('returns only back-camera lenses when back facing passed', () async {
```
navaronbracke, line 441: Should we add a similar test for front facing lenses?

navaronbracke (approved):
LGTM with a nit about a test.

Can you also add something in the changelog, to mention the bugfix?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -298,11 +298,14 @@ class MobileScannerHandler(
     }
 
     /**
-     * Get the list of supported lens types on this device.
+     * Get the list of supported lens types on this device for a given facing direction.
+     *
+     * Accepts an optional "facing" argument (0 = front, 1 = back). Defaults to back.
```
navaronbracke, line 303: The default is null, so this is wrong?

```suggestion
     * Accepts an optional "facing" argument (0 = front, 1 = back).
```

test/method_channel/get_supported_lenses_test.dart
```diff
@@ -360,4 +361,164 @@ void main() {
...
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+          .setMockMethodCallHandler(platform.methodChannel, (
+            methodCall,
+          ) async {
+            if (methodCall.method ==
+                MethodChannelMobileScanner.kGetSupportedLensesMethodName) {
+              final args = methodCall.arguments as Map?;
```
navaronbracke, line 448: This can be an if-case expression

test/method_channel/get_supported_lenses_test.dart
```diff
@@ -360,4 +361,164 @@ void main() {
...
+
+    test('returns only front-camera lenses when front facing passed', () async {
+      TestDefaultBinaryMessengerBinding.instance.defaultBinaryMessenger
+          .setMockMethodCallHandler(platform.methodChannel, (
+            methodCall,
+          ) async {
+            if (methodCall.method ==
+                MethodChannelMobileScanner.kGetSupportedLensesMethodName) {
+              final args = methodCall.arguments as Map?;
+              if (args?['facing'] == 0) {
```
navaronbracke, line 478: Same remark for the if case

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 
 * [Apple] Fixed a race condition in `captureOutput` where `latestBuffer` was read on a background thread after being deallocated on the main thread, causing a crash in `VTCreateCGImageFromCVPixelBuffer`.
 * [Apple] Fixed an issue where `captureOutput` was being processed on the main thread, causing overheating issues on device.
+* [Apple] Fixed `getSupportedLenses` enumerating constituent physical cameras of virtual devices, which returned duplicate/unavailable lenses. It now only enumerates logical cameras. Also added an optional `facing` filter to `getSupportedLenses`.
```
navaronbracke, line 15:
```suggestion
* [Apple] Fixed `getSupportedLenses` enumerating constituent physical cameras of virtual devices, which returned duplicate/unavailable lenses. It now only enumerates logical cameras.
```

navaronbracke, line 15: The line `Also added an optional facing filter to getSupportedLenses.` is not exclusive to Apple platforms

navaronbracke (changes requested):

juliansteenbakker:
I created the following PR with these fixes: #1741

- Implemented the facing filter on iOS by passing an optional AVCaptureDevice.Position into the discovery session loop. macOS and the web ignore the filter; the README and changelog now say so.
- Removed a duplicate @override annotation in the method channel.
- Relabeled the changelog entry as [Android] and documented the removal of the silent default-camera fallback as a behavior change.
- Fixed stale docs: the Kotlin kdoc still mentioned physical sub-camera enumeration, and the handler doc claimed the facing argument defaults to back.
- Applied the if-case pattern suggestions from review in the method channel tests.

PR #1711, opened 2026-06-02
feat: add getBestQrScanningLens() to select optimal lens for QR scanning

marnberg:
## Summary

Adds `getBestQrScanningLens()` — a method that automatically selects the camera lens best suited for close-range QR code scanning, based on minimum focus distance data from the platform.

```dart
Future<CameraLensType> getBestQrScanningLens({
  CameraFacing facing = CameraFacing.back,
})
```

### Platform behaviour

**iOS 15+:** Iterates all selectable `AVCaptureDevice` instances for the given position and returns the one with the lowest `minimumFocusDistance` (millimeters; lower = can focus closer). Fixed-focus cameras (`minimumFocusDistance == -1`) are skipped. On newer iPhones this tends to return the ultra-wide lens, which has macro autofocus capability down to ~2 cm. On older models it returns the standard 1× camera.

**iOS < 15 / macOS:** Falls back to `CameraLensType.normal` (the standard 1× camera).

**Android:** Always returns `CameraLensType.normal`. `LENS_INFO_MINIMUM_FOCUS_DISTANCE` is not reliably populated for logical cameras on modern multi-camera devices — the data lives on physical sub-cameras that are not independently selectable via CameraX. The normal camera has the most capable autofocus system for close-range scanning on virtually all Android devices.

**Web:** Returns `CameraLensType.normal`. The MediaDevices API has no concept of lens type or minimum focus distance.

### Typical usage

```dart
final bestLens = await controller.getBestQrScanningLens(
  facing: CameraFacing.back,
);

// Check that the lens is actually available before switching
// (pairs with #1710 — getSupportedLenses facing filter)
final supported = await controller.getSupportedLenses(
  facing: CameraFacing.back,
);

if (supported.contains(bestLens)) {
  await controller.switchCamera(
    SelectCamera(facingDirection: CameraFacing.back, lensType: bestLens),
  );
}
```

> **Note:** the usage example above references `getSupportedLenses(facing:)`, which is proposed in PR #1710. The two PRs are code-independent and can be merged in any order.

## Test plan

- [ ] `fvm flutter test test/method_channel/get_best_qr_scanning_lens_test.dart` — 7 tests: raw value mapping (0/1/2/null), facing argument forwarding (back/front), correct method name
- [ ] `fvm flutter test test/mobile_scanner_controller/get_best_qr_scanning_lens_test.dart` — 4 tests: disposed guard, result forwarding, facing parameter pass-through
- [ ] Manual on iOS 15+ device: confirm the method returns a lens that focuses closer than the default (hold a QR code at ~5 cm and verify it scans without switching manually)
- [ ] Manual on Android: confirm `CameraLensType.normal` is returned and camera starts without crash


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerCameraLensSelector.kt
```diff
@@ -259,6 +259,28 @@ object MobileScannerCameraLensSelector {
...
+     * Determine the best lens type for QR code scanning.
+     *
+     * On Android, [LENS_INFO_MINIMUM_FOCUS_DISTANCE] is not reliably populated for
+     * logical cameras on modern multi-camera devices — it is only present on physical
+     * sub-cameras, which are not independently selectable via CameraX. Because of this,
+     * the normal (main) camera is always returned: it has the most capable autofocus
+     * system and handles close-range QR scanning well on virtually all Android devices.
+     * Wide-angle cameras frequently have fixed focus or slower autofocus.
+     *
+     * For platforms where minimum focus distance data is available per selectable camera
```
navaronbracke: This pargraph is irrelevant for the Android implementation?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerCameraLensSelector.kt
```diff
@@ -259,6 +259,28 @@ object MobileScannerCameraLensSelector {
...
+     * (e.g. iOS via AVCaptureDevice.minimumFocusDistance), the platform implementation
+     * performs a genuine comparison and may return a different lens type.
+     *
+     * @param cameraManager The CameraManager instance (unused, kept for API symmetry)
+     * @param facing 0 = front, 1 = back (default); currently unused, normal is returned for both
+     * @return [LENS_TYPE_NORMAL]
+     */
+    fun getBestQrScanningLens(@Suppress("UNUSED_PARAMETER") cameraManager: CameraManager, facing: Int = 1): Int {
```
navaronbracke: Can you remove the unused argument here?

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScannerHandler.kt
```diff
@@ -300,6 +301,20 @@ class MobileScannerHandler(
     /**
      * Get the list of supported lens types on this device.
      */
+    private fun getBestQrScanningLens(call: MethodCall, result: MethodChannel.Result) {
+        val facing: Int = call.argument<Int>("facing") ?: 1
```
navaronbracke, line 307: Should we document it defaults to back facing if not provided?

android/src/test/kotlin/dev/steenbakker/mobile_scanner/MobileScannerCameraLensSelectorTest.kt
```diff
@@ -282,4 +282,21 @@ internal class MobileScannerCameraLensSelectorTest {
...
+    // ==========================================================================
+    //
+    // On Android, LENS_INFO_MINIMUM_FOCUS_DISTANCE is not reliably populated for
+    // logical cameras on modern multi-camera devices (data lives on physical
+    // sub-cameras which are not independently selectable via CameraX). Therefore
+    // getBestQrScanningLens always returns LENS_TYPE_NORMAL — the main camera has
+    // the most capable autofocus on virtually all Android devices.
+
+    @Test
+    fun getBestQrScanningLens_alwaysReturnsNormal() {
```
navaronbracke: This test is wrong? It never calls `MobileScannerCameraLensSelector.getBestQrScanningLens()`, but the test name implies that.

Also, the constant is tested on line 280

darwin/mobile_scanner/Sources/mobile_scanner/MobileScannerCameraSelector.swift
```diff
@@ -106,6 +106,56 @@ class MobileScannerCameraSelector {
...
+    ///
+    /// Falls back to `LensType.wideAngle` (raw value 0 = Dart `normal`) on iOS < 15 or macOS,
+    /// which is the standard main camera and a safe default for QR scanning.
+    ///
+    /// - Parameter position: The camera position to check (default: .back)
+    /// - Returns: The LensType raw value best suited for close-up QR scanning
+    static func getBestQrScanningLens(position: AVCaptureDevice.Position = .back) -> Int {
+#if os(iOS)
+        if #available(iOS 15.0, *) {
+            if #available(iOS 13.0, *) {
```
navaronbracke: `if #available(iOS 13.0, *) {` is redundant, since you require iOS 15 just before it?

test/mobile_scanner_controller/get_best_qr_scanning_lens_test.dart
```diff
@@ -0,0 +1,120 @@
...
+            (e) => e.errorCode,
+            'errorCode',
+            MobileScannerErrorCode.controllerDisposed,
+          ),
+        ),
+      );
+    });
+
+    test('returns best lens from platform', () async {
+      MobileScannerPlatform.instance = FakeMobileScannerPlatform(
```
navaronbracke, line 35: This test isn't actually any different from "returns normal lens as default from platform" ?
I think this is more of a readability issue with the default parameter?

What if you provide the default parameter in the other test, with a lint ignore for the argument matching the default?

navaronbracke (changes requested):

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -513,6 +518,17 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
     return lensTypes.whereType<int>().map(CameraLensType.fromRawValue).toSet();
   }
 
+  @override
+  Future<CameraLensType> getBestQrScanningLens({
```
navaronbracke, line 522: I'm a bit concerned with this API, in relation to devices that have no cameras.

I think that should be handled as well, so it is in line with the `unsupported` state elsewhere in the plugin.

lib/src/web/mobile_scanner_web.dart
```diff
@@ -352,6 +352,15 @@ class MobileScannerWeb extends MobileScannerPlatform {
     }
   }
 
+  @override
+  Future<CameraLensType> getBestQrScanningLens({
+    CameraFacing facing = CameraFacing.back,
+  }) async {
+    // The web MediaDevices API has no concept of lens type or minimum focus
+    // distance, so the facing parameter is intentionally unused.
+    return CameraLensType.normal;
```
navaronbracke, line 361: What if the camera is unavailable? Then this gives a false positive?

test/mobile_scanner_controller/get_best_qr_scanning_lens_test.dart
```diff
@@ -0,0 +1,112 @@
...
+import 'package:mobile_scanner/src/mobile_scanner_controller.dart';
+import 'package:mobile_scanner/src/mobile_scanner_exception.dart';
+import 'package:mobile_scanner/src/mobile_scanner_platform_interface.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  group('getBestQrScanningLens', () {
+    test('throws when controller is disposed', () async {
+      MobileScannerPlatform.instance = FakeMobileScannerPlatform(
```
navaronbracke, line 14: Should this be rest in a teardown for each test?

navaronbracke (changes requested):

juliansteenbakker:
Closing this for #1742, which adresses all issues pointed out by @navaronbracke

PR #1726, opened 2026-06-25
fix: support for enableR8.fullMode in AGP 8.0

### LINKED ISSUE #1722: An unexpected error occurred
I have been using mobile_scanner 7.20 successfully in my apps.  

The issue arises on the android release version of the app.  Tested on Android version 14.

I recently upgraded to Flutter 3.44.1 and since then in the released version of my app I am getting the following error when I call mobile_scanner.

- An unexpected error occurred. Attempt to invoke virtual method 'y4.d y4.c.a(u4.b)' on a null object reference

The debug version of the app created on 3.44.1 works as expected.

Going back to the earlier release version created using flutter 3.41.x also works as expected.

Bishop19:
Updated proguard-rules.pro since android.enableR8.fullMode is set to true on AGP 8.0 (see [https://developer.android.com/topic/performance/app-optimization/full-mode](https://developer.android.com/topic/performance/app-optimization/full-mode))

This intends to fix #1722.


juliansteenbakker:
Thanks!

PR #1746, opened 2026-07-16
fix: Android preview orientation fix

BartS-CCV:
### Summary

On Android devices that allow multiple orientations (e.g. tablets), the camera preview frequently ends up rotated 90°/180°. We traced this to three independent bugs in the orientation pipeline. All three were reproduced on a Samsung Galaxy Tab S9 (originally with mobile_scanner 7.2.1; the affected files are unchanged in 7.3.0), and all three are fixed by small patches we've been running in a vendored fork, verified on device. PR with the fixes: <link>.

### Bug 1: stale orientation reported on camera start

**Repro:** start the scanner in orientation A → `stop()` → rotate the device to orientation B → `start()` → the preview renders as if still in orientation A, until the device is physically rotated again.

**Cause:** `DeviceOrientationListener` caches the orientation in `lastOrientation`, which is only updated from the `onDisplayChanged` callback — and `stop()` unregisters the display listener, so rotations while the camera is stopped are never observed. On the next start, `getOrientation()` returns `lastOrientation ?: getUIOrientation()`, so the stale cache wins over the live display rotation, and the wrong `initialDeviceOrientation` reaches `RotatedPreview`/`CameraPreview`. The listener instance lives at plugin level, so the stale cache even survives disposing and recreating the controller.

**Fix:** re-sync in `DeviceOrientationListener.start()`:

```kotlin
fun start() {
    lastOrientation = getUIOrientation()
    // ...register display listener as before
}
```

### Bug 2: rotation while the camera is running is sometimes missed

**Repro:** with the camera running, rotate the device — occasionally (timing-dependent, notably on Samsung) the preview keeps the old orientation.

**Cause:** rotation detection relies solely on `DisplayManager.onDisplayChanged`. On some devices that callback can fire *before* `display.rotation` reflects the new value — the listener then reads the old rotation, deduplicates it as "no change", and no correction event is ever sent. Seamless 180° landscape↔landscape rotations can also fail to produce a useful callback.

**Fix:** funnel more triggers into the same deduplicated check: re-check the display rotation ~150ms/500ms after `onDisplayChanged` (catches stale-at-callback-time reads), and add an `OrientationEventListener` that triggers the check when the device physically turns a quadrant (catches rotations with no display callback). The orientation is still always read from the display rotation, so rotation lock and `SystemChrome.setPreferredOrientations` remain respected, and the dedup ensures no extra events reach Dart.

### Bug 3: orientation events die after `switchCamera()`

**Repro:** `switchCamera()` (front↔back), then rotate the device — the preview orientation is frequently (but not always) wrong from then on, and no longer follows rotation at all.

**Cause:** `switchCamera()` performs `stop()` + `start()`, and `stop()` in `MethodChannelMobileScanner` nulls the cached `_deviceOrientationStream`, so the next `start()` creates a **new** `EventChannel.receiveBroadcastStream()` instance on the same channel. `EventChannel`'s `onCancel` **unconditionally** clears the channel's message handler and sends `cancel` to the platform — even when a newer stream instance has since taken over the channel. So if a listener of the *old* stream instance unsubscribes *after* the new instance sent `listen` — which happens whenever no frame renders between `stop()` and `start()`, because `RotatedPreview`'s subscription (created in `initState`, never updated) is only torn down on the rebuild after the restart — the late cancel wipes the new stream's message handler and nulls the native event sink. From that point **no orientation event ever reaches Dart again** (neither the controller nor `RotatedPreview`), so the preview freezes at the switch-time orientation. Whether a frame renders inside the stop/start gap is scheduling luck, which makes the failure intermittent.

**Fix (two parts):**

1. Keep the cached `_deviceOrientationStream` alive across stop/start cycles (don't null it in `stop()`). With a single stream instance, the `EventChannel` listen/cancel reference counting is correct regardless of subscription timing. The events stream doesn't need this, since all of its subscriptions are owned by the controller and cancelled before the restart; the orientation stream is special because `RotatedPreview` subscribes to it with widget-tree lifecycle timing.
2. Emit the current orientation from the native side on every camera start, so listeners that survive a stop/start cycle are re-synced even if the device was rotated while the camera was stopped (no rotation change event was observable in that window).

A `didUpdateWidget` was also added to `RotatedPreview` so it resubscribes if the stream instance it was built with ever does change, instead of silently keeping a subscription to a defunct stream.

Related: the comment in `mobile_scanner_method_channel.dart` claiming the native stream "will emit the current orientation when the first listener is attached" doesn't match the Kotlin implementation — `onListen` emits nothing.

### Environment

- mobile_scanner: 7.2.1 / 7.3.0 (affected files identical)
- Flutter: 3.44.6 (stable)
- Device: Samsung Galaxy Tab S9 (app allows all four orientations); iOS unaffected
- Fixes verified on device across: cold start after rotation, live rotation (90°/180°, fast/slow), and camera switching combined with rotation


android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -33,13 +34,46 @@ class DeviceOrientationListener(
     // Called with the new [Surface.ROTATION_*] value whenever the display rotation changes.
     var onDisplayRotationChanged: ((Int) -> Unit)? = null
 
+    private val handler = Handler(Looper.getMainLooper())
```
juliansteenbakker, line 37: The following still uses the inline handler, it would be better to let it use this one instead:

https://github.com/BartS-CCV/mobile_scanner/blob/e3b46d17f3b9e61a92453d445973fea5dd806ae9/android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt#L174

lib/src/method_channel/rotated_preview.dart
```diff
@@ -118,6 +118,25 @@ final class _RotatedPreviewState extends State<RotatedPreview> {
     return rotationDegrees -= deviceOrientationDegrees;
   }
 
+  @override
+  void didUpdateWidget(RotatedPreview oldWidget) {
```
juliansteenbakker, line 122: It would be great if we could test this in a widget test

juliansteenbakker (changes requested):
Thanks for the PR, overall it looks good, just two minor things.

lib/src/method_channel/rotated_preview.dart
BartS-CCV: dded a widget test in test/method_channel/rotated_preview_test.dart covering the resubscribe-on-stream-change behavior

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
BartS-CCV: Switched to the shared handler.

BartS-CCV:
To be transparent: these fixes were found and made by AI (Fable 5).
Only with these fixes we were able to reliably rotate the camera preview on various Android tablets.

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
```diff
@@ -55,19 +89,58 @@ class DeviceOrientationListener(
...
+        handler.removeCallbacks(rotationCheck)
         onDisplayRotationChanged = null
     }
 
+    /**
+     * Schedule delayed re-checks of the display rotation, to catch rotations
+     * that are only reflected on the display after the triggering callback.
+     */
+    private fun scheduleDelayedRotationChecks() {
+        handler.postDelayed(rotationCheck, 150)
```
navaronbracke: Why is this specifically 150 and 500?

lib/src/method_channel/mobile_scanner_method_channel.dart
```diff
@@ -463,7 +463,17 @@ class MethodChannelMobileScanner extends MobileScannerPlatform {
     _pausing = false;
     _surfaceProducerDelegate = null;
     _eventsStream = null;
-    _deviceOrientationStream = null;
+    // The device orientation stream is deliberately kept alive across
```
navaronbracke: Should `_deviceOrientationStream = null;` be moved to dispose() ?

lib/src/method_channel/rotated_preview.dart
```diff
@@ -118,6 +118,25 @@ final class _RotatedPreviewState extends State<RotatedPreview> {
...
+    super.didUpdateWidget(oldWidget);
+    // The device orientation stream is recreated when the camera is restarted,
+    // e.g. when switching between the front and back camera. Resubscribe and
+    // adopt the new initial orientation, otherwise this widget keeps listening
+    // to the old, defunct stream and no longer follows device rotation.
+    if (!identical(
```
navaronbracke: Nit: use an early return?

test/method_channel/rotated_preview_test.dart
```diff
@@ -0,0 +1,141 @@
...
+      buildPreview(
+        stream: controller.stream,
+        initialOrientation: DeviceOrientation.portraitUp,
+      ),
+    );
+
+    controller.add(DeviceOrientation.landscapeRight);
+    await tester.pumpAndSettle();
+
+    expect(quarterTurns(tester), 2);
```
navaronbracke: Isn't this 1 turn?

navaronbracke (approved):
LGTM with nits

android/src/main/kotlin/dev/steenbakker/mobile_scanner/DeviceOrientationListener.kt
BartS-CCV: Honestly: they're empirical, not derived from any platform guarantee. The problem is that onDisplayChanged (and the sensor callback) can fire before display.rotation reports the new value, and Android gives no signal for when the rotation has settled — so the only options are to re-read it later or to miss the change entirely.

150ms covers the common case where the new rotation has landed shortly after the triggering callback; 500ms is a fallback for slower devices (the Galaxy Tab S9 where I hit this needed more than one attempt). Both go through checkRotation(), which only emits when the orientation actually changed, so a re-check that observes no change is a no-op — the delays only need to be "late enough", never exact.

I've pushed a commit naming them ROTATION_RECHECK_DELAY_MS / ROTATION_RECHECK_FALLBACK_DELAY_MS with that reasoning as a comment. If you'd prefer a different shape here — fewer/more attempts, or a short bounded poll until the rotation stops changing — happy to switch; I went with two fixed posts as the least machinery that fixed it in practice.

lib/src/method_channel/mobile_scanner_method_channel.dart
BartS-CCV: I'd argue it shouldn't be nulled there either, for the same reason as in stop().

MobileScannerPlatform._instance is a long-lived static, and dispose() doesn't destroy it — the same MethodChannelMobileScanner serves the next controller. So nulling in dispose() still allows a second receiveBroadcastStream() instance to be created on that channel while a RotatedPreview from a route that hasn't been removed yet is still subscribed to the first. When that old subscription is eventually cancelled, EventChannel's onCancel unconditionally clears the channel's message handler and sends cancel — killing the new instance. That's reachable today: multiple controllers can coexist (which is what _platformSessionOwner in 7.3.0 is about), so a controller can be disposed while another one's preview is still mounted.

The other half of the argument is that there's nothing to reclaim: the cached value is just a Stream object, and the platform-side handler is only registered between the first listen and the last cancel. While unlistened it holds no platform resources, and it's fully reusable — a later listen re-registers the handler (nothing calls endOfStream on this channel, so the underlying broadcast controller is never closed).

I've extended the comment to say this explicitly for dispose() too, since it's the obvious next question when reading it. If you'd rather have the field cleared somewhere for symmetry with _eventsStream, the safe version would be clearing it only once it has no listeners — but that's extra bookkeeping for no gain that I can see.

lib/src/method_channel/rotated_preview.dart
BartS-CCV: Done, pushed.

test/method_channel/rotated_preview_test.dart
BartS-CCV: Good catch that it looks wrong — but 2 is correct here, because RotatedPreview doesn't rotate by the device orientation. It applies the correction that's left after CameraPreview has already rotated the preview by the orientation (landscapeRight → 1 turn there). Per _computeRotationDegrees with sensorOrientationDegrees: 0, facingSign: 1:

portraitUp:     ((0 -  0 + 360) % 360) -  0 =   0° -> 0 turns
landscapeRight: ((0 - 90 + 360) % 360) - 90 = 180° -> 2 turns
I picked those two because they're unambiguous: with a zero sensor orientation portraitUp and portraitDown both come out as 0 turns, so they can't be used to prove the widget reacted. I've pushed a commit documenting that arithmetic in the test and naming the expected values, so the next reader doesn't have to re-derive it.

navaronbracke:
@BartS-CCV could you amend this with a patch version bump and changelog entry?

juliansteenbakker (approved):

Issue #1725, opened 2026-06-24
[Android] Unexpected error/behaviour using another FFI package on `--release`


LotharieSlayer:
### Context
Happening on both:
- Xiaomi Mi 9T - Android 11 (physical) 
- Android 15 (API 35) emulator

Things to know:
- mobile_scanner: ^7.2.0 (also tested with 7.1.3 and 7.0.1)
- Only happens on `--release`, mobile_scanner works perfectly on debug and profile
- Only an Android issue, works well on iOS, not tested on other platforms
- My code don't use minification or proguard rules

### Steps to reproduce
When I add [ffmpeg-kit-extended](https://github.com/akashskypatel/ffmpeg-kit-extended) to pubspec which i need for my app, then when I try to call mobile_scanner, this happens : 

```
An unexpected error occured. Attempt to invoke virtual method 'l4.d l4.c.a(h4.c)' on a null object reference
```
<img width="589" height="468" alt="Image" src="https://github.com/user-attachments/assets/27543d8c-9449-4bbb-b2e9-32a87397e153" />

I suppose there is a weird thing happening with FFI/compiler but I'm not sure, and there are no logs at all in console. I opened an issue on ffmpeg-kit-extended repo too. https://github.com/akashskypatel/ffmpeg-kit-extended/issues/28

Do you have any idea on what's happening? or any configuration/outcome for that case?

Thanks a lot ! :)

LotharieSlayer:
After investigations with de-obfuscation tool ProGuard's Retrace, here's what i found:

[stacktrace-retraced.txt](https://github.com/user-attachments/files/29315872/stacktrace-retraced.txt)

```
E  Failed to handle method call
   java.lang.NullPointerException: Attempt to invoke virtual method 'l4.d l4.c.a(h4.c)' on a null object reference
       at com.google.mlkit.vision.barcode.BarcodeScanning.getClient(com.google.android.gms:play-services-mlkit-barcode-scanning@@18.3.1:3)
       at dev.steenbakker.mobile_scanner.MobileScanner$Companion.defaultBarcodeScannerFactory(MobileScanner.kt:106)
       at dev.steenbakker.mobile_scanner.MobileScanner$1.invoke(MobileScanner.kt:60)
       at dev.steenbakker.mobile_scanner.MobileScanner$1.invoke(MobileScanner.kt:60)
       at dev.steenbakker.mobile_scanner.MobileScanner.start(MobileScanner.kt:397)
[...]

PR #1763, opened 2026-07-29
chore: Update ProGuard rules for ML Kit

navaronbracke:
Investigating https://github.com/juliansteenbakker/mobile_scanner/issues/1725


juliansteenbakker (changes requested):
Shouldn't it be `{ *; }` instead of `{ ; }` on all three lines? R8 fails to parse the rules file right now.

navaronbracke:
Woops, yeah, typo. Fixed now

juliansteenbakker (approved):
There seemed to be an issue with the spaces used. I've replaced them for plain spaces, looks good to me now.

Issue #1373, opened 2025-03-30
Failed to acquire image


raymond-holmboe:
on the latest tag, v7.0.0-beta.8, I never get to onDetect, I get a java error.
My code: 

import 'dart:async';

import 'package:flutter/foundation.dart';
import 'package:flutter/material.dart';
import 'package:mobile_scanner/mobile_scanner.dart';
import 'scanned_barcode_label.dart';
import 'scanner_button_widgets.dart';
import 'scanner_error_widget.dart';

class BarcodeScannerWithZoom extends StatefulWidget {
  const BarcodeScannerWithZoom({super.key});

  @override
  State<BarcodeScannerWithZoom> createState() => _BarcodeScannerWithZoomState();
}

class _BarcodeScannerWithZoomState extends State<BarcodeScannerWithZoom> {
  final MobileScannerController controller = MobileScannerController(
    torchEnabled: false,
    autoStart: true,
    detectionSpeed: DetectionSpeed.normal,
    // autoZoom: true,
    invertImage: true,
    cameraResolution: Size(1024, 720),
    returnImage: false,
    detectionTimeoutMs: 500,
  );
  double _zoomFactor = 0.0;
  bool isProcessing = false;

  Widget _buildZoomScaleSlider() {
    return ValueListenableBuilder(
      valueListenable: controller,
      builder: (context, state, child) {
        if (!state.isInitialized || !state.isRunning) {
          return const SizedBox.shrink();
        }

        final TextStyle labelStyle = Theme.of(context).textTheme.headlineMedium!.copyWith(color: Colors.white);

        return Padding(
          padding: const EdgeInsets.symmetric(horizontal: 8.0),
          child: Row(
            children: [
              Text('0%', overflow: TextOverflow.fade, style: labelStyle),
              Expanded(
                child: Slider(
                  value: _zoomFactor,
                  onChanged: (value) {
                    setState(() {
                      _zoomFactor = value;
                      controller.setZoomScale(value);
                    });
                  },
                ),
              ),
              Text('100%', overflow: TextOverflow.fade, style: labelStyle),
            ],
          ),
        );
      },
    );
  }

  void onBackPressed() {
    Navigator.of(context).pop();
  }

  void onScanPressed(String barcode) {
    Navigator.of(context).pop({'barcode': barcode});
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      // appBar: AppBar(title: const Text('With zoom slider')),
[...]

Issue #1310, opened 2025-01-28
Android - "java.lang.IllegalStateException: Image is already closed"


teunklijn:
Hi,

When I try to scan a QR code on my Android phone the app crashes with a IllegalStateException (Image is already closed). I'm currently using version 6.0.4 of the package, I cannot reproduce this issue on version 5.2.3. 

### Platform

**OS**: Android
**OS version**: Android 14 (OxygenOS 14.0)
**Phone**: OnePlus 8T (KB2003)
**Package version**: 6.0.4

### Example code

```
import 'package:flutter/material.dart';
import 'package:mobile_scanner/mobile_scanner.dart';

void main() {
  runApp(
    MaterialApp(
      home: Scaffold(
        body: SizedBox.expand(
          child: MobileScanner(
            onDetect: (BarcodeCapture? capture) {
              print(capture?.barcodes.firstOrNull?.rawValue);
            },
            onDetectError: (Object error, StackTrace trace) {
              print('Error: $error');
            },
          ),
        ),
      ),
    ),
  );
}
```

### Expected behaviour

The QR code is scanned and the onDetect callback is invoked with the result.

### Actual behaviour

The app crashes when detecting a QR code with the following exception:

```
E/AndroidRuntime(28830): FATAL EXCEPTION: main
E/AndroidRuntime(28830): Process: com.example.mobile_scanner_test, PID: 28830
E/AndroidRuntime(28830): java.lang.IllegalStateException: Image is already closed
E/AndroidRuntime(28830):        at android.media.Image.throwISEIfImageIsInvalid(Image.java:79)
E/AndroidRuntime(28830):        at android.media.ImageReader$SurfaceImage.getWidth(ImageReader.java:1194)
E/AndroidRuntime(28830):        at dev.steenbakker.mobile_scanner.MobileScanner$captureOutput$1$1$1.invoke(MobileScanner.kt:135)
E/AndroidRuntime(28830):        at dev.steenbakker.mobile_scanner.MobileScanner$captureOutput$1$1$1.invoke(MobileScanner.kt:96)
E/AndroidRuntime(28830):        at dev.steenbakker.mobile_scanner.MobileScanner.captureOutput$lambda$4$lambda$2$lambda$0(MobileScanner.kt:96)
[...]

PR #1764, opened 2026-07-29
fix: always release ImageProxy in the Android analyzer

WladmirJunior:
## Description

On Android, the `ImageAnalysis` analyzer leaves the `ImageProxy` open on several exit paths:

- `imageProxy.image` returning null (`return@Analyzer` without close)
- ML Kit `process()` failing (`addOnFailureListener` never closes the frame)
- ML Kit `process()` being canceled (neither listener runs)
- any exception thrown inside the success callback or during `InputImage` creation

With `STRATEGY_KEEP_ONLY_LATEST`, CameraX does not deliver the next frame until the current proxy is closed, so a single leaked frame permanently stalls barcode analysis while the camera preview keeps running. The user sees a live preview that never scans, with no error surfaced. This looks related to the "Failed to acquire image" reports in #1373.

This PR makes the analyzer release each frame exactly once on every exit path, synchronous or asynchronous:

- a close-once `releaseFrame()` helper (guarded by `AtomicBoolean`) that also recycles the inverted bitmap consistently; the close-once guard also prevents double-close crashes like #1310
- `try`/`catch` around `InputImage` creation, the success callback body, the `returnImage` coroutine, and the synchronous `process()` call, all funneling into `releaseFrame()`
- `addOnFailureListener` and `addOnCanceledListener` now release the frame before reporting the error
- the NORMAL detection-speed timeout is now re-armed in the same branch that sets `scannerTimeout`, so a synchronous failure can no longer leave the scanner in a permanent timeout state

Testability: the analyzer previously called the static `InputImage.fromMediaImage` directly, which cannot be mocked. It is now an injectable `inputImageFactory` constructor parameter (defaulting to `InputImage::fromMediaImage`), and `scanner` is `@VisibleForTesting internal`, so the analyzer paths can be unit tested.

Tests: five new Robolectric regression tests cover the null-image, ML Kit failure, ML Kit cancellation, empty-result, and synchronous-throw paths (`./gradlew :mobile_scanner:testDebugUnitTest`, 9/9 passing). Also smoke-tested on a physical Samsung SM-A346M: camera starts, frames flow through the analyzer, and ML Kit keeps receiving frames.

## Checklist

- [x] The PR title follows [[Conventional Commits](https://www.conventionalcommits.org/)](https://www.conventionalcommits.org/) (e.g. `feat: ...`, `fix: ...`, `chore: ...`): this is required by CI and drives the automated changelog/release.
- [x] (Optional) Each individual commit also follows Conventional Commits: if so, we'll merge with a real merge commit to preserve your full history; otherwise we'll squash the PR into one commit.
- [x] Tests were added or updated, if applicable.
- [ ] Documentation (`README.md`) was updated, if applicable.


android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -112,128 +116,180 @@ class MobileScanner(
...
-                        barcodeMap,
-                        byteArray,
-                        bmWidth,
-                        bmHeight
-                    )
-
-                    // Clean up resources
-                    rotatedBitmap.recycle()
-                    imageProxy.close()
+                    // Use Coroutine to process the image and generate the Bitmap to prevent main UI
```
navaronbracke, line 227:
```suggestion
                    // Use a Coroutine to process the image and prevent dropping frames on the main thread.
```

android/src/main/kotlin/dev/steenbakker/mobile_scanner/MobileScanner.kt
```diff
@@ -112,128 +116,180 @@ class MobileScanner(
...
+                        var baseBitmap: Bitmap? = null
+                        var rotatedBitmap: Bitmap? = null
+                        try {
+                            // Get bitmap for image return. reuse inverted bitmap if available, otherwise create from imageProxy
+                            val sourceBitmap = invertedBitmap ?: imageProxy.toBitmap()
+                            baseBitmap = sourceBitmap
+
+                            // Rotate the bitmap based on the camera's rotation degrees
+                            var outputBitmap = rotateBitmap(sourceBitmap, camera?.cameraInfo?.sensorRotationDegrees ?: 90)
+                            rotatedBitmap = outputBitmap
```
navaronbracke, line 238: Is the code path for the rotated bitmap still correct? I only see it assigned but not actually used beyond this point?

navaronbracke (dismissed):
Just a minor nit about a variable that I think was not ported over correctly.

This will also need a patch version bump & changelog entry

WladmirJunior:
Thanks for picking this up while I was out, dropping the extra variable reads better than what I had.

One small thing, not a blocker: when `invertImage` is true and the sensor rotation is 0, `rotateBitmap` returns the source instance, so `outputBitmap` is the same object as `invertedBitmap`, which `releaseFrame` also recycles. The `isRecycled` guards keep it safe today, so nothing is broken.

If you want it tighter:

```kotlin
if (outputBitmap !== invertedBitmap && !outputBitmap.isRecycled) {
    outputBitmap.recycle()
}
```

Happy to push that, or leave it as is. Good to go from my side either way.

juliansteenbakker:
> One small thing, not a blocker: when `invertImage` is true and the sensor rotation is 0, `rotateBitmap` returns the source instance, so `outputBitmap` is the same object as `invertedBitmap`, which `releaseFrame` also recycles. The `isRecycled` guards keep it safe today, so nothing is broken.
> 
> If you want it tighter:
> 
> ```kotlin
> if (outputBitmap !== invertedBitmap && !outputBitmap.isRecycled) {
>     outputBitmap.recycle()
> }
> ```

Thanks! I've looked at it, but I think that guard would never actually fire. `invertBitmapColors` always allocates a new bitmap, so when `invertImage` is `true` `outputBitmap` has already been reassigned to `revertedBitmap` by the time we reach the finally. It can't be `invertedBitmap` there. And when `invertImage` is `false`, `invertedBitmap` is `null`, so there's nothing to alias either way.

juliansteenbakker (approved):

WladmirJunior:
@juliansteenbakker do you can release a version with this fix?

PR #1779, opened 2026-08-20
fix(web): release the camera when a polling barcode reader stops

aicorey:
## Description

`PollingBarcodeReader.stop()` drops its `MediaStream` reference without stopping the tracks, so the camera acquired by `getUserMedia` stays in `readyState: 'live'` after the scanner is closed. The device camera and the OS camera indicator stay on until the page is reloaded.

Both default web backends extend this class — `BarcodeDetectorReader` and `ZXingWasmBarcodeReader` — so `WebBarcodeReader.auto` leaks the camera on every browser. The legacy `WebBarcodeReader.zxingJs` backend is unaffected, because `ZXingBarcodeReader.stop()` releases the stream through ZXing's `reset()`.

```dart
Future<void> stop() async {
  ...
  _videoElement = null;
  _videoStream = null;   // <- tracks are never stopped
  disposeDecoder();
}
```

This PR stops every track and detaches the stream from the video element before dropping the references, via a new `stopVideoStream()` helper in `web_camera_utility.dart`.

### Reproducing

1. Open a scanner on the web with the default `WebBarcodeReader.auto`.
2. Close it (or let a successful decode close it).
3. The OS camera indicator stays lit, and the acquired stream's tracks still report `readyState: 'live'`.

### Tests

`test/web/polling_barcode_reader_stop_test.dart` (browser, `@TestOn('browser')`) acquires a real `MediaStream` from `HTMLCanvasElement.captureStream()` — no camera permission needed — and asserts the tracks reach `readyState: 'ended'` and `videoElement.srcObject` is cleared after `stop()`. The tests fail on `develop` and pass with the fix.

Run with `flutter test --platform chrome test/web/`. Note that `test/web/web_library_versions_test.dart` already fails under `--platform chrome` on `develop` (it reads `package.json` from disk); that is unrelated to this change.

## Checklist

- [x] The PR title follows [Conventional Commits](https://www.conventionalcommits.org/) (e.g. `feat: ...`, `fix: ...`, `chore: ...`) — this is required by CI and drives the automated changelog/release.
- [x] (Optional) Each individual commit also follows Conventional Commits — if so, we'll merge with a real merge commit to preserve your full history; otherwise we'll squash the PR into one commit.
- [x] Tests were added or updated, if applicable.
- [ ] Documentation (`README.md`) was updated, if applicable.

### DIFF
--- a/test/mobile_scanner_controller/dispose_during_start_test.dart
+++ b/test/mobile_scanner_controller/dispose_during_start_test.dart
@@ -1,0 +1,93 @@
+import 'dart:async';
+
+import 'package:flutter/services.dart';
+import 'package:flutter_test/flutter_test.dart';
+import 'package:mobile_scanner/src/enums/camera_facing.dart';
+import 'package:mobile_scanner/src/enums/torch_state.dart';
+import 'package:mobile_scanner/src/mobile_scanner_controller.dart';
+import 'package:mobile_scanner/src/mobile_scanner_platform_interface.dart';
+import 'package:mobile_scanner/src/mobile_scanner_view_attributes.dart';
+import 'package:mobile_scanner/src/objects/barcode_capture.dart';
+import 'package:mobile_scanner/src/objects/start_options.dart';
+
+void main() {
+  TestWidgetsFlutterBinding.ensureInitialized();
+
+  late SlowStartMobileScannerPlatform platform;
+
+  setUp(() {
+    platform = SlowStartMobileScannerPlatform();
+    MobileScannerPlatform.instance = platform;
+    MobileScannerController.resetPlatformSessionOwner();
+  });
+
+  test(
+    'a controller disposed while starting does not claim the camera session',
+    () async {
+      final controller = MobileScannerController(autoStart: false)..attach();
+
+      final startFuture = controller.start();
+
+      await controller.dispose();
+
+      // The platform start only completes after the controller is gone,
+      // which is the interleaving a user creates by dismissing the scanner
+      // while the camera is still being acquired.
+      platform.completeStart();
+      await startFuture;
+
+      // The disposed controller must not be left holding the platform camera
+      // session: the next controller to dispose would then skip the platform
+      // teardown, and the camera would stay live.
+      final next = MobileScannerController(autoStart: false)..attach();
+
+      await next.dispose();
+
+      expect(platform.disposeCalls, 2);
+    },
+  );
+}
+
+class SlowStartMobileScannerPlatform extends MobileScannerPlatform {
+  int disposeCalls = 0;
+  int stopCalls = 0;
+
+  final Completer<void> _startGate = Completer<void>();
+
+  void completeStart() => _startGate.complete();
+
+  @override
+  Stream<BarcodeCapture?> get barcodesStream => const Stream.empty();
+
+  @override
+  Stream<TorchState> get torchStateStream =>
+      Stream.value(TorchState.unavailable);
+
+  @override
+  Stream<double> get zoomScaleStateStream => Stream.value(1);
+
+  @override
+  Future<MobileScannerViewAttributes> start(StartOptions startOptions) async {
+    await _startGate.future;
+
+    return const MobileScannerViewAttributes(
+      cameraDirection: CameraFacing.back,
+      currentTorchMode: TorchState.unavailable,
+      size: Size(200, 200),
+      numberOfCameras: 1,
+      initialDeviceOrientation: DeviceOrientation.portraitUp,
+    );
+  }
+
+  @override
+  Future<void> stop() {
+    stopCalls++;
+    return Future.value();
+  }
+
+  @override
+  Future<void> dispose() {
+    disposeCalls++;
+    return Future.value();
+  }
+}
--- a/test/web/mobile_scanner_web_start_teardown_test.dart
+++ b/test/web/mobile_scanner_web_start_teardown_test.dart
@@ -1,0 +1,215 @@
+@TestOn('browser')
+library;
+
+import 'dart:async';
+import 'dart:js_interop';
+import 'dart:js_interop_unsafe';
+import 'dart:ui';
+
+import 'package:flutter_test/flutter_test.dart';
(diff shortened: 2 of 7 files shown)


test/mobile_scanner_controller/dispose_during_start_test.dart
```diff
@@ -0,0 +1,93 @@
...
+      await startFuture;
+
+      // The disposed controller must not be left holding the platform camera
+      // session: the next controller to dispose would then skip the platform
+      // teardown, and the camera would stay live.
+      final next = MobileScannerController(autoStart: false)..attach();
+
+      await next.dispose();
+
+      expect(platform.disposeCalls, 2);
```
navaronbracke, line 46: Should this somehow also assert on the camera session? Or is that redundant?

test/mobile_scanner_controller/dispose_during_start_test.dart
```diff
@@ -0,0 +1,93 @@
...
+    },
+  );
+}
+
+class SlowStartMobileScannerPlatform extends MobileScannerPlatform {
+  int disposeCalls = 0;
+  int stopCalls = 0;
```
navaronbracke, line 53: The stop calls is not used?

test/web/mobile_scanner_web_start_teardown_test.dart
```diff
@@ -0,0 +1,215 @@
...
+import 'package:mobile_scanner/src/enums/detection_speed.dart';
+import 'package:mobile_scanner/src/enums/web_barcode_reader.dart';
+import 'package:mobile_scanner/src/mobile_scanner_exception.dart';
+import 'package:mobile_scanner/src/objects/start_options.dart';
+import 'package:mobile_scanner/src/web/mobile_scanner_web.dart';
+import 'package:web/web.dart' as web;
+
+/// `HTMLCanvasElement.captureStream()` yields a real [web.MediaStream] backed
+/// by a live video track, without prompting for camera permission.
+extension type _CaptureStreamCanvas(JSObject _) implements JSObject {
```
navaronbracke, line 22: Should some of this become test only utilities for web tests?

test/web/mobile_scanner_web_start_teardown_test.dart
```diff
@@ -0,0 +1,215 @@
...
+  initialZoom: 1,
+);
+
+@JS('eval')
+external JSAny? _jsEval(String code);
+
+/// Replaces the global `BarcodeDetector` with one whose constructor throws, so
+/// that a start fails at the point where it has already acquired the camera.
+void _installFailingBarcodeDetector() {
+  _jsEval('''
```
navaronbracke, line 62: Is there a reason we need to use eval, instead of an extension type binding for tests, on the Window? (i recommend looking at the docs for mocking extension types for the web: https://dart.dev/interop/js-interop/mock)

test/web/mobile_scanner_web_start_teardown_test.dart
```diff
@@ -0,0 +1,215 @@
...
+  delete globalThis.__msOriginalBarcodeDetector;
+}
+''');
+}
+
+/// Replaces `navigator.mediaDevices.getUserMedia` with a stub whose promise
+/// resolves only when the test says so, so that teardown can be interleaved
+/// with an in-flight camera acquisition — the interleaving a user creates by
+/// dismissing the scanner before the camera is ready.
+class _GetUserMediaStub {
```
navaronbracke, line 87: Same here for the mocking

test/web/mobile_scanner_web_start_teardown_test.dart
```diff
@@ -0,0 +1,215 @@
...
+
+      final startFuture = plugin.start(_startOptions);
+
+      await getUserMedia.awaitCall();
+
+      await plugin.stop();
+
+      final stream = getUserMedia.resolve();
+
+      await expectLater(startFuture, throwsA(isA<MobileScannerException>()));
```
navaronbracke, line 186: should this also have a reason?

test/web/mobile_scanner_web_start_teardown_test.dart
```diff
@@ -0,0 +1,215 @@
...
+
+      expect(allEnded(stream), isTrue);
```
navaronbracke, line 188: same here for the reason

test/web/polling_barcode_reader_stop_test.dart
```diff
@@ -0,0 +1,153 @@
...
+import 'package:mobile_scanner/src/objects/barcode.dart';
+import 'package:mobile_scanner/src/objects/start_options.dart';
+import 'package:mobile_scanner/src/web/polling_barcode_reader.dart';
+import 'package:mobile_scanner/src/web/web_camera_utility.dart';
+import 'package:web/web.dart' as web;
+
+/// `HTMLCanvasElement.captureStream()` yields a real [web.MediaStream] backed
+/// by a live video track, without prompting for camera permission — the only
+/// way to observe track teardown in a headless browser.
+extension type _CaptureStreamCanvas(JSObject _) implements JSObject {
```
navaronbracke, line 21: This is duplicated with the other test? (which is why i would go for test only utilities for web tests)

test/web/polling_barcode_reader_stop_test.dart
```diff
@@ -0,0 +1,153 @@
...
+      await reader.start(
+        _startOptions,
+        videoElement: web.HTMLVideoElement()..muted = true,
+        videoStream: stream,
+      );
+
+      await reader.stop();
+      await reader.stop();
+
+      expect(tracks.every((t) => t.readyState == 'ended'), isTrue);
```
navaronbracke, line 150: should we assert that no exception is thrown?

navaronbracke (changes requested):
Overall this seems okay to me, if the purpose is to guard against race conditions.

The tests could use some love though.

theartcher:
@aicorey Have you had time to review his requested changes? I'm personally experiencing issues with this and would LOVE to see this fix being implemented!

If you don't have time to work on it do let me know, then we can go from there and perhaps have others (including me) jump in as well

No pressure, just wanna know where you're standing in regard to this PR

PR #1784, opened 2026-08-27
fix(android): bypass false-positive KGP warning using pluginManager

AndresMontaniv:
### Description
Swaps the legacy `apply plugin:` syntax for the `PluginManager` API in the Android build script to bypass a false-positive static regex scanner in recent Flutter versions.

### Why this is needed
Users with `android.builtInKotlin=true` are currently receiving this warning:
`WARNING: Your app uses the following plugins that apply Kotlin Gradle Plugin (KGP): mobile_scanner`

This happens because `FlutterPluginUtils.getSubprojectPluginState` regex-scans `android/build.gradle` as plain text, meaning it ignores the `if (!builtInKotlin)` conditional and flags the file anyway. 

### Safety & Compatibility
* `Project.apply(plugin: id)` delegates directly to `pluginManager.apply(id)`, so this introduces **zero behavioral changes**.
* The AGP 8 fallback path remains fully intact for backward compatibility.

### Verification
Ran the package's `example` app to confirm the warning is eliminated and builds succeed across both legacy and current Flutter environments. 

Tested environments:
* Flutter 3.47.1 / AGP 9.1.0
* Flutter 3.44.5

| `builtInKotlin` | KGP Warning (Before) | KGP Warning (After) | Build Success |
| :--- | :--- | :--- | :--- |
| `false` | Yes | **No** | ✓ |
| `true` | Yes (false positive) | **No** | ✓ |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 **Bug Fixes**
 
 * [Android] Fixed a host-app crash when CameraX fails to initialize while starting the scanner (e.g. "Available cameras: 0", the camera being held by another process, or a transient HAL error). The failure is now routed through the normal error callback so it can be surfaced via `errorBuilder` instead of terminating the app.
+* [Android] Bypassed false-positive Kotlin Gradle Plugin (KGP) warning by migrating to the `PluginManager` API.
 
 ## 7.4.0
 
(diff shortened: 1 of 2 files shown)


navaronbracke (approved):
LGTM, although this will need a changelog entry

AndresMontaniv:
Good call! I just pushed a commit adding the entry under the 7.4.1 Bug Fixes section. Let me know if we are good to go!

juliansteenbakker:
Thanks for the check! Your diagnosis is right, but i've ended up fixing it in #1786 as part of a bigger migration from Groovy to KTS. Since it replaces the file this patch edits, I'm closing this as superseded rather than merged.

